PR #1, opened 2017-11-06
Pass through jest --watch argument

sergeyevstifeev:


klippx:
Isn't this only for old versions of yarn?

sergeyevstifeev:
@klippx you're right, upgraded, got this:
```warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.```

PR #2, opened 2017-11-06
Keep refreshing metadata on repeated broker not found errors

sergeyevstifeev:
Assuming coordinator broker will always be found on the second attempt is wishful. This PR tries to address this by using the retry policy.


tulios:
Awesome. I'm currently working on the consumer, making it more resilient to cluster failures. I can tell you that I'm not super happy with the retry architecture, I have added some retry capabilities to the connection but I'm wondering if I should remove and deal with retries only on the high-level structures (consumer, producer, etc)

tulios:
I have refactored the retry mechanism and pushed the consumer group, both producer and consumer are a lot more resilient after this change :)

PR #5, opened 2017-12-06
Improve user API

tulios:
* Add `offsetLag` to consumer batch
* Expose to `eachBatch` if the consumer is running (`isRunning`), which allow users to stop processing the batch when the consumer is shutting down

@Nevon ^

### DIFF
--- a/src/consumer/batch.js
+++ b/src/consumer/batch.js
@@ -25,11 +25,12 @@ module.exports = class Batch {
   }
 
   offsetLag() {
-    return this.isEmpty()
-      ? '0'
-      : Long.fromValue(this.highWatermark)
-          .add(-1)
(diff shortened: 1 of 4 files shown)


src/consumer/batch.js
```diff
@@ -23,4 +23,13 @@ module.exports = class Batch {
           .toString()
       : this.messages[this.messages.length - 1].offset
   }
+
+  offsetLag() {
+    return this.isEmpty()
+      ? '0'
+      : Long.fromValue(this.highWatermark)
+          .add(-1)
```
Nevon: Why was this -1 necessary again? I think it makes sense to make this a named constant to avoid having to ask this question again.

tulios: `highWatermark` has the next possible offset. The last offset of the topic is `highWatermark - 1`

PR #7, opened 2017-12-11
Recover from rebalance in progress when joining

tulios:


src/consumer/runner.js
```diff
@@ -23,8 +27,22 @@ module.exports = class Runner {
   }
 
   async join() {
-    await this.consumerGroup.join()
-    await this.consumerGroup.sync()
+    return this.retrier(async (bail, retryCount, retryTime) => {
```
Nevon: We already talked F2F about this, but just for the record, I don't think this will work in the general case, as each node that joins will trigger a rebalancing, so if we sequentially have 10 nodes join, the last one will fail 9 times before possibly succeeding.

We should instead keep retrying forever, as long as the error we get is still that the group is rebalancing.

tulios: Agreed. We have to review the retry mechanism of some things, some errors should be constantly retried.

PR #11, opened 2017-12-14
Fix index out of range errors when decoding partial messages

tulios:
@matthiasfeist have reported this bug


src/protocol/messageSet/decoder.js
```diff
@@ -64,7 +65,20 @@ const EntriesDecoder = (decoder, compressedMessage) => {
 }
 
 const EntryDecoder = decoder => {
+  if (!decoder.canReadInt64()) {
```
Nevon: Won't this mean that we still commit to that offset, even though we didn't get enough bytes to be able to read the last message?

src/protocol/messageSet/decoder.js
```diff
@@ -64,7 +65,20 @@ const EntriesDecoder = (decoder, compressedMessage) => {
...
+    throw new KafkaJSPartialMessageError(
+      `Tried to decode a partial message: There isn't enough bytes to read the offset`
+    )
+  }
+
   const offset = decoder.readInt64().toString()
+
+  if (!decoder.canReadInt32()) {
```
Nevon: We should probably consider refactoring the decoder readInt* methods to do this check instead, so we don't keep running into this problem.

src/protocol/messageSet/decoder.js
tulios: I thought the same but I did a check and we commit the offset of the last message

https://github.com/tulios/kafkajs/blob/76bca638368693ffde6caee0ad3d6242ce9cafbe/src/consumer/batch.js#L24

https://github.com/tulios/kafkajs/blob/76bca638368693ffde6caee0ad3d6242ce9cafbe/src/consumer/runner.js#L147

https://github.com/tulios/kafkajs/blob/76bca638368693ffde6caee0ad3d6242ce9cafbe/src/consumer/runner.js#L157

So we are safe on this

src/protocol/messageSet/decoder.js
tulios: Yeah, this solution isn't great but I'm still not seeing a better pattern here. If we move this to the decoder we lose the context since it's too deep in the implementation to understand that it's a partial message or something else. Maybe adding the boundary checks there and recover from this error everywhere? I'm not sure yet

PR #19, opened 2017-12-20
Don't use seed broker for consumer errors

tulios:
The seed broker connection is been dropped forcing the consumers to rejoin the group triggering rebalances every X minutes. This PR removes the need for the seed broker connection and makes `cluster#refreshMetadata` reconnects if it needs to fetch data from the seed broker.

`consumerGroup#findGroupCoordinator` will automatically connect/reconnect the broker if it's not connected so removing the check for a connected cluster in there.

\+ Improved logs


tulios:
We have the same issue with the producer but it should be addressed in a different PR

tulios:
This PR isn't complete yet

src/cluster/index.js
```diff
@@ -105,6 +105,7 @@ module.exports = class Cluster {
    * @returns {Promise}
    */
   async refreshMetadata() {
+    await this.connect()
```
Nevon: This is kind of an aside, but the `connect` method won't actually cycle through all brokers when trying to set up the seed broker. It will obey the user defined retry configuration, which by default means that it will retry 5 of the brokers (or cycle through the same broker multiple times if brokers < retries). I think we should change `cluster.connect` to use a retrier with retries = brokers - 1.

tulios:
This PR should be complete now, it introduces a new abstraction layer called `BrokerPool` which should handle the access to brokers preventing future code to rely on the seed brokers.

src/cluster/index.js
tulios: This logic was moved to the `BrokerPool` and doesn't rely on the seed brokers anymore.

PR #23, opened 2017-12-28
Add consumer seek

tulios:
This PR exposes a new function to allow consumers to seek offsets for a given topic/partition. This feature is supported by the java client and it's quite useful for error handling.

The proposed API is:

```javascript
await consumer.subscribe({ topic: 'topic-1' })
consumer.run({ eachMessage: async () => true })
consumer.seek({ topic: 'topic-1', partition: 1, offset: '123854' })
consumer.seek({ topic: 'topic-1', partition: 2, offset: '59839' })
consumer.seek({ topic: 'topic-1', partition: 3, offset: '983412' })
```

`consumer#seek` must be called after `run` because the `ConsumerGroup` must be initialized but the seek only happens in the fetch loop of the consumer assigned to the topic/partition after the join/sync.

### DIFF
--- a/src/consumer/consumerGroup.js
+++ b/src/consumer/consumerGroup.js
@@ -1,6 +1,7 @@
 const flatten = require('../utils/flatten')
 const OffsetManager = require('./offsetManager')
 const Batch = require('./batch')
+const SeekOffsets = require('./seekOffsets')
 const { KafkaJSError } = require('../errors')
 const { HEARTBEAT } = require('./instrumentationEvents')
 
@@ -40,7 +41,7 @@ module.exports = class ConsumerGroup {
     this.maxBytes = maxBytes
     this.maxWaitTime = maxWaitTimeInMs
 
-    this.seekOffset = []
+    this.seekOffset = new SeekOffsets()
     this.coordinator = null
     this.generationId = null
     this.leaderId = null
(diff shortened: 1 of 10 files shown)


src/consumer/consumerGroup.js
```diff
@@ -119,6 +120,21 @@ module.exports = class ConsumerGroup {
...
+  /**
+   * Update the consumer offset for the given topic/partition. This will be used
+   * on the next fetch. If this API is invoked for the same topic/partition more
+   * than once, the latest offset will be used on the next fetch.
+   *
+   * @param {string} topic
+   * @param {number} partition
+   * @param {string} offset
+   */
+  seek({ topic, partition, offset }) {
```
Nevon: Given that you only want a single offset per topic and partition, I would suggest changing this to using a Map instead, keying by topic and partition.

```
seek({ topic, partition, offset }) {
  seekOffset.set([topic, partition], offset)
}
```

The usage of the seekOffset becomes a little more tedious, as you have to get an iterable, get the first element and then delete it from the map, but I think it's worth it because you would be guaranteed uniqueness per topic/partition without this somewhat confusing code (took me longer than I'd like to admit to understand why you were filtering the offsets).

To mitigate the annoyance of not having `pop` for a map, you could wrap it into another datastructure that does that under the hood.

Nevon: Fixed this in b276ccf. Let me know if you think it's alright.

PR #24, opened 2017-12-29
Improve logger and add docs

tulios:
This PR does 3 things:

1) Changes the log function into a factory function. The change improves how KafkaJS can interoperate with different log libraries, it also add tests for the logger module
2) Uses the client retry configurations as the default values for producers and consumers
3) Adds extensive but still incomplete documentation

This is the beginning of a proper documentation.

@ticolucci @matthiasfeist can you also take a look at the readme?

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,15 +160,20 @@ KAFKAJS_LOG_LEVEL=info node code.js
 
 __How to create a log creator?___
 
-A log creator is a function which receives a log level and returns a log function. The log function receives two arguments: namespace and log.
+A log creator is a function which receives a log level and returns a log function. The log function receives: namespace, level, label, and log.
 
 `namespace` identifies the component which is performing the log, for example, connection or consumer.
 
-`log` is an object with the following keys: `level`, `timestamp`, `logger`, and `message`
+`level` is the log level of the log entry.
+
+`label` is a text representation of the log level, example: 'INFO'.
+
+`log` is an object with the following keys: `timestamp`, `logger`, `message`, and the extra keys given by the user. (`logger.info('test', { extra_data: true })`)
 
 ```javascript
 {
-  level: 'INFO', // NOTHING, ERROR, WARN, INFO, or DEBUG
+  level: 4,
+  label: 'INFO', // NOTHING, ERROR, WARN, INFO, or DEBUG
   timestamp: '2017-12-29T13:39:54.575Z',
   logger: 'kafkajs',
   message: 'Started',
@@ -179,22 +184,28 @@ A log creator is a function which receives a log level and returns a log functio
 The general structure looks like this:
 
 ```javascript
-const MyLogCreator = logLevel => (namespace, log) => {
+const MyLogCreator = logLevel => ({ namespace, level, label, log }) => {
   // Example:
-  // const { level, timestamp, logger, message, ...others } = log
-  // console.log(`${level} [${namespace}] ${message} ${JSON.stringify(others)}`)
+  // const { timestamp, logger, message, ...others } = log
+  // console.log(`${label} [${namespace}] ${message} ${JSON.stringify(others)}`)
 }
 ```
 
 Example using [winston](https://github.com/winstonjs/winston):
 
 ```javascript
+const { logLevel } = require('kafkajs')
 const winston = require('winston')
 const toWinstonLogLevel = level => switch(level) {
-  case 'NOTHING':
+  case logLevel.ERROR:
+  case logLevel.NOTHING:
     return 'error'
-  default:
-    return level.toLowerCase()
+  case logLevel.WARN:
+    return 'warn'
+  case logLevel.INFO:
+    return 'info'
+  case logLevel.DEBUG:
+    return 'debug'
 }
 
 const WinstonLogCreator = logLevel => {
@@ -206,11 +217,11 @@ const WinstonLogCreator = logLevel => {
     ]
   })
 
-  return (namespace, { level, message, ...extra }) => {
+  return ({ namespace, level, { message, ...extra } }) => {
     logger.log({
       level: toWinstonLogLevel(level),
       message,
-      extra
+      extra,
     })
   }
 }
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ __In active development - alpha__
     - [Custom partitioner](#producing-messages-custom-partitioner)
     - [Retry](#producing-messages-retry)
   - [Consuming messages from Kafka](consuming-messages)
+    - [eachMessage](consuming-messages-each-message)
+    - [eachBatch](consuming-messages-each-batch)
     - [Options](consuming-messages-options)
     - [Custom assigner](#consuming-messages-custom-assigner)
   - [Instrumentation](#instrumentation)
@@ -58,7 +60,7 @@ const kafka = new Kafka({
 
 #### <a name="setup-client-ssl"></a> SSL
 
-The `ssl` option can be used to configure the TLS sockets. The options are passed directly to `tls.connect` and used to create the TLS Secure Context, all options are accepted.
+The `ssl` option can be used to configure the TLS sockets. The options are passed directly to [`tls.connect`](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback) and used to create the TLS Secure Context, all options are accepted.
 
 ```javascript
 const fs = require('fs')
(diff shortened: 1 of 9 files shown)


tulios:
It's missing a topic about GZIP compression in the producer section

README.md
```diff
@@ -3,30 +3,233 @@
...
 const kafka = new Kafka({
   clientId: 'my-app',
   brokers: ['kafka1:9092', 'kafka2:9092']
 })
 ```
 
-### Producing Messages to Kafka
+#### <a name="setup-client-ssl"></a> SSL
+
+The `ssl` option can be used to configure the TLS sockets. The options are passed directly to `tls.connect` and used to create the TLS Secure Context, all options are accepted.
```
matthiasfeist: maybe link to the docs for tls.connect

README.md
```diff
@@ -3,30 +3,233 @@
...
+new Kafka({
+  clientId: 'my-app',
+  brokers: ['kafka1:9092', 'kafka2:9092'],
+  connectionTimeout: 3000
+})
+```
+
+#### <a name="setup-client-default-retry"></a> Default Retry
+
+The `retry` option can be used to set the default configuration. The retry mechanism uses a randomization function that grows exponentially.
```
matthiasfeist: You could mention what you actually retry: Just the connect or also the production of messages or ...?
Also: Mention what happens after the max number of retries has exceeded.

README.md
```diff
@@ -3,30 +3,233 @@
...
+KAFKAJS_LOG_LEVEL=info node code.js
+```
+
+__How to create a log creator?___
+
+A log creator is a function which receives a log level and returns a log function. The log function receives two arguments: namespace and log.
+
+`namespace` identifies the component which is performing the log, for example, connection or consumer.
+
+`log` is an object with the following keys: `level`, `timestamp`, 'logger', and `message`
```
matthiasfeist: you did use a apostrophe around "logger" instead of an accent. I'm not sure why and which effect that has...

README.md
```diff
@@ -3,30 +3,233 @@
...
+  // Example:
+  // const { level, timestamp, logger, message, ...others } = log
+  // console.log(`${level} [${namespace}] ${message} ${JSON.stringify(others)}`)
+}
+```
+
+Example using [winston](https://github.com/winstonjs/winston):
+
+```javascript
+const winstom = require('winston')
```
matthiasfeist: not winsto**_m_** but winsto**_n_** (several times in your code)

README.md
```diff
@@ -3,30 +3,233 @@
...
+  logLevel: logLevel.ERROR,
+  logCreator: WinstomLogCreator
+})
+```
+
+### <a name="producing-messages"></a> Producing Messages to Kafka
+
+To publish messages to Kafka you have to create a producer. By default the producer is configured to distribute the messages with the following logic:
+
+- If a partition is specified in the message, use it
```
matthiasfeist: maybe demo how to specify a partition in the producer

README.md
```diff
@@ -46,7 +249,48 @@ async () => {
...
+
+To Configure your partitioner use the option `createPartitioner`.
+
+```javascript
+kafka.producer({ createPartitioner: MyPartitioner })
+```
+
+#### <a name="producing-messages-retry"></a> Retry
+
+The option `retry` can be used to customize the configurations for the producer.
```
matthiasfeist: configuration: singular not configuration**s**

README.md
```diff
@@ -46,7 +249,48 @@ async () => {
...
+
+Take a look at [Retry](#setup-client-default-retry) for more information.
+
+### <a name="consuming-messages"></a> Consuming messages from Kafka
+
+KafkaJS only supports consumer groups.
```
matthiasfeist: and not ... what else is there?

README.md
```diff
@@ -46,7 +249,48 @@ async () => {
...
+
+Consumer groups allow a group of machines or processes to coordinate access to a list of topics, distributing the load among the consumers. When a consumer fails the load is automatically distributed to other members of the group. Consumer groups must have unique group ids.
```
matthiasfeist: and they also must be unique per topic right? wasn't that one of the problems I had?

README.md
```diff
@@ -93,35 +337,36 @@ await consumer.run({
         }
       })
 
-      resolveOffset(message.offset)
+      await resolveOffset(message.offset)
```
matthiasfeist: this comment is actually about the code further up (but Github doesn't let me comment on it).
I think a bit more documentation on the difference between eachMessage and eachBatch would be great.
Also the options in eachBatch such as highWatermark are not clear. I see that this is more a problem of the Kafka docs and not KafkaJS but maybe a link to the right place would be great.

README.md
```diff
@@ -93,35 +337,36 @@ await consumer.run({
...
+- __sessionTimeout__ - Timeout in milliseconds used to detect failures. The consumer sends periodic heartbeats to indicate its liveness to the broker. If no heartbeats are received by the broker before the expiration of this session timeout, then the broker will remove this consumer from the group and initiate a rebalance. default: `30000`
+- __heartbeatInterval__ - The expected time in milliseconds between heartbeats to the consumer coordinator. Heartbeats are used to ensure that the consumer's session stays active. The value must be set lower than session timeout. default: `3000`
+- __maxBytesPerPartition__ - The maximum amount of data per-partition the server will return. This size must be at least as large as the maximum message size the server allows or else it is possible for the producer to send messages larger than the consumer can fetch. If that happens, the consumer can get stuck trying to fetch a large message on a certain partition. default: `1048576` (1MB)
+- __minBytes__ - Minimum amount of data the server should return for a fetch request, otherwise wait up to `maxWaitTimeInMs` for more data to accumulate. default: `1`
+- __maxBytes__ - default: `10485760` (10MB)
```
matthiasfeist: no explanation on what maxBytes does?

README.md
```diff
@@ -46,7 +249,48 @@ async () => {
...
+    return 0
+  }
+}
+```
+
+`partitionMetadata` is an array of partitions with the following structure:
+
+```javascript
+[
+  { partitionId: 1, leader: 1 },
```
Nevon: Maybe include the signature of this. I would presume that `leader` is the node id of a broker?

README.md
```diff
@@ -93,35 +337,36 @@ await consumer.run({
...
-  sasl: {
-    mechanism: 'plain',
-    username: 'my-username',
-    password: 'my-password',
-  },
-})
-```
+- __createPartitionAssigner__ - default: `round robin`
+- __sessionTimeout__ - Timeout in milliseconds used to detect failures. The consumer sends periodic heartbeats to indicate its liveness to the broker. If no heartbeats are received by the broker before the expiration of this session timeout, then the broker will remove this consumer from the group and initiate a rebalance. default: `30000`
+- __heartbeatInterval__ - The expected time in milliseconds between heartbeats to the consumer coordinator. Heartbeats are used to ensure that the consumer's session stays active. The value must be set lower than session timeout. default: `3000`
```
Nevon: It would be good to include information on how to choose these values.

README.md
tulios: done

README.md
tulios: done

README.md
tulios: done

README.md
tulios: done

README.md
tulios: done

README.md
tulios: I've changed the text a bit, I think it's better

README.md
tulios: done

README.md
```diff
@@ -3,30 +3,233 @@
...
+  // console.log(`${level} [${namespace}] ${message} ${JSON.stringify(others)}`)
+}
+```
+
+Example using [winston](https://github.com/winstonjs/winston):
+
+```javascript
+const winstom = require('winston')
+const toWinstomLogLevel = level => switch(level) {
+  case 'NOTHING':
```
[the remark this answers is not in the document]
tulios: So the enum actually maps to numbers, in the end, I chose to expose the "label" to the creator because it allows the user to integrate with other libraries without knowing what is a 5 or a 4.

README.md
tulios: done

README.md
tulios: Maybe on the next write up, I want to change the defaults as well. WDYT?

PR #26, opened 2018-01-08
WIP: Lag based assigner

### LINKED ISSUE #25: Assign partitions based on lag
In cases where the offset lag is unevenly spread across partitions, it makes sense to assign partitions across consumers based on the current offset lag, rather than completely round robin.

Essentially:

1. **For each topic**
1.1. Calculate the offset lag for each partition in the topic
1.2. Sort the partitions by offset lag in descending order
1.3. **For each partition**
1.3.1. Assign the partition to the consumer with the lowest number of assigned partitions (across all topics)
1.3.2. If two or more consumers have the same number of assigned partitions, assign the partition to the one where the sum of the offset lag for the assigned partitions is the lowest

Nevon:
This is an initial stab at creating a lag based assigner to be used when the lag is not evenly distributed across partitions. It should help recover from cases where one consumer gets stuck and builds up a large amount of lag, by distributing the backed up partitions over the available consumers.

In 778cc68 I added it as the default assigner, just to be able to run all the tests using it. That should likely be reverted before any potential merge, as the round-robin assigner may be fine for most cases.

I'm also not exposing the assigner publicly in any way at the moment.

The distribution algorithm seems to work, in that it produces balanced assignments with the lag being evenly distributed, but the way the offset lag is fetched and calculated doesn't adequately handle errors, and it seems like it's causing problems in some of the test cases (hence the WIP).

Fixes #25


src/consumer/assigners/lagBasedAssigner/index.js
```diff
@@ -0,0 +1,136 @@
...
+    )
+
+    return {
+      topic,
+      partitions: consumerOffsetForPartititions.map(
+        ({ partition: consumerOffsetPartition, offset: startingOffset }) => ({
+          partition: consumerOffsetPartition,
+          offsetLag:
+            parseInt(
+              latestOffsetsByPartition.find(
```
tulios: Let's extract this to a function to make it easier to read

tulios:
It should work now, let's test this branch in production before we merge and do further improvements

tulios:
We need to update this PR based on the changes of #27

Nevon:
Yes. I am updating it now.

Nevon:
Closing this as it's become stale, and we don't currently have a need for it nor time to complete it. If the need arises, we can revive this PR.

tulios:
https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=71011801

PR #27, opened 2018-01-13
Fix JoinGroup assigner protocol

tulios:
When calling the `JoinGroup` request the consumers must send a list of available assigners (`groupProtocols`), kafkajs was mistakenly sending "default". Another issue was the lack of metadata for the assigner, in this case, version and the list of topics the consumer is subscribed. The low-level protocol implementation was always sending version 0 and an empty object as metadata.

This fix also changes the signature of the assigner, it should now contain two attributes: name and version; and two functions: assign and protocol. Example:

```javascript
const MyAssigner = ({ cluster }) => ({
  name: 'MyAssigner',
  version: 1,
  assign({ members, topics }) {},
  protocol({ topics }) {}
})
```


Nevon:
One interesting comment from the kafka client-side assignment proposal is this: 

> The coordinator only returns a protocol which is supported by all members. The disagreement that I was referring to is just disagreement about ordering (which is used to convey preference). For example, if one member provides ["A", "B"], and another provides ["B", "A"], then the coordinator will choose between "A" and "B" randomly. However, if the both provide ["A", "B"], then the coordinator will chose "A."

I assumed that the coordinator would have some more advanced selection, like maybe respecting the order of the leader, or picking the one that is preferred by most members. It would be interesting to see if this comment is actually correct.

tulios:
Yes, we can act on the JoinGroup response in another PR

PR #31, opened 2018-01-19
Add experimental describe group

tulios:
This PR adds a `describeGroup` API to the consumer, allowing the consumer to query the status of its own group.

```javascript
async () => {
  const data = await consumer.describeGroup()
  // {
  //  errorCode: 0,
  //  groupId: 'consumer-group-id-f104efb0e1044702e5f6',
  //  members: [
  //    {
  //      clientHost: '/172.19.0.1',
  //      clientId: 'test-3e93246fe1f4efa7380a',
  //      memberAssignment: Buffer,
  //      memberId: 'test-3e93246fe1f4efa7380a-ff87d06d-5c87-49b8-a1f1-c4f8e3ffe7eb',
  //      memberMetadata: Buffer,
  //    },
  //  ],
  //  protocol: 'RoundRobinAssigner',
  //  protocolType: 'consumer',
  //  state: 'Stable',
  // },
}
```


src/consumer/index.js
```diff
@@ -172,12 +173,39 @@ module.exports = ({
     consumerGroup.seek({ topic, partition, offset })
   }
 
+  /**
+   * @returns Promise<GroupDescription>
```
Nevon: I like this a lot. We should do this for all places where we're currently just returning `Promise`.

PR #33, opened 2018-01-20
Fix member assignment protocol

tulios:
This is a continuation of the work done in PR #27. Kafka documentation doesn't mention anything about the member assignment field, it's just a byte field and the clients can transfer data in any way. KafkaJS was doing that and everything was working fine, however, some monitoring tools weren't able to properly assess the health of the consumer groups. Apparently, there is an implicit protocol that many clients and monitoring tools expect, I believe this was based on the java client and became a rule.

This PR changes the member assignment protocol to be compliant with this implicit protocol. The `memberAssignment` field now contains an encoded message with the following attributes: __version__, __assignment__, and __userData__. Example:

```javascript
{
  version: 1,
  assignment: { 'topic-test': [2, 5, 4, 1, 3, 0] },
  userData: Buffer.alloc(0)
}
```

`userData` won't be used now, thus, is set to an empty buffer (it's a binary field).

For example, when running [KastleX](https://github.com/klarna/kastlex)

__before:__
screen shot 2018-01-20 at 20 05 17

__now:__
screen shot 2018-01-20 at 20 05 32

This code is compatible with previous versions of the client. A rolling update should work just fine.


tulios:
I have also improved some test infrastructure to have more stable tests in Travis.
Highlights:

- Update Jest to the latest version (there were some bugs in the version we were using)
- Prevent unhandled rejection when using the event emitter
- Reduce the number of workers when running on Travis (this greatly improved the performance)
- Stop waiting for consumers when running with `NODE_ENV=test`, some timeouts were caused by long waits
- Fix some consumers tests without `subscribe` or `connected`. Some tests were relying on the fact the consumer will reconnect if not connected
- Use log level `debug` if the `VERBOSE` env is in place, this allows us to better debug issues with  Travis (still not perfect)

tulios:
I've also emailed Travis to see if we can have access to the "Debug Build" button. This would help us improve the stability of our tests in the CI environment since they work fine locally.

PR #38, opened 2018-02-23
Only retry failed brokers

### LINKED ISSUE #22: Don't retry on all brokers when producing
https://github.com/tulios/kafkajs/blob/3804af9232aadc212876024ea27a372d0e000124/src/producer/index.js#L42-L48

If one fails, we currently retry all, which could lead to duplicate messages.

https://github.com/tulios/kafkajs/blob/3804af9232aadc212876024ea27a372d0e000124/src/producer/sendMessages.js#L27-L35

Instead, we should only retry the ones that fail.

tulios:
This PR fixes #22 

This is hard to simulate with integration tests, so I've written a regular unit test with a lot of mocks (which I don't like, but it's the best solution for now)


src/producer/sendMessages.js
```diff
@@ -1,36 +1,51 @@
...
-    const broker = await cluster.findBroker({ nodeId })
-    const response = await broker.produce({ acks, timeout, compression, topicData })
-    return responseSerializer(response)
-  })
-
-  return flatten(await Promise.all(requests))
+const { keys } = Object
+
+module.exports = ({ logger, cluster, partitioner }) => {
+  const retrier = createRetry()
```
Nevon: Make this configurable.

Nevon: Or not, since the high-level retry will retry with fetching new metadata etc, so if this is configured to a large number, the total number of retries will be very large.

tulios: We default to 5 retries; maybe we should at least hardcode the number of retries for now. WDYT?

PR #41, opened 2018-03-09
Add consumer pause & resume

Nevon:
This also fixes a bug where if you specify `0` as the partition for a message, the partitioner would fall back to using the hash of the key, as it would consider `message.partition` falsy.

### DIFF
--- a/src/consumer/consumerGroup.js
+++ b/src/consumer/consumerGroup.js
@@ -1,4 +1,5 @@
 const flatten = require('../utils/flatten')
+const sleep = require('../utils/sleep')
 const OffsetManager = require('./offsetManager')
 const Batch = require('./batch')
 const SeekOffsets = require('./seekOffsets')
@@ -219,6 +220,17 @@ module.exports = class ConsumerGroup {
 
       const pausedTopics = this.subscriptionState.paused()
       const activeTopics = topics.filter(topic => !pausedTopics.includes(topic))
+
+      if (activeTopics.length === 0) {
+        this.logger.debug(`No active topics, sleeping for ${this.maxWaitTime}ms`, {
+          topics,
+          activeTopics,
+          pausedTopics,
+        })
+
+        await sleep(this.maxWaitTime)
+        return []
+      }
 
       this.logger.debug(`Fetching from ${activeTopics.length} out of ${topics.length} topics`, {
         topics,
--- a/src/consumer/consumerGroup.js
+++ b/src/consumer/consumerGroup.js
@@ -2,6 +2,7 @@ const flatten = require('../utils/flatten')
 const OffsetManager = require('./offsetManager')
 const Batch = require('./batch')
 const SeekOffsets = require('./seekOffsets')
+const SubscriptionState = require('./subscriptionState')
 const { KafkaJSError, KafkaJSNonRetriableError } = require('../errors')
 const { HEARTBEAT } = require('./instrumentationEvents')
 const { MemberAssignment } = require('./assignerProtocol')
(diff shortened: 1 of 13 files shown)


soroushhakami:
👏

tulios:
Looks good, I would just mention in the readme that the topic won't be fetched in the next cycle, you might get messages from the paused topic in the current batch.

tulios:
You can also add the signature of the methods in the readme and mention that the library currently doesn't support specific partitions.

Nevon:
This works fine as long as you are subscribed to multiple topics, and don't pause all of them. If you do, however, it looks like the fetch loop becomes synchronous, so nothing else ever gets a chance to execute.

Nevon:
@tulios: 🏖💻🍹

src/consumer/consumerGroup.js
```diff
@@ -160,6 +162,24 @@ module.exports = class ConsumerGroup {
...
+  }
+
+  resume(topics) {
+    this.logger.info(`Resuming fetching from ${topics.length} topics`, {
+      topics,
+    })
+    this.subscriptionState.resume(topics)
+  }
+
+  paused() {
```
joaomilho: small stuff, but wouldn't a `get paused()` be a nicer API?

src/consumer/subscriptionState.js
```diff
@@ -0,0 +1,26 @@
+module.exports = class SubscriptionState {
+  constructor() {
+    this.pausedTopics = new Set()
```
joaomilho: fancy

src/producer/partitioners/default/index.js
```diff
@@ -26,7 +26,7 @@ module.exports = () => {
     const availablePartitions = partitionMetadata.filter(p => p.leader >= 0)
     const numAvailablePartitions = availablePartitions.length
 
-    if (message.partition) {
+    if (message.partition !== null && message.partition !== undefined) {
```
joaomilho: why? Is `false` a valid partition?

Nevon: No, but `0` is.

src/consumer/subscriptionState.js
Nevon: Don't want any of those filthy duplicates in here. 💅

Nevon:
This is as fascinating as it is terrifying. I'm running my consumer and pausing/resuming by calling an HTTP endpoint. If I pause consuming of all topics, KafkaJS goes into a busy loop that won't let me call the endpoint to resume consuming.

The culprit appears to be: https://github.com/tulios/kafkajs/blob/b474e53ad5d08fdab216702c57b1bf013edd70d8/src/consumer/consumerGroup.js#L265 in `ConsumerGroup#fetch`. `requests` will be an empty array, so my thinking was that maybe `await Promise.all([])` resolves synchronously or something (doesn't make any sense, but nothing makes sense in this world). So I tried changing it to:

```javascript
const results = await Promise.all(requests.length > 0 ? requests : [new Promise(resolve => process.nextTick(() => resolve([])))])
```

This did not help. At this point I _know_ that it resolves asynchronously, but still, express never got a chance to do anything.

So, let's try with `setTimeout`:

```javascript
const results = await Promise.all(requests.length > 0 ? requests : [new Promise(resolve => setTimeout(() => resolve([]), 0))])
```

This works.

At this point someone needs to explain to me how the event loop works, because clearly it does not work the way I think it does.

Nevon:
I'm still confused, but at least I now know why `process.nextTick` didn't work. https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/#process-nexttick

Scheduling fetch using `setImmediate` makes more sense.

Nevon:
I've been running this for a while now, and it seems to be working fine. However, if you're running with a loglevel >= info, you'll produce an insane amount of logs, as each fetch loop logs, and without any topics to fetch from, it will basically produce logs as fast as it can.

I think it makes sense to add some throttling to the fetch loop, at least when you're not subscribed to any topic.

PR #44, opened 2018-04-04
Allow auto-resolve to be disabled for eachBatch

### LINKED ISSUE #18: How to shutdown consumer when using eachBatch
The `eachBatch` callback offers a more advanced use of the consumer, the current implementation has a drawback which prevents it to be fully used. Currently, the offset of the last message will be resolved after the callback runs, but when the consumer is shutting down there is no way to stop consuming the batch without committing the last offset, so the consumer has to wait for the full batch to be consumed before it can shut down, which might take a while and get the consumer killed by the OS, Docker, etc.

This callback already receives `isRunning` and `resolveOffset` but it doesn't offer any mechanisms to prevent the last offset from being committed.

Suggestions:

1) Expose a function which should be called for each message, this function will interrupt the function and allow the consumer to commit the last resolved offset instead of the last offset in the batch.

```javascript
function eachBatch({ batch, resolveOffset, stopIfNotRunning }) {
  for (let message of batch.messages) {
    await stopIfNotRunning()
    await processMessage(message)
    await resolveOffset(message.offset)
  }
}
```

2) Expose a function to register a function to run after `eachBatch` is complete

```javascript
function eachBatch({ batch, resolveOffset, after }) {
  try {
    for (let message of batch.messages) {
      await processMessage(message)
      await resolveOffset(message.offset)
    }
  } catch (e) {
    after(() => {
      // just log
      // or resolveOffset(batch.lastOffset())
    })
  }
}
```

3) Add an option to disable the auto-resolve from `eachBatch`

```javascript
consumer.run({
  eachBatchAutoResolve: false,
  eachBatch: ({ batch, resolveOffset, isRunning }) => {
    try {
      for (let message of batch.messages) {
        if (!isRunning()) break
        await processMessage(message)
        await resolveOffset(message.offset)
      }
    } catch (e) {
      if (e instanceof SkipToTheLastOffsetOfTheBatchError) {
        await resolveOffset(batch.lastOffset())
      }
    }
  }
})
```

tulios:
This is the last piece missing to allow `eachBatch` to have full control over the message processing.

```javascript
consumer.run({
  eachBatchAutoResolve: false,
  eachBatch: ({ batch, resolveOffset, isRunning }) => {
    try {
      for (let message of batch.messages) {
        if (!isRunning()) break
        await processMessage(message)
        await resolveOffset(message.offset)
      }
    } catch (e) {
      // messages won't be automatically resolved and committed
    }
  }
})
```

This PR resolves #18

### DIFF
--- a/src/consumer/__tests__/runner.spec.js
+++ b/src/consumer/__tests__/runner.spec.js
@@ -69,7 +69,7 @@ describe('Consumer > Runner', () => {
       consumerGroup.fetch.mockImplementationOnce(() => [batch])
       await runner.start()
       expect(onCrash).not.toHaveBeenCalled()
-      expect(consumerGroup.resolveOffset).not.toHaveBeenCalledWith({ topic, partition, offset: 4 })
+      expect(consumerGroup.resolveOffset).not.toHaveBeenCalled()
     })
   })
 
--- a/src/consumer/__tests__/runner.spec.js
+++ b/src/consumer/__tests__/runner.spec.js
@@ -1,4 +1,5 @@
 const Runner = require('../runner')
+const Batch = require('../batch')
 const { KafkaJSProtocolError } = require('../../errors')
 const { createErrorFromCode } = require('../../protocol/error')
 const { newLogger } = require('testHelpers')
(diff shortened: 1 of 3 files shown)


tulios:
The readme won't be updated in this PR, after releasing `v0.8.0` I'll focus on getting the readme ready for `v1.0.0`

src/consumer/__tests__/runner.spec.js
```diff
@@ -34,6 +42,37 @@ describe('Consumer > Runner', () => {
...
+      const batch = new Batch(topic, {
+        partition,
+        highWatermark: 5,
+        messages: [{ offset: 4, key: '1', value: '2' }],
+      })
+
+      consumerGroup.fetch.mockImplementationOnce(() => [batch])
+      await runner.start()
+      expect(onCrash).not.toHaveBeenCalled()
+      expect(consumerGroup.resolveOffset).not.toHaveBeenCalledWith({ topic, partition, offset: 4 })
```
Nevon: Just check that it hasn't been called at all, since we're not testing offset calculation here, and don't want this to accidentally pass.

Issue #65, opened 2018-05-27
Create Topic


al66:
Is it planned to add a method to create new topics? 
( running kafka with auto.create.topics.enable )

tulios:
Hi, it's planned. The effort now is to enable native support to Kafka 0.11.0+, but we plan to add all the admin operations later on. If you run Kafka with auto create it should work.

al66:
Yes it works, but throws an error [LeaderNot Available] during first commit.
But sounds good, that it is planned!

tulios:
It's possible to create topics with KafkaJS now, this feature should be in the next version (`v1.1.0`), I have a couple more things to add to this version, but it should be out soon.

PR #73, opened 2018-06-11
Add create topic

tulios:
closes issue #65 

```javascript
const kafka = new Kafka(...)
const admin = kafka.admin()

try {
  await admin.connect()
  await admin.createTopics({
    validateOnly: <boolean>,
    waitForLeaders: <boolean>
    timeout: <Number>,
    topics: <Topic[]>,
  })
} finally {
  await admin.disconnect()
}
```


tulios:
@Nevon I will go ahead and merge this, but please take your time and review the PR. I can apply the fixes later. I want to get some other things into the next version before I resume my work on the Kafka 0.11 integration.

This PR made the tests quite stable now; they are also faster.

tulios:
@Nevon the master build was green on the first try; I think it's a sign of the new times :D

src/admin/index.js
```diff
@@ -0,0 +1,97 @@
+const createRetry = require('../retry')
+const waitFor = require('../utils/waitFor')
+const { KafkaJSNonRetriableError } = require('../errors')
+
+const retryOnLeaderNotAvailable = (fn, opts = {}) => {
+  const callback = async () => {
+    try {
+      return await fn()
+    } catch (e) {
+      console.error(e)
```
Nevon: Tsk tsk tsk

src/cluster/index.js
```diff
@@ -102,6 +104,29 @@ module.exports = class Cluster {
...
 
+  /**
+   * @public
+   * @returns {Promise<Broker>}
+   */
+  async findControllerBroker() {
+    const { metadata } = this.brokerPool
+
+    // controllerId is an int32, it's safe to cast to Number
+    if (!metadata || Number.isNaN(Number(metadata.controllerId))) {
```
Nevon: I would change this to `metadata.controllerId == null` instead. It's true that since it's an int32 it is safe to cast it, but it feels really wrong when all you really want to do is check if it's set. `== null` (note the loose comparison) will be true if it's undefined or null, but false if it's 0.

src/admin/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+        return true
+      } catch (e) {
+        if (e.type === 'NOT_CONTROLLER') {
+          logger.warn('Could not create topics', { error: e.message, retryCount, retryTime })
+          throw e
+        }
+
+        if (e.type === 'TOPIC_ALREADY_EXISTS') {
+          logger.warn(e.message)
```
Nevon: Should we really warn about this? Given that we already return the boolean indicating that the topic already exists, this should at the very least not be a warning, maybe an info.

README.md
```diff
@@ -635,6 +637,55 @@ const data = await consumer.describeGroup()
...
+  replicationFactor: <Number>, // default: 1
+  replicaAssignment: <Array>,  // Example: [{ partition: 0, replicas: [0,1,2] }] - default: []
+  configEntries: <Array>       // Example: [{ name: 'cleanup.policy', value: 'compact' }] - default: []
+}
+```
+
+| property       | description | default |
+| -------------- | ----------- | ------- |
+| topics         | Topic definition | |
+| validateOnly   | If this is `true`, the request will be validated, but the topic won't be created. | false |
```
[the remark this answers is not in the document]
tulios: This is a feature of Kafka create topics, so it's exposed here. I believe this will be useful when creating topics with unusual configurations, such as compacted topics, or specialized replication schemes. I'm mainly making it feature complete.

src/admin/index.js
tulios: Fixed here 3a9abc64dcf31d44ab98cfb2b4969f1ec50ea17c

src/admin/index.js
tulios: Fixed here 3a9abc64dcf31d44ab98cfb2b4969f1ec50ea17c

src/cluster/index.js
tulios: Fixed here 31fdac162bab810f829b0dd392e3641d23727d20

Issue #80, opened 2018-06-26
Handle parallel calls during SCRAM authentication flow


Nevon:
If you are producing to multiple topics in parallel, the broker gets confused during the SCRAM authentication flow. From the broker perspective, it looks something like this:

1. Client -> I want to authenticate as Nevon
2. Broker <- Sure, send me the proof
3. Client -> I want to authenticate as Nevon
4. Broker <- Uh... What?
5. Client -> Here's the proof!
6. Broker <- ... Are you okay?

When the client is in the SCRAM authentication flow, we should set some kind of flag so that we don't start it again if another request causes us to try to authenticate again.

PR #81, opened 2018-06-27
Support concurrent broker connect

tulios:
@carlrygart

fixes issue #80

### DIFF
--- a/src/broker/index.js
+++ b/src/broker/index.js
@@ -12,14 +12,23 @@ const SASLAuthenticator = require('./saslAuthenticator')
  * @param {Object} logger
  * @param {Object} [versions=null] The object with all available versions and APIs
  *                                 supported by this cluster. The output of broker#apiVersions
+ * @param {number} [authenticationTimeout=1000]
  */
 module.exports = class Broker {
-  constructor({ connection, logger, allowExperimentalV011, nodeId = null, versions = null }) {
+  constructor({
+    connection,
+    logger,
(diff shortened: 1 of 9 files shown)


src/utils/lock.js
```diff
@@ -0,0 +1,40 @@
...
+          clearTimeout(timeoutId)
+          this.emitter.removeListener('releaseLock', tryToAcquire)
+          return resolve()
+        }
+      }
+
+      this.emitter.on('releaseLock', tryToAcquire)
+      timeoutId = setTimeout(
+        () => reject(new KafkaJSError('Timeout while acquiring lock')),
+        this.timeout
```
Nevon: Maybe this should be exposed as a configuration, since the expected timeout could be different between clusters.

src/broker/index.js
```diff
@@ -39,21 +41,31 @@ module.exports = class Broker {
...
-    this.authenticated = false
-    await this.connection.connect()
+    try {
+      await this.lock.acquire()
 
-    if (!this.versions) {
-      this.versions = await this.apiVersions()
-    }
+      if (this.isConnected()) {
+        return false
```
Nevon: I don't think I agree with this. I don't think anyone outside of the broker should be aware of whether you connected or if you became connected because some other request caused you to become connected.

src/utils/lock.js
tulios: a3a9893

Issue #79, opened 2018-06-26
Produce to multiple topics at once


Nevon:
Currently you can produce to only a single topic, but in case you want to produce to multiple topics, you currently have to make multiple calls.

Instead of doing this, we should provide the ability to produce to multiple topics with one call. For example:

```javascript
await producer.sendBatch([{
  topic: 'foo',
  messages: [ ... ]
},
{
  topic: 'bar',
  messages: [ ... ]
}])
```

Where `sendBatch` would take an array of topic-messages (same object that `Producer.send` currently receives).

Nevon:
Fixed by #82

PR #82, opened 2018-06-28
Add `sendBatch` method to producer

Nevon:
Allows the producer to produce messages to multiple topics at the same time.

Example:

```javascript
await producer.sendBatch([
  {
    topic: 'topic-name',
    messages: [
      { key: 'key1', value: 'hello world' },
      { key: 'key2', value: 'hey hey!' }
    ],
  },
  {
    topic: 'topic-name-2',
    messages: [
      { key: 'key1', value: 'foo bar' },
    ],
  }
])
```

Co-authored-by: Túlio Ornelas <ornelas.tulio@gmail.com>

@carlrygart

### DIFF
--- a/src/producer/index.js
+++ b/src/producer/index.js
@@ -14,6 +14,66 @@ module.exports = ({
   const sendMessages = createSendMessages({ cluster, partitioner })
   const logger = rootLogger.namespace('Producer')
 
+  const sendBatch = async ({ acks, timeout, compression, topicMessages }) => {
+    if (topicMessages.some(({ topic }) => !topic)) {
+      throw new KafkaJSNonRetriableError(`Invalid topic`)
+    }
+
+    for (let { topic, messages } of topicMessages) {
+      if (!messages) {
+        throw new KafkaJSNonRetriableError(
+          `Invalid messages array [${messages}] for topic "${topic}"`
+        )
+      }
+    }
+
+    return retrier(async (bail, retryCount, retryTime) => {
+      try {
+        return await sendMessages({
+          acks,
+          timeout,
+          compression,
+          topicMessages,
+        })
+      } catch (error) {
+        if (!cluster.isConnected()) {
+          logger.debug(`Cluster has disconnected, reconnecting: ${error.message}`, {
+            retryCount,
+            retryTime,
+          })
+          await cluster.connect()
+          await cluster.refreshMetadata()
+          throw error
+        }
+
+        // This is necessary in case the metadata is stale and the number of partitions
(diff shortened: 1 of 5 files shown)


tulios:
Fixes issue #79

src/producer/index.js
```diff
@@ -14,6 +14,66 @@ module.exports = ({
   const sendMessages = createSendMessages({ cluster, partitioner })
   const logger = rootLogger.namespace('Producer')
 
+  const sendBatch = async ({ acks, timeout, compression, topicMessages }) => {
+    if (topicMessages.some(({ topic }) => !topic)) {
```
tulios: Let's validate that `topicMessages` is not empty

src/producer/index.js
```diff
@@ -26,61 +86,40 @@ module.exports = ({
     disconnect: async () => await cluster.disconnect(),
 
     /**
-     * @param {string} topic
-     * @param {Array} messages An array of objects with "key" and "value", example:
+     * @param {ProduceRequest} ProduceRequest
+     * @returns {Promise}
+     *
+     * @typedef {Object} ProduceRequest
```
tulios: Here it should be `TopicMessage`, right?

Nevon: No. A `TopicMessage` is an object with a string `topic` and an array of messages. The signature for `send` is different, in order to not break backwards compatibility. For `send`, the object also contains `acks`, `timeout` and `compression`.

src/producer/index.js
Nevon: c231143

src/producer/index.js
```diff
@@ -14,8 +14,26 @@ module.exports = ({
...
+   *                            1 = only waits for the leader to acknowledge
+   * @property {number} [timeout=30000] The time to await a response in ms
+   * @property {Compression.Types} [compression=Compression.Types.None] Compression codec
+   *
+   * @param {SendBatchRequest}
+   * @returns {Promise}
+   */
   const sendBatch = async ({ acks, timeout, compression, topicMessages }) => {
-    if (topicMessages.some(({ topic }) => !topic)) {
+    if (topicMessages && topicMessages.length > 0 && topicMessages.some(({ topic }) => !topic)) {
```
tulios: What happens if the user sends an empty array of `topicMessages`? My point with the previous comment was to validade if the array is empty, if we do that we can keep this check simpler

Nevon: Clearly I should not write code before having had coffee. Not sure what I was thinking. 🙈 a5a8f12

PR #89, opened 2018-07-10
Add consumer auto commit policies

tulios:
```javascript
consumer.run({
  autoCommitInterval: 5000,
  // ...
})

consumer.run({
  autoCommitThreshold: 100,
  // ...
})
```


README.md
```diff
@@ -35,6 +35,7 @@ KafkaJS is battle-tested and ready for production.
     - [Other](#producing-messages-compression-other)
 - [Consuming messages](#consuming-messages)
   - [eachMessage](#consuming-messages-each-message)
+  - [autoCommit](#autoCommit)
```
Nevon: This should be `consuming-messages-auto-commit`

README.md
```diff
@@ -448,9 +449,33 @@ async () => {
 }
 ```
 
+### <a name="consuming-messages-auto-commit"></a> autoCommit
+
+The messages are always fetched in batches from Kafka, even when using the `eachMessage` handler. All resolved offsets will be committed to Kafka after processing the whole batch. Auto-commit offers more flexibility when committing offsets; there are two flavors available:
```
Nevon: Maybe include something about the use case here. Something like:

> Committing offsets periodically during a batch allows the consumer to recover from group rebalances, stale metadata and other issues before it has completed the entire batch. However, committing more often increases network traffic and slows down processing.

src/consumer/offsetManager/index.js
```diff
@@ -140,6 +173,21 @@ module.exports = class OffsetManager {
...
+
+    const timeoutReached =
+      this.autoCommitInterval != null && now >= this.lastCommit + this.autoCommitInterval
+
+    const thresholdReached =
+      this.autoCommitThreshold != null &&
+      this.countResolvedOffsets().gte(Long.fromValue(this.autoCommitThreshold))
+
+    if (timeoutReached || thresholdReached) {
+      return this.commitOffsets()
```
Nevon: If `autoCommitThreshold` and `autoCommitInterval` are both undefined, `commitOffsetsIfNecessary` will never commit offsets. That feels unintuitive, but I'm not sure if the alternative is better...

src/consumer/offsetManager/index.js
```diff
@@ -85,6 +92,32 @@ module.exports = class OffsetManager {
...
 
+  /**
+   * @returns {Long}
+   */
+  countResolvedOffsets() {
+    const toPartitions = topic => keys(this.resolvedOffsets[topic])
+
+    const subtractOffsets = (resolvedOffset, committedOffset) => {
+      const resolvedOffsetLong = Long.fromValue(resolvedOffset)
+      return committedOffset === '-1'
```
Nevon: Is this right? I'm assuming this means that we have not committed any offsets, but `clearOffsets` will actually make  `this.committedOffsets[topic][partition]` undefined, right, so this would not be true.

tulios: Before each `fetch` the consumer group will first resolve the offsets
https://github.com/tulios/kafkajs/blob/e65522ede62f3f8c0cf65ecabf715968ed389474/src/consumer/consumerGroup.js#L243

In `offsetManager#resolveOffsets` any invalid offset will be fetched from Kafka (`undefined`)
https://github.com/tulios/kafkajs/blob/e65522ede62f3f8c0cf65ecabf715968ed389474/src/consumer/offsetManager/index.js#L204

so `countResolvedOffsets` should be fine

src/consumer/offsetManager/index.js
tulios: I agree but my understanding of `necessary` in this functions is defined the by policies, so if it's not configured then it's not necessary. Like we discussed earlier, we could eventually expose `commitOffsets`.

PR #92, opened 2018-07-18
Throw when creating misconfigured consumer

### LINKED ISSUE #91: Notify user when setting heartbeat interval to same or higher than session timeout
A user was seeing constant rebalancing because they had set `heartbeatInterval` to the same value as `sessionTimeout`.

If we see that `heartbeatInterval` is the same or higher than `sessionTimeout`, we should notify the user somehow.

Nevon:
Fixes #91


src/consumer/index.js
```diff
@@ -39,6 +39,12 @@ module.exports = ({
   let runner = null
   let consumerGroup = null
 
+  if (heartbeatInterval >= sessionTimeout) {
+    throw new KafkaJSNonRetriableError(
+      `Consumer heartbeatInterval (${heartbeatInterval}) must be lower than sessionTimeout (${sessionTimeout})`
```
tulios: Maybe change the text to be something like: "The heartbeat interval be at least 1/3 of the session timeout"

PR #94, opened 2018-07-25
Constantly refresh metatada

### LINKED ISSUE #93: Constantly refresh metadata for consumers and producers
Add support to metadata max age. From the Kafka docs:

> `metadata.max.age.ms` | The period of time in milliseconds after which we force a refresh of metadata even if we haven't seen any partition leadership changes to proactively discover any new brokers or partitions. | long | 300000

On producers:

```javascript
kafka.producer({ metadataMaxAgeInMs: 300000 })
```

On consumers:

```javascript
kafka.consumer({ groupId: 'my-group', metadataMaxAgeInMs: 300000 })
```

Kafka default is 5 minutes; we should keep the same.

tulios:
Fixes #93

### DIFF
--- a/src/consumer/assigners/roundRobinAssigner/index.js
+++ b/src/consumer/assigners/roundRobinAssigner/index.js
@@ -41,9 +41,6 @@ module.exports = ({ cluster }) => ({
       assignment[memberId] = {}
     })
 
-    // Make sure the assigner always work with up to date metadata
-    await cluster.refreshMetadataIfNecessary()
-
     topics.forEach(topic => {
       const partitionMetadata = cluster.findTopicPartitionMetadata(topic)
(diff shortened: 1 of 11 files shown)


src/consumer/assigners/roundRobinAssigner/index.js
```diff
@@ -41,6 +41,9 @@ module.exports = ({ cluster }) => ({
       assignment[memberId] = {}
     })
 
+    // Make sure the assigner always work with up to date metadata
+    await cluster.refreshMetadataIfNecessary()
```
Nevon: I think we should move this to the consumerGroup before calling the assigner, so that we don't need to add this to every assigner.

tulios: fe54b58

PR #100, opened 2018-07-30
Discard messages with a lower offset than requested

tulios:
Fetch can return message sets with a lower offset than the requested, for example, we fetch data from offset 77 but receive a message set starting with offset 51. This behavior happens due to how Kafka stores the message sets on disk. When seeking to different offsets than the natural flow (from the beginning or the end), it can lead to "wrong" batches.


src/consumer/__tests__/batch.spec.js
```diff
@@ -3,9 +3,27 @@ const Batch = require('../batch')
 describe('Consumer > Batch', () => {
   const topic = 'topic-name'
 
+  it('discards messages with a lower offset than the requested', () => {
+    const fetchedOffset = 3
```
Nevon: Not that it matters for these cases, but shouldn't `fetchedOffset` always be a string?

tulios: Everything will be converted into a long so, technically yes, but in the end it doesn't matter that much. Numbers highlight better in the test

PR #101, opened 2018-08-01
Add consumer group offset commands to the admin client

tulios:
This PR mainly adds 3 new methods to the admin client:

```javascript
await admin.fetchOffsets({ groupId, topic })

await admin.resetOffsets({ groupId, topic }) // latest by default
// await admin.resetOffsets({ groupId, topic, earliest: true })

await admin.setOffsets({
  groupId: 'my-consumer-group',
  topic: 'custom-topic',
  partitions: [
    { partition: 0, offset: '35' },
    { partition: 3, offset: '19' },
  ]
})
```

### DIFF
--- a/src/admin/__tests__/fetchOffsets.spec.js
+++ b/src/admin/__tests__/fetchOffsets.spec.js
@@ -1,0 +1,68 @@
+const createAdmin = require('../index')
+const { secureRandom, createCluster, newLogger, createTopic } = require('testHelpers')
+
+describe('Admin', () => {
+  let topicName, groupId, admin, consumer
+
+  beforeEach(async () => {
+    topicName = `test-topic-${secureRandom()}`
+    groupId = `consumer-group-id-${secureRandom()}`
+
+    await createTopic({ topic: topicName })
+  })
+
+  afterEach(async () => {
+    await admin.disconnect()
+    consumer && (await consumer.disconnect())
+  })
+
+  describe('fetchOffsets', () => {
+    test('throws an error if the groupId is invalid', async () => {
+      admin = createAdmin({ cluster: createCluster(), logger: newLogger() })
+      await expect(admin.fetchOffsets({ groupId: null })).rejects.toHaveProperty(
+        'message',
+        'Invalid groupId null'
+      )
+    })
+
+    test('throws an error if the topic name is not a valid string', async () => {
+      admin = createAdmin({ cluster: createCluster(), logger: newLogger() })
+      await expect(admin.fetchOffsets({ groupId: 'groupId', topic: null })).rejects.toHaveProperty(
+        'message',
+        'Invalid topic null'
+      )
+    })
+
+    test('returns unresolved consumer group offsets', async () => {
+      const cluster = createCluster()
+      admin = createAdmin({ cluster, logger: newLogger() })
+
+      await admin.connect()
+      const offsets = await admin.fetchOffsets({
+        groupId,
+        topic: topicName,
+      })
+
+      expect(offsets).toEqual([{ partition: 0, offset: '-1' }])
+    })
+
+    test('returns the current consumer group offset', async () => {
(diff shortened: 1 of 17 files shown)


src/admin/__tests__/setOffsets.spec.js
```diff
@@ -0,0 +1,85 @@
...
+        groupId,
+        topic: topicName,
+        partitions: [{ partition: 0, offset: 13 }],
+      })
+
+      const offsets = await admin.fetchOffsets({ groupId, topic: topicName })
+      expect(offsets).toEqual([{ partition: 0, offset: '13' }])
+    })
+
+    test('throws an error if the consumer group is runnig', async () => {
```
Nevon: Typo on "running"

src/admin/index.js
```diff
@@ -87,9 +99,123 @@ module.exports = ({ retry = { retries: 5 }, logger: rootLogger, cluster }) => {
...
+    const coordinator = await cluster.findGroupCoordinator({ groupId })
+    const partitionsToFetch = partitions.map(partition => ({ partition }))
+
+    const { responses } = await coordinator.offsetFetch({
+      groupId,
+      topics: [{ topic, partitions: partitionsToFetch }],
+    })
+
+    return responses
+      .filter(response => response.topic === topic)
```
Nevon: If the coordinator is unaware of the topic, what will we return here? I assume undefined, but maybe we should rather reject? Or will the offsetFetch already reject?

src/constants.js
```diff
@@ -0,0 +1,7 @@
+const EARLIEST_OFFSET = -2
```
Nevon: Ah, maybe this answers my question from before. Are these constants actually part of the protocol?

src/admin/__tests__/fetchOffsets.spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+    test('returns the current consumer group offset', async () => {
+      const cluster = createCluster()
+      admin = createAdmin({ cluster, logger: newLogger() })
+
+      await admin.connect()
+      await admin.setOffsets({
+        groupId,
+        topic: topicName,
+        partitions: [{ partition: 0, offset: 13 }],
```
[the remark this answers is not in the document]
tulios: By the time the consumer starts, it will get an out of range error and it will reassign the offset based on the default configuration (`fromBeginning`).

src/admin/__tests__/resetOffsets.spec.js
```diff
@@ -0,0 +1,107 @@
...
+        topic: topicName,
+        earliest: true,
+      })
+
+      const offsets = await admin.fetchOffsets({
+        groupId,
+        topic: topicName,
+      })
+
+      expect(offsets).toEqual([{ partition: 0, offset: '-2' }])
```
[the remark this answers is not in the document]
tulios: `-1` and `-2` are special offsets

src/admin/index.js
tulios: Good point, if the `metadata` request can't create the topic it might throw an error. I will investigate

src/constants.js
tulios: 👍

src/admin/index.js
tulios: It will throw `UNKNOWN_TOPIC_OR_PARTITION`

```javascript
{ 
  KafkaJSProtocolError: This server does not host this topic-partition
    at createErrorFromCode (./src/protocol/error.js:463:10)
    at Object.parse (./src/protocol/requests/metadata/v0/response.js:56:11)
    at Connection.send (./src/network/connection.js:249:35)
    at process._tickCallback (internal/process/next_tick.js:68:7)
  name: 'KafkaJSProtocolError',
  retriable: true,
  type: 'UNKNOWN_TOPIC_OR_PARTITION',
  code: 3
}
```

PR #108, opened 2018-08-10
Split tests when running on travis

tulios:


scripts/waitForKafka.js
```diff
@@ -63,7 +63,9 @@ waitForNode(kafka3ContainerId)
 
 console.log('\nAll nodes up:')
 console.log(
-  execa.shellSync(`HOST_IP=${process.env.HOST_IP} docker-compose ps`).stdout.toString('utf-8')
+  execa
+    .shellSync(`HOST_IP=${process.env.HOST_IP} docker-compose -f ${process.env.COMPOSE_FILE} ps`)
```
Nevon: This is missing `-p kafkajs`

Nevon: Hmm, that's strange. I actually do get all running containers back when running this without the `-p` flag, even though I launched them with it. When developing kafkajs-dlq, I encountered an issue where `docker-compose ps` wouldn't return any containers unless I specified the project name.

PR #109, opened 2018-08-10
WIP: Add TypeScript definitions

pthm:
This is still a work in progress, will be fleshing it out further as I implement more.


kafkajs.d.ts
```diff
@@ -0,0 +1,261 @@
...
+    password: string;
+}
+export interface RetryOptions {
+    maxRetryTime?: number;
+    initialRetryTime?: number;
+    factor?: number;
+    multiplier?: number;
+    retries?: number;
+}
+export enum ELogLevel {
```
Nevon: Is it a convention in TS to prefix enum types with `E`? If it's not, I would prefer to omit that and just go with `LogLevel`, `CompressionTypes` etc.

tulios:
Awesome @pthm!
I have the same questions regarding the enums, I never used TS before so I don't know the conventions of the language.

kafkajs.d.ts
pthm: It's a convention I follow, although not sure if that's pervasive in the wider TS community. In this instance I prefixed it to avoid conflicts with the exported const's

pthm:
Happy to add a TypeScript example although usage is largely the same as in JS.

In terms of keeping the typing in sync with the code-base we can try compiling some sort of "kitchen-sink" example although that example would also need to be updated to capture issues with any new APIs etc.

ianwsperber:
@pthm @tulios @Nevon Haven't seen any action on this in a while, do we need to keep it open?

tulios:
Thanks @pthm, we will get back to this soon.

PR #110, opened 2018-08-11
Add connection events for consumer, producer and admin

### LINKED ISSUE #107: Add instrumentation events for connect/disconnect and consumer stop
New proposed events:

`consumer.events.CONNECTED`
`consumer.events.DISCONNECTED`
`consumer.events.STOPPED`

`producer.events.CONNECTED`
`producer.events.DISCONNECTED`

`admin.events.CONNECTED`
`admin.events.DISCONNECTED`

Nevon:
Fixes #107


src/admin/index.js
```diff
@@ -1,9 +1,18 @@
...
+const InstrumentationEventEmitter = require('../instrumentation/emitter')
+const events = require('./instrumentationEvents')
+const { CONNECT, DISCONNECT } = require('./instrumentationEvents')
 const { LEVELS } = require('../loggers')
 const { KafkaJSNonRetriableError } = require('../errors')
 
+const { values, keys } = Object
+const eventNames = values(events)
+const eventKeys = keys(events)
+  .map(key => `consumer.events.${key}`)
```
tulios: Here it should be `admin` instead of consumer, right?

src/producer/index.js
```diff
@@ -1,8 +1,17 @@
...
+const InstrumentationEventEmitter = require('../instrumentation/emitter')
+const events = require('./instrumentationEvents')
+const { CONNECT, DISCONNECT } = require('./instrumentationEvents')
 const { KafkaJSNonRetriableError } = require('../errors')
 
+const { values, keys } = Object
+const eventNames = values(events)
+const eventKeys = keys(events)
+  .map(key => `consumer.events.${key}`)
+  .join(', ')
```
tulios: Same here, it should be `producer`

Nevon: And this, kids, is why we try not to copy & paste code 😅I was considering breaking out some of the shared code between consumer, producer and admin, but it felt a bit premature. I'll just fix the copypaste errors first.

tulios: :smile:

Nevon: Turns out the tests for this were passing for the wrong reason. `toThrow` only takes a single argument, which can be either a constructor, in which case it checks that the error is an instance of that, or a string or regex, in which case it matches on the error message. The tests were trying to do both, using `.toThrow(KafkaJSNonRetriableError, /blah blah consumer.events./)`, but since the regex was never taken into consideration, they were passing even if the event names were wrong.

Jest doesn't actually have a way of testing both, so for now I'm just testing the message, and not the error type. To test both, we'd have to write a custom matcher.

PR #115, opened 2018-08-16
Fix a bug when parsing multiple responses

bwdeng:
The previous implementation was assuming that the `rawData` only contains the bytes of a single response. This could cause response loss when mutiple reponses are concatenated into one `rawData`.

### DIFF
--- a/src/network/connection.js
+++ b/src/network/connection.js
@@ -304,43 +304,43 @@ module.exports = class Connection {
 
     this.buffer = Buffer.concat([this.buffer, rawData])
 
-    // Not enough bytes to read the expected response size, keep buffering
-    if (Buffer.byteLength(this.buffer) <= Decoder.int32Size()) {
-      return
+    // Process data if there are enough bytes to read the expected response size,
+    // otherwise keep buffering
+    while (Buffer.byteLength(this.buffer) > Decoder.int32Size()) {
+      const data = Buffer.from(this.buffer)
+      const decoder = new Decoder(data)
+      const expectedResponseSize = decoder.readInt32()
+
+      if (!decoder.canReadBytes(expectedResponseSize)) {
+        return
+      }
+
+      const response = new Decoder(decoder.readBytes(expectedResponseSize))
+      // Reset the buffer as the rest of the bytes
+      this.buffer = decoder.readAll()
+
+      if (this.authHandlers) {
+        return this.authHandlers.onSuccess(data.slice(0, this.offset + byteLength))
+      }
+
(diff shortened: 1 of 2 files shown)


tulios:
Thanks for the PR @bwdeng, have you faced this problem while using the library? We have been using KafkaJS for a while and I haven't seen this behavior.

src/network/connection.js
```diff
@@ -304,43 +304,43 @@ module.exports = class Connection {
...
-      return
-    }
+      const response = new Decoder(decoder.readBytes(expectedResponseSize))
+      // Reset the buffer as the rest of the bytes
+      this.buffer = decoder.readAll()
 
-    // The full payload is loaded, erase the temporary buffer
-    this.buffer = Buffer.alloc(0)
+      if (this.authHandlers) {
+        return this.authHandlers.onSuccess(data.slice(0, this.offset + byteLength))
```
tulios: `byteLength` isn't defined, this is breaking the tests

src/network/connection.js
```diff
@@ -304,43 +304,43 @@ module.exports = class Connection {
...
-    if (this.authHandlers) {
-      return this.authHandlers.onSuccess(data)
-    }
+      const correlationId = response.readInt32()
+      const payload = response.readAll()
 
-    const correlationId = decoder.readInt32()
-    const payload = decoder.readAll()
+      const entry = this.pendingQueue[correlationId]
+      delete this.pendingQueue[correlationId]
```
tulios: I wonder if we should remove the previous data from the buffer, I fear to end up leaking memory.

src/network/connection.js
bwdeng: good catch! copy-paste bug...

bwdeng:
@tulios It's more likely to happen if multiple requests were sent currently. We were facing a problem when starting a new consumer group. Sometimes, it causes the response of `OffsetCommit` got lost during offset resetting, causing consumer stuck on initialization.

src/network/connection.js
bwdeng: I think we are okey, since the buffer is reset using `decoder.readAll()` which only copies (not slicing) the rest of data. So the original buffer could be collected by GC once the response gets consumed.

tulios:
@bwdeng the code makes sense, I just wanted to confirm. I want to name some things (like `data.slice(0, Decoder.int32Size() + expectedResponseSize)`), but I can do it later.

Thanks again!

tulios:
@bwdeng Thanks again, I'm releasing `v1.3.1` with this fix.

o/

Issue #57, opened 2018-05-15
Add support to fetch protocol versions v4 to v6


tulios:
```txt
Fetch Request (Version: 4) => replica_id max_wait_time min_bytes max_bytes isolation_level [topics] 
  replica_id => INT32
  max_wait_time => INT32
  min_bytes => INT32
  max_bytes => INT32
  isolation_level => INT8
  topics => topic [partitions] 
    topic => STRING
    partitions => partition fetch_offset max_bytes 
      partition => INT32
      fetch_offset => INT64
      max_bytes => INT32
```

```txt
Fetch Response (Version: 4) => throttle_time_ms [responses] 
  throttle_time_ms => INT32
  responses => topic [partition_responses] 
    topic => STRING
    partition_responses => partition_header record_set 
      partition_header => partition error_code high_watermark last_stable_offset [aborted_transactions] 
        partition => INT32
        error_code => INT16
        high_watermark => INT64
        last_stable_offset => INT64
        aborted_transactions => producer_id first_offset 
          producer_id => INT64
          first_offset => INT64
      record_set => RECORDS
```

More information at:
https://kafka.apache.org/protocol.html#The_Messages_Fetch

Investigate API v7

PR #116, opened 2018-08-18
Fix fetch v4 for partial messages on record batch

### LINKED ISSUE #113: Messages with headers fail to be decoded in Fetch V4
Decoding the messages fails if messages produced include headers.

tulios:
Fix #113 

Related to  #57

### DIFF
--- a/src/consumer/__tests__/consumeMessages.spec.js
+++ b/src/consumer/__tests__/consumeMessages.spec.js
@@ -10,6 +10,7 @@ const {
   newLogger,
   waitFor,
   waitForMessages,
+  testIfKafka011,
 } = require('testHelpers')
 
 describe('Consumer', () => {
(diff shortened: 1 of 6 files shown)


src/consumer/__tests__/consumeMessages.spec.js
```diff
@@ -177,6 +178,139 @@ describe('Consumer', () => {
...
+      acks: 1,
+      topicMessages: [
+        { topic: topicName, messages: messages1 },
+        { topic: topicName2, messages: messages2 },
+      ],
+    })
+
+    await waitForMessages(messagesConsumed, { number: messages1.length + messages2.length })
+
+    expect(cluster.refreshMetadataIfNecessary).toHaveBeenCalled()
```
Nevon: Why is this check being done?

src/protocol/requests/fetch/v4/response.js
```diff
@@ -26,7 +26,13 @@ const MAGIC_OFFSET = 16
  */
 
 const decodeMessages = async decoder => {
-  const messagesBuffer = decoder.readBytes()
+  const messagesSize = decoder.readInt32()
+
+  if (messagesSize <= 0 || !decoder.canReadBytes(messagesSize)) {
```
Nevon: I get what this is doing, but it feels really strange that we'd get `messageSize` > 0 and yet not be able to read enough bytes, and to then return an empty set of messages. Feels wrong. In that situation, shouldn't we buffer until we have enough bytes to read the expected `messageSize`?

src/consumer/__tests__/consumeMessages.spec.js
tulios: The other tests also have the check, it's just to make sure that the function is called

src/protocol/requests/fetch/v4/response.js
Nevon: Discussed outside of Github. Basically the response is complete, so maybe the broker is trying to tell us something by returning a partial record batch, but either way, by returning an empty set of messages here, we should get the complete record batch on the next fetch.

Related: https://github.com/dpkp/kafka-python/blob/ed85c58726f867d956f705194765b18ed3ac3855/kafka/protocol/message.py#L173

PR #135, opened 2018-09-23
Handle API version not supported

tulios:
This PR does two things:

1) All single response protocols now validate error code 35 (`UNSUPPORTED_VERSION`) and throw the appropriate error

```javascript
{
  "level": "ERROR",
  "timestamp": "2018-09-23T10:39:18.904Z",
  "logger": "kafkajs",
  "message": "[Connection] Response ApiVersions(key: 18, version: 2)",
  "broker": "192.168.43.30:9093",
  "clientId": "test-9e51c3d1c7f66590661b",
  "error": "The version of API is not supported",
  "correlationId": 0,
  "size": 10
}
```

\* this particular error is suppressed

2) `Broker#apiVersions` will try to use the latest supported version of the APIVersions protocol. Currently, KafkaJS is always using version 0. This change will allow us to use newer versions of some protocols.

The error log for `APIVersion` was suppressed to avoid unnecessary error logs for an operation that will always happen.

### DIFF
--- a/src/protocol/error.js
+++ b/src/protocol/error.js
@@ -473,7 +473,7 @@ const createErrorFromCode = code => {
 }
 const failIfVersionNotSupported = code => {
   if (code === UNSUPPORTED_VERSION_CODE) {
-    throw new KafkaJSProtocolError(errorCodes.find(e => e.code === code))
+    throw createErrorFromCode(UNSUPPORTED_VERSION_CODE)
   }
 }
 
--- a/src/protocol/error.js
+++ b/src/protocol/error.js
@@ -465,13 +465,21 @@ const unknownErrorCode = errorCode => ({
 })
 
 const SUCCESS_CODE = 0
+const UNSUPPORTED_VERSION_CODE = 35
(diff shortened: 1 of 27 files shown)


src/protocol/error.js
```diff
@@ -465,13 +465,21 @@ const unknownErrorCode = errorCode => ({
...
 const SUCCESS_CODE = 0
+const UNSUPPORTED_VERSION_CODE = 35
+
 const failure = code => code !== SUCCESS_CODE
 const createErrorFromCode = code => {
   return new KafkaJSProtocolError(errorCodes.find(e => e.code === code) || unknownErrorCode(code))
 }
+const failIfVersionNotSupported = code => {
+  if (code === UNSUPPORTED_VERSION_CODE) {
+    throw new KafkaJSProtocolError(errorCodes.find(e => e.code === code))
```
Nevon: Just delegate this to `createErrorFromCode` instead, so you don't need to replicate the logic for finding the error.

PR #136, opened 2018-09-23
Make consumers detect that a topic was updated

### LINKED ISSUE #133: Trigger rebalance if partitions change after metadata update
With the autorefresh metadata functionality, if the partitions have changed, we should also trigger a rebalance. Currently, if you add a new partition to a topic, the consumer group won't rebalance.

tulios:
Consumers can now detect that their assignment is old and force the group to rebalance. The consumers will now verify that they are aware of the assigned partitions after the rebalance and perform a refresh if necessary.

Fixes #133

### DIFF
--- a/src/consumer/consumerGroup.js
+++ b/src/consumer/consumerGroup.js
@@ -10,7 +10,7 @@ const { MemberAssignment } = require('./assignerProtocol')
 const {
   KafkaJSError,
   KafkaJSNonRetriableError,
-  KafkaJSStaleTopicMetadataAssigment,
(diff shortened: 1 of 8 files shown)


src/consumer/consumerGroup.js
```diff
@@ -110,6 +115,7 @@ module.exports = class ConsumerGroup {
       }
 
       await this.cluster.refreshMetadataIfNecessary()
```
Nevon: Maybe change this to always refresh metadata if you are the leader, so we don't need to do another round of re-syncs if the leader is out of date.

PR #151, opened 2018-10-08
Filter buffer data from connection failure debug logs

paambaati:
This PR replaces `Buffer` payloads from connection debug logs. They're quite huge, and almost always useless.

If the payload is a `Buffer`, it is replaced with the string `[filtered]`.


Nevon:
I'm not so sure about this. When I debug log, it's because I want to see _everything_. I seem to recall us quite often using this output when investigating protocol decoding issues.

Is it a problem for you that the output is large when running in debug mode? Maybe things like this could be toggled with an environment flag if this really is an issue (something like `DEBUG_LOG_OMIT_BUFFERS` maybe).

paambaati:
@Nevon I wrote this because I had a lot of `OFFSET_OUT_OF_RANGE` errors, and the payload was humongous (on a topic with 400 partitions), actually slowing down my terminal app.

> When I debug log, it's because I want to see _everything_.

Forgive my ignorance, but does a raw buffer in the output give you useful information?

Nevon:
As an application developer, no, it's not particularly useful, but as a KafkaJS developer it's useful, because if my consumer is crashing when I receive a response, I can copy that buffer into a fixture and write a test to figure out why my decoding is failing.

But like I said, I'm open to making this configurable with a runtime flag, because as you said, it's very unlikely to be useful to most people.

paambaati:
@Nevon Fair enough! I'll close this PR, thanks for reviewing this.

PR #154, opened 2018-10-08
Replace event emitter Lock

### LINKED ISSUE #153: MaxListenersExceededWarning on high concurrency/big cluster
The broker uses a lock mechanism to prevent issues with authentication when creating connections in parallel; this can be caused by the use of `producer.send` in a `Promise.all` or in recovery scenarios where the brokers have to re-connect.

The lock mechanism is based on `EventEmitter`, which has max listeners set to 11. If the number of connections is higher than 11, it will warn with something like:

```sh
(node:20592) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 releaseLock listeners added. Use emitter.setMaxListeners() to increase limit
```

To solve this, we must set `maxListeners` to the number of brokers in the cluster; this can be done after metadata refresh in the `BrokerPool`.

https://github.com/tulios/kafkajs/blob/cf0005f07a1465cca0697463f1b7d7f8c6888e07/src/cluster/brokerPool.js#L111

tulios:
Fixes #153
This PR updates the lock class to use a pure implementation, completely removing the event emitter.

I also had to update all test certificates because they expired today.


tulios:
We reviewed this PR F2F yesterday. @Nevon go ahead and add more comments if you need

PR #174, opened 2018-10-11
Expose a "getTopicMetadata" method on admin instances

ianwsperber:
Hiya! 

## Description

We had a need to lookup the partition metadata for a topic but found this method was not exposed in the public API, though the functionality exists on cluster instances. Accordingly I've added a method `getTopicMetadata` on the admin instance to permit users to lookup topic metadata. Since [admin instances are created with the `allowAutoTopicCreation ` option set to `false`](https://github.com/tulios/kafkajs/blob/master/src/index.js#L104) we should expect this method to throw an error when run against a topic that does not exist. Please let me know if there are changes we should make to bring this up to snuff, or if you have concerns about whether this method should be exposed.

## Example

```javascript
const admin = kafka.admin();
await admin.connect();
const { partitions } = await admin.getTopicMetadata('my-topic');
const numPartitions = partitions.length;
```

## Addendum 

As something of an aside, this work came up because we have a need to assign our messages to a partition according to a property other than _key_, because we are [compacting](http://cloudurable.com/blog/kafka-architecture-log-compaction/index.html) on the key. Currently we plan to accomplish this by adding a custom partitioner, identical to the [default](https://github.com/tulios/kafkajs/blob/master/src/producer/partitioners/default/index.js), except with the following as the first condition:

```javascript
if (typeof message.partition === 'string') {
      return toPositive(murmur2(message.partition)) % numPartitions
}
```

Basically if the `partition` property is a string we hash on that property instead of the key. Do you think this is a common enough use case that it would be worth porting into the library? Or is it best left outside the library as a custom partitioner for the client to implement?

Thanks!

### DIFF
--- a/src/admin/index.js
+++ b/src/admin/index.js
@@ -419,10 +419,17 @@ module.exports = ({ retry = { retries: 5 }, logger: rootLogger, cluster }) => {
   }
 
   /**
-   * Retrieve metadata for a topic.
+   * Fetch metadata for provided topics.
    *
-   * @param {string} topic
-   * @return {Promise<TopicMetadata>}
+   * If no topics are provided fetch metadata for all topics of which we are aware.
+   * @see https://kafka.apache.org/protocol#The_Messages_Metadata
+   *
+   * @param {Object} [options]
+   * @param {Array<string>} [options.topics]
+   * @return {Promise<TopicsMetadata>}
+   *
+   * @typedef {Object} TopicsMetadata
+   * @property {Array<TopicMetadata>} topics
    *
    * @typedef {Object} TopicMetadata
    * @property {String} name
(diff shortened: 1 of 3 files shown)


src/admin/index.js
```diff
@@ -418,6 +418,39 @@ module.exports = ({ retry = { retries: 5 }, logger: rootLogger, cluster }) => {
...
+   * @typedef {Object} PartitionMetadata
+   * @property {number} partitionErrorCode Response error code
+   * @property {number} partitionId Topic partition id
+   * @property {number} leader  The id of the broker acting as leader for this partition.
+   * @property {Array<number>} replicas The set of all nodes that host this partition.
+   * @property {Array<number>} isr The set of nodes that are in sync with the leader for this partition.
+   *
+   * @see https://kafka.apache.org/protocol#The_Messages_Metadata
+   */
+  const getTopicMetadata = async topic => {
```
tulios: Can you change this to accept any number of topics?

```javascript
admin.getTopicMetadata({
  topics: ['topic-A', 'topic-B']
})
```

The use of an object seems strange to begin with, but it allows us to extend the API later, for example, adding a timeout option, etc.

The API should also allow you to fetch metadata for all topics, refresh metadata will do that if no topics are included with the target topic API.

```javascript
admin.getTopicMetadata()
```

tulios:
I think you are making a good use of the partitioner, but I don't think this logic belongs to the library unless we find out that this is some sort of standard.

Thanks again for the PR, we are adding methods to the admin as we need them, so feel free to create new issues or go straight to PRs.

src/admin/index.js
ianwsperber: Seems like a sensible design choice to me @tulios. I had the method return an object with property `partitions` rather than the partitions array itself for similar reasons. I'll push that change today.

ianwsperber:
🎉 Thanks @tulios & @Nevon !

PR #176, opened 2018-10-14
Don't log buffer on network send errors by default

### LINKED ISSUE #172: Add runtime flag for displaying buffers in debug output
#151 brought up that it can be problematic for people to log the entire buffers when debug logging, especially when consuming from topics with many partitions. Logging the buffers is very useful to us when developing KafkaJS, but not very useful to our end users.

I would propose that we don't log buffers by default, and only log them when the environment flag `DEBUG_LOG_BUFFERS` is set. We can add that by default to our yarn scripts while developing.

When that flag is not present, instead of logging `{"type":"Buffer","data":[1,2,3,4,5]}` we should log `{"type":"Buffer","data":"[filtered]"}`

Nevon:
Exposes a new environment flag `DEBUG_LOG_BUFFERS`, which when set to `1` will log the full payload.
If not set, the payload will be logged as:

```
{ "type": "Buffer", "data": "[filtered]"}
```

Fixes #172 

@paambaati

### DIFF
--- a/src/network/connection.js
+++ b/src/network/connection.js
@@ -59,6 +59,7 @@ module.exports = class Connection {
 
     this.logDebug = log('debug')
     this.logError = log('error')
+    this.shouldLogBuffers = process.env['DEBUG_LOG_BUFFERS'] === '1'
   }
 
   /**
@@ -265,10 +266,12 @@ module.exports = class Connection {
         })
       }
 
+      const isBuffer = Buffer.isBuffer(payload)
       this.logDebug(`Response ${requestInfo(entry)}`, {
         error: e.message,
         correlationId,
-        payload,
+        payload:
+          isBuffer && !this.shouldLogBuffers ? { type: 'Buffer', data: '[filtered]' } : payload,
       })
 
       throw e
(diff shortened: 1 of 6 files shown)


tulios:
I like the change but I think it will be hard to remember where the special env vars are, so any thoughts about where to keep this variable? I understand that the connection is where it's used, but I think having "random" envs lurking through the code will make it hard to maintain.

src/network/connection.js
```diff
@@ -59,6 +59,7 @@ module.exports = class Connection {
 
     this.logDebug = log('debug')
     this.logError = log('error')
+    this.shouldLogBuffers = process.env['DEBUG_LOG_BUFFERS'] === '1'
```
tulios: Let's also scope the variable, maybe `KAFKAJS_DEBUG_PROTOCOL_BUFFERS`

Nevon:
I had the same thought about where to put it. I considered resolving it somewhere high up and then pass it all the way down to the connection, but then testing that it gets passed all the way through the connection builder etc is a hassle. Another option I considered was to actually put it in the logger, but then I'd have to check recursively for buffers, which I most certainly don't want to do for performance reasons.

Let's chat and figure out where to put it.

Issue #165, opened 2018-10-10
Add protocol findCoordinator v1


tulios:
Since the transaction coordinator is at the center assigning PIDs and managing transactions, the first thing a producer has to do is issue a FindCoordinatorRequest to any broker to discover the location of its coordinator.

`coordinator_type`: The type of coordinator to find (0 = group, 1 = transaction)

```
FindCoordinator Request (Version: 1) => coordinator_key coordinator_type 
  coordinator_key => STRING
  coordinator_type => INT8
```

```
FindCoordinator Response (Version: 1) => throttle_time_ms error_code error_message coordinator 
  throttle_time_ms => INT32
  error_code => INT16
  error_message => NULLABLE_STRING
  coordinator => node_id host port 
    node_id => INT32
    host => STRING
    port => INT32
```

https://kafka.apache.org/protocol.html#The_Messages_FindCoordinator

ianwsperber:
Hey @tulios! FYI I'm planning to begin implementing this protocol today. LMK if you'd like this work PR'd into a specific branch or have any other guidance to offer. Thanks 😄

tulios:
You can create one branch per issue; we can merge things faster this way. Just ping me if you need anything.

tulios:
Implemented in #189

PR #189, opened 2018-10-31
EoS: Implement FindCoordinator v1

ianwsperber:
See #165

This PR implements the `FindCoordinator` V1 protocol as described in the issue. This protocol will now be consumed by the broker in its `findGroupCoordinator` call. I suppose this could have gone behind some sort of feature flag (like `allowExperimentalV011`) but AFAIK it should play find with the existing code.

For backwards compatibility I chose to maintain the `groupId` option and map it to the new `coordinatorKey` option.

I've had some trouble getting the full test suite to run consistently on my local machine, but the relevant tests appear to be passing 😬

This is my first PR touching the protocols & the first starting work on the EoS stuff, so please let me know if there's anything that should be done differently 😄 Thanks!


tulios:
@ianwsperber awesome work. I've added a couple of commits to accept the new argument on the broker.

About the tests, they have been working really well for me in the local environment. What I usually do is to run the Docker container on a separate terminal and then run `./scripts/createScramCredentials.sh` to create the credentials. Then you just have to use `yarn test:local` to run all tests or `yarn test:local:watch src/<>` for some specific part you are developing. Maybe we should document this workflow on the readme.

>  I chose to maintain the groupId option and map it to the new coordinatorKey option.

This is a sensible decision

ianwsperber:
Thanks for the review @tulios! If you don't get to it first I could add those testing steps in the next day or so. Maybe it'd be helpful to create a `CONTRIBUTING.md` to collect this an related info?

PR #190, opened 2018-10-31
EoS #166: Add protocol InitProducerId V0

### LINKED ISSUE #166: Add protocol initProducerId v0
After discovering the location of its coordinator, the next step is to retrieve the producer’s ID. This is achieved by issuing an `initProducerId` to the transaction coordinator.

In addition to returning the ID, the initProducerId performs the following tasks:

- Bumps up the epoch of the ID, so that any previous zombie instance of the producer is fenced off and cannot move forward with its transaction.
- Recovers (rolls forward or rolls back) any transaction left incomplete by the previous instance of the producer.
- The handling of the`initProducerId` is synchronous. Once it returns, the producer can send data and start new transactions.

When an TransactionalId is not specified:
If no TransactionalId is specified in the configuration, a fresh ID is assigned, and the producer only enjoys idempotent semantics and transactional semantics within a single session.

```
InitProducerId Request (Version: 0) => transactional_id transaction_timeout_ms 
  transactional_id => NULLABLE_STRING
  transaction_timeout_ms => INT32
```

```
InitProducerId Response (Version: 0) => throttle_time_ms error_code producer_id producer_epoch 
  throttle_time_ms => INT32
  error_code => INT16
  producer_id => INT64
  producer_epoch => INT16
```

https://kafka.apache.org/protocol.html#The_Messages_InitProducerId

ianwsperber:
Resolves #166 

Implements `InitProducerId` protocol V0 and exposes on the broker.


src/protocol/requests/initProducerId/v0/response.js
```diff
@@ -0,0 +1,37 @@
...
+  const decoder = new Decoder(rawData)
+  const throttleTime = decoder.readInt32()
+  const errorCode = decoder.readInt16()
+
+  failIfVersionNotSupported(errorCode)
+
+  return {
+    throttleTime,
+    errorCode,
+    producerId: decoder.readInt64(),
```
tulios:
```suggestion
    producerId: decoder.readInt64().toString(),
```

It's better to keep the int64 as a string to avoid any coupling with `Long`, we plan to replace `Long` with the new BigInt. It's mainly used for operations.

src/protocol/requests/initProducerId/index.js
```diff
@@ -0,0 +1,12 @@
+const versions = {
+  0: ({ transactionalId, transactionTimeout }) => {
+    const request = require('./v0/request')
+    const response = require('./v0/response')
+    return { request: request({ transactionalId, transactionTimeout }), response }
```
tulios: What do you think about a default `transactionTimeout` here? Maybe `5000`.

src/protocol/requests/initProducerId/v0/response.js
ianwsperber: Done!

src/protocol/requests/initProducerId/index.js
ianwsperber: I'm in favor of providing a default. I suppose it could go here _or_ in the broker... but I'll add here for now

PR #195, opened 2018-11-06
EoS TxnOffsetCommit v0

### LINKED ISSUE #169: Add protocol TxnOffsetCommit v0
The producer will send a TxnOffsetCommit to the consumer coordinator to persist the offsets in the `__consumer-offsets` topic. The consumer coordinator validates that the producer is allowed to make this request (and is not a zombie) by using the ID and producer epoch which are sent as part of this request.

The consumed offsets are not visible externally until the transaction is committed.

```
TxnOffsetCommit Request (Version: 0) => transactional_id group_id producer_id producer_epoch [topics] 
  transactional_id => STRING
  group_id => STRING
  producer_id => INT64
  producer_epoch => INT16
  topics => topic [partitions] 
    topic => STRING
    partitions => partition offset metadata 
      partition => INT32
      offset => INT64
      metadata => NULLABLE_STRING
```

```
TxnOffsetCommit Response (Version: 0) => throttle_time_ms [topics] 
  throttle_time_ms => INT32
  topics => topic [partitions] 
    topic => STRING
    partitions => partition error_code 
      partition => INT32
      error_code => INT16
```

https://kafka.apache.org/protocol.html#The_Messages_TxnOffsetCommit

ianwsperber:
Resolves #169.

In testing I was not able to get a rejection for an invalid transaction id/producer id/producer epoch. Accordingly I wrote my broker test to just use mock values for these fields to make that behavior explicit. @tulios do you have any thoughts on this? [According to the design doc](https://docs.google.com/document/d/11Jqy_GjUGtdXJK94XGsEIK7CP1SnQGdp2eF0wSw9ra8/edit#bookmark=id.5xa2hzric4h0) my request should be rejected, but I'm unsure if that's Kafka's actual behavior.

Also added a little bit of documentation to the transaction methods in the broker.


src/broker/__tests__/txnOffsetCommit.spec.js
```diff
@@ -0,0 +1,125 @@
...
+    await transactionBroker.connect()
+    const result = await transactionBroker.initProducerId({
+      transactionalId,
+      transactionTimeout: 30000,
+    })
+
+    producerId = result.producerId
+    producerEpoch = result.producerEpoch
+
+    await transactionBroker.addPartitionsToTxn({
```
ianwsperber: @tulios I've added the preceding steps in the protocol here. You can see below we still don't get a rejection for invalid values.

tulios: Yeah, well that's unfortunate, but I guess we will have more data once we start the actual producer. If we discover something new, we can come back add more tests.

ianwsperber: Sounds good to me @tulios. If we ultimately conclude that this is indeed Kafka's behavior today it may be prudent to raise an issue with the Kafka team to identify whether it's the intended behavior or a bug.

tulios:
@ianwsperber I was trying to test the commit flow following the steps in here

https://cwiki.apache.org/confluence/display/KAFKA/KIP-98+-+Exactly+Once+Delivery+and+Transactional+Messaging#KIP-98-ExactlyOnceDeliveryandTransactionalMessaging-DataFlow

and I just realized that we are not passing the transaction attributes to the produce request, I started some of this work here:

https://github.com/tulios/kafkajs/tree/add-description-to-lock-instances

I have a theory that we do not see the correct behavior because we haven't produced any messages, so there's nothing associated with the `producerId`. I haven't managed to finish the test since I can't produce so you might want to take a look.

I'm busy with some other things this week and the next, but I should be able to join this initiative soon.

ianwsperber:
Ok, I'll take a look at your branch @tulios. If I get anywhere I'll create a separate branch and PR. It's curious because I _was_ able to validate all those failure test cases for `EndTxn` in #198, so if producing messages fixes this issue then the behavior is at least inconsistent.

Issue #196, opened 2018-11-07
Propagate transactional attributes to produce V3 and RecordBatch


tulios:
Currently `producerId`, `producerEpoch` and the transactional attribute of the RecordBatch are not propagated through `broker.produce`.

https://github.com/tulios/kafkajs/tree/add-transactional-attrs-to-record-batch

The branch above has some of the work required, but it's not complete

tulios:
Fixed on #199

PR #199, opened 2018-11-07
Add transactional attributes to record batch

tulios:
This PR is ready now; it's missing a test for the producer with the transactional flag on but we will have to wait for EOS support to test it properly.


src/protocol/recordBatch/v0/index.js
```diff
@@ -4,6 +4,9 @@ const crc32C = require('../crc32C')
 const { Types: Compression, lookupCodec } = require('../../message/compression')
 
 const MAGIC_BYTE = 2
+const COMPRESSION_MASK = 11 // The lowest 3 bits
```
ianwsperber: Why not continue to use `3`, as we have in the [compression module](https://github.com/tulios/kafkajs/blob/master/src/protocol/message/compression/index.js#L3)? As far as I can tell they're effectively the same, except this masks a 4th bit that won't be present.

tulios: You are right, I tried to move away from the hex definition but messed up the mask. It should be `3`. We will have to bump this in the future to support ZSTD, but let's do it when the time comes.

tulios: ef252f7

ianwsperber:
Awesome, thanks @tulios, LGTM!

ianwsperber:
This resolves #196

Issue #173, opened 2018-10-11
Transactional producer


tulios:
The transactional producer allows an application to send messages to multiple partitions (and topics) atomically. Each transactional producer has a unique `transactionalId`, which is used to enable transaction recovery across multiple sessions.

There can be only one open transaction per instance of the producer.

__Proposed API:__

```javascript
const producer = kafka.producer({ transactional: true })

const run = async () => {
  await producer.connect()
  const tx = await producer.tx()

  try {
    for (let i = 0; i < 100; i++) {
      await tx.add({
        topic,
        messages: [{ key: `key-${i}`, value: `value-${i}` }]
      })
    }
 
    await tx.commit()
  } catch (e) {
    // Investigate class of errors where the producer can't recover, i.e: ProducerFencedException
    await tx.abort()
  }

  await producer.disconnect()
}
```

__Workflow:__

`await producer.connect()`:

* Should call `initProducerId` to initialize the producer, since it can only happen once per instance it could be combined with `connect`
* Add a new log line to indicate the use of a transactional producer

`const tx = await producer.tx()`:

I still have mixed feeling about this design, but the idea is to create a transaction object where the user will be able to add records, commit, and abort.
* It should check for other producers with the same ID
* In memory flag to indicate that a transaction was created. There can be only one open transaction per instance of the producer.

`await tx.add({ ... })`:

Adds the records to some record accumulator

`await tx.commit()`:

* Calls `AddPartitionsToTxn`
* Calls `EndTxnRequest` with COMMIT

`await tx.abort():`

* Clear the in-memory accumulator
* Calls `EndTxnRequest` with ABORT

We have to investigate the errors that cause the consumer to disconnect instead of abort. From the Java client, we have: `ProducerFencedException`, `OutOfOrderSequenceException`, and `AuthorizationException` where the authorization error is specific to the current `transactionalId`

`await producer.disconnect()`:

Should close the connection as usual

__Open questions:__
Do we need guidelines for the `transactionalId`?

Should we auto-generate the `transactionId` if it's not defined?

__NOTE__:

The API design is still a draft

Depends on #165 #166 #167 #170

@Nevon @ianwsperber ^

Nevon:
Some thoughts from our discussion.

`tx.add` could be `tx.send` if we send the messages to the server, rather than buffering them up on the client. This would align the transaction API with the producer API nicely. `abort` would then become `rollback`.

An alternative way to create transactions could be something like:

```javascript
[...]

PR #203, opened 2018-11-13
EoS #200 idempotent producer

### LINKED ISSUE #200: Idempotent Producer
Allow users to create idempotent producers. We create an idempotent producer by passing an `idempotent` flag during creation:

```
const idempotentProducer = kafka.producer({ idempotent: true });
```

Creating an idempotent producer is a prerequisite for the transactional producer (see related discussion in #173 )

**Tasks**

- [x] Generate PID using protocol `InitProducerId v0` during initialization (perhaps when we call `producer.connect()`)
- [x] Track sequence numbers per TopicPartition
- [x] Send the PID, epoch and sequence with each record batch (see work already done in #199)
- [x] Ensure we enforce acks, retries & max requests (or just log a warning...). Partially blocked by #148 

**Spec**

From KIP-98:
> When idempotence is enabled, we enforce that acks=all, retries > 1, and max.inflight.requests.per.connection=1. Without these values for these configurations, we cannot guarantee idempotence. If these settings are not explicitly overidden by the application, the producer will set acks=all, retries=Integer.MAX_VALUE, and max.inflight.requests.per.connection=1 when idempotence is enabled.

ianwsperber:
Resolves #200 

Add an `idempotent` flag to the producer factory method to support creation of idempotent producers. Documented this as an "experimental" feature, since I haven't had a chance to observe its behavior in practice yet.

Implementation details
- Idempotent producer will init the PID & epoch during the `producer.connect()` call
- Idempotent producer uses the `transactionManager` to track sequences across messages and include in `produce` request
- Idempotent producer enforces config options`acks=-1` and `retries>0`. Default retry of MAX_SAFE_INTEGER

This PR additionally includes a few tests verifying the recent work allowing the `firstSequence` option to be provided in record batches.


src/producer/index.js
```diff
@@ -16,13 +17,28 @@ module.exports = ({
...
+    logger.warn('Limiting retries for the idempotent producer may invalidate EoS guarantees')
+  }
+
   const partitioner = createPartitioner()
   const retrier = createRetry(Object.assign({}, cluster.retry, retry))
   const instrumentationEmitter = new InstrumentationEventEmitter()
-  const logger = rootLogger.namespace('Producer')
-  const sendMessages = createSendMessages({ logger, cluster, partitioner })
+  const transactionManager = createTransactionManager({ logger, cluster })
+  const sendMessages = createSendMessages({ logger, cluster, partitioner, transactionManager })
```
ianwsperber: Currently I pass `transactionManager` to the factory method. I do wonder if ultimately it will make more sense to provide a transaction manager in the `sendMessages` call. Currently it doesn't really matter but it may be something to refactor in #173

Nevon: In what way would it make the transaction manager less stateful? Maybe I'm missing something, but passing it into the factory method seems much more appropriate than passing it with each sendMessage call.

src/producer/sendMessages.js
```diff
@@ -67,18 +76,36 @@ module.exports = ({ logger, cluster, partitioner }) => {
...
+
           const expectResponse = acks !== 0
-          responsePerBroker.set(broker, expectResponse ? responseSerializer(response) : [])
+          const formattedResponse = expectResponse ? responseSerializer(response) : []
+
+          formattedResponse.forEach(({ topicName, partition }) => {
+            const size = topicMetadata.get(topicName).messagesPerPartition[partition].length
+            const previous = topicMetadata.get(topicName).sequencePerPartition[partition]
+
+            transactionManager.updateSequence(topicName, partition, previous + size)
```
tulios: The sequence number is an `int32`, it will eventually reach the max number. I think we need to rotate the number, similar to what I do with the correlation id.

https://github.com/tulios/kafkajs/blob/095b6e0c5b7623634eb37eca176433808f9d6154/src/network/connection.js#L316-L322

WDYT?

src/producer/transactionManager.js
```diff
@@ -0,0 +1,97 @@
...
+     * Overwrites any existing state in this transaction manager
+     */
+    initProducerId: async () => {
+      await cluster.refreshMetadataIfNecessary()
+
+      // If non-transactional we can request the PID from any broker
+      const broker = await cluster.findControllerBroker()
+      const result = await broker.initProducerId({ transactionTimeout })
+
+      producerId = result.producerId
```
tulios: Shouldn't we preserve this across reconnections?

src/producer/index.js
ianwsperber: You're probably right @Nevon, it's just that once we introduce transactions the transaction manager will become a bit of a state machine and I was trying to think of strategies to minimize the state we have to manage. One way of minimizing that would to be to have the transaction manager fetch the producer id on instantiation, so that there's no ambiguity about whether the idempotent producer has already be initialized. Since that fetch occur after connection that would probably mean passing the transaction manager into the send messages method, or lazily creating the sendMessages instance. Ultimately I do think what I've implemented is the most straight forward solution, with the benefit that it closely mirrors how transactions are managed in other Kafka libraries. Thus far in practice I haven't had an issue working with the state machine ([see preview here](https://github.com/tulios/kafkajs/blob/eos-173-transactional-producer/src/producer/transactionManager.js), just thinking through the alternatives.

src/producer/transactionManager.js
ianwsperber: @tulios Yes I think we should, good point! I'll update the producer so that it only calls `initProducerId` if uninitialized.

src/producer/sendMessages.js
ianwsperber: Sounds good to me! Worth noting that the Java implementation appears not to bother with this check (will implement nonetheless): https://github.com/apache/kafka/blob/9a0ea25fee85837748145d37c69cf4d9bb7f9933/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java#L417

ianwsperber: @tulios Also, `Number.MAX_VALUE` is significantly larger than the maximum (signed) Int32 value (2 ^ 32 vs 2 ^ 1024). AFAIK JS has no predefined constant for Int32, so I'll need to define my own.

ianwsperber:
@tulios @Nevon Thanks for the feedback! Made the requested changes for updating sequence

src/producer/transactionManager.js
tulios: @ianwsperber will you do this on the transactional producer PR or this one? (If you decide for the transactional PR we can merge)

ianwsperber: @tulios Forgot about this, just pushed a commit to address

PR #209, opened 2018-11-20
Eos 201 list offsets v2

### LINKED ISSUE #201: Add protocol ListOffsets v2
Consumers must be able to send `ListOffsets` version >= 2 to provide their `isolation_level`

> isolation_level | This setting controls the visibility of transactional records. Using READ_UNCOMMITTED (isolation_level = 0) makes all records visible. With READ_COMMITTED (isolation_level = 1), non-transactional and COMMITTED transactional records are visible. To be more concrete, READ_COMMITTED returns all data from offsets smaller than the current LSO (last stable offset), and enables the inclusion of the list of aborted transactions in the result, which allows consumers to discard ABORTED transactional records

```
ListOffsets Request (Version: 2) => replica_id isolation_level [topics] 
  replica_id => INT32
  isolation_level => INT8
  topics => topic [partitions] 
    topic => STRING
    partitions => partition timestamp 
      partition => INT32
      timestamp => INT64
```

```
ListOffsets Response (Version: 2) => throttle_time_ms [responses] 
  throttle_time_ms => INT32
  responses => topic [partition_responses] 
    topic => STRING
    partition_responses => partition error_code timestamp offset 
      partition => INT32
      error_code => INT16
      timestamp => INT64
      offset => INT64
```

https://kafka.apache.org/protocol#The_Messages_ListOffsets

ianwsperber:
Resolves #201 

Implement protocol `ListOffsets` `v2`, so that we can now provide an isolation level. For now continue with the default of "read_comitted".

The v2 response format is divergent from v0, as it returns a single offset rather than an array. AFAIK there's no pattern for normalizing responses (I'm unsure if that's even desirable), so I decided to handle it in an ad-hoc fashion within the cluster module.

Made the base branch `eos` so that this can land with the transactional producer & consumer.


src/broker/index.js
```diff
@@ -341,20 +341,23 @@ module.exports = class Broker {
...
+   * @property {number} [timestamp=-1]
+   *
+   *
    * @returns {Promise}
    */
-  async listOffsets({ replicaId, topics }) {
+  async listOffsets({ replicaId, isolationLevel, topics }) {
     const listOffsets = this.lookupRequest(apiKeys.ListOffsets, requests.ListOffsets)
-    return await this.connection.send(listOffsets({ replicaId, topics }))
+    return await this.connection.send(listOffsets({ replicaId, isolationLevel, topics }))
```
Nevon: I would suggest doing the normalization between versions here, rather than in the cluster. We don't really have an established pattern for where to normalize. We have done it in the protocol in some place (unfortunately I can't remember where...), but I think we should try to keep our protocol true to the Kafka protocol. But we shouldn't let that bleed through too far, like the cluster, so I think the broker is the best place for it that we have at the moment.

Nevon:
Looks good. 👏 Just don't forget that we need v1 as well.

src/broker/index.js
ianwsperber: Ok. I _do_ like currently that the broker is a thin wrapper to execute the protocols. I also agree that we should not handle the normalization in the protocols (it's done for fetch with record sets vs messages, and it's giving me some trouble now that we're working with transactions). I think ideally there'd be a wrapper around the broker that did normalization. But I'm fine moving into the broker for now, just didn't want to introduce a new pattern without consensus 😄

ianwsperber:
@Nevon Are you just generally noting that we'd like to add ListOffsets v1? Or are you saying that it's necessary?

tulios:
KafkaJS generally assumes that all the versions in between the ranges are implemented so that we will need v1. I just checked, and it's similar to v2, but we can do it in another PR if you want.

PR #211, opened 2018-11-21
Document fromBeginning option

paambaati:
This PR documents the `fromBeginning` option for consumers. See https://github.com/tulios/kafkajs/issues/204 for the related discussion.


Nevon:
Thanks for the contribution!

PR #216, opened 2018-11-26
Accept max in-flight requests on the connection

tulios:
This PR adds support to max in-flight requests. The option is set to `null` now so we can test it out, I want to understand the impact of more or less concurrent requests before we commit to a default value.


src/network/requestQueue/socketRequest.js
```diff
@@ -0,0 +1,116 @@
...
+        })
+      )
+    }
+
+    this.timeoutId = setTimeout(timeoutCallback, this.requestTimeout)
+  }
+
+  completed({ size, payload }) {
+    if (this[PRIVATE.STATE] !== REQUEST_STATE.SENT) {
+      return
```
ianwsperber: Should these non-matching states be error conditions?

src/network/requestQueue/socketRequest.js
```diff
@@ -0,0 +1,116 @@
...
+
+const REQUEST_STATE = {
+  PENDING: Symbol('PENDING'),
+  SENT: Symbol('SENT'),
+  COMPLETED: Symbol('COMPLETED'),
+  REJECTED: Symbol('REJECTED'),
+}
+
+/**
+ * @typedef {Object} SocketRequest
```
ianwsperber: Might be nice to add a line explaining the purpose of this class and why it's needed

src/network/requestQueue/index.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  createRequest({ entry, expectResponse, send: sendRequest }) {
+    const request = new SocketRequest({
+      entry,
+      expectResponse,
+      broker: this.broker,
+      requestTimeout: this.requestTimeout,
+      send: () => {
+        this.inflight.set(entry.correlationId, request)
```
ianwsperber: It's a little weird that the created SocketRequest could flag an inflight request without having been `push`ed on to the queue (meaning the request never gets removed). I'd either move this binding to within `push` or just remove this method from the public api, meaning we'd call `push` with `{ entry, expectResponse, send }`.

src/network/requestQueue/index.js
```diff
@@ -0,0 +1,117 @@
...
+        sendRequest()
+      },
+    })
+
+    return request
+  }
+
+  push(request) {
```
ianwsperber: Currently we advance through the queue essentially by chaining callbacks - when one requests succeeds/fails we add the next pending message to the queue. If for whatever reason an error or bug prevented the expected end condition from being met we could encounter some sort of a deadlock. I do think that in most cases the timeout will account for this, but it does make me wonder if it'd make sense to have some sort of "cleanup" process running on a setInterval to ensure the queue is advancing. Just a thought, I don't think it's necessarily something we have to implement now.

src/network/requestQueue/index.js
```diff
@@ -0,0 +1,117 @@
...
+        broker: this.broker,
+        correlationId: request.correlationId,
+      })
+
+      this.inflight.delete(correlationId)
+      request.completed({ size: 0, payload: null })
+    }
+  }
+
+  onResponse({ correlationId, payload, size }) {
```
ianwsperber: I find this name a little confusing, since `onWhatever` is usually reserved as a convention for event hooks ("`onResponseEvent(data, callback)`"). Might be more intuitive to name this "markRequestCompleted" or something

src/network/requestQueue/socketRequest.js
tulios: Right, I intended to prevent silly errors, but it's better to throw.

src/network/requestQueue/index.js
tulios: This case is covered, besides the timeout, all requests are rejected on socket errors/timeout (https://github.com/tulios/kafkajs/blob/master/src/network/connection.js#L116), the request queue will reject all inflight and pending requests.

src/network/requestQueue/socketRequest.js
tulios: 6556a3d

src/network/requestQueue/socketRequest.js
tulios: 8b1173d

src/network/requestQueue/index.js
tulios: Yeah, I wasn't happy with this. The new approach is better 866eb43

src/network/requestQueue/index.js
tulios: I changed to `fulfillRequest` 866eb43

tulios:
@ianwsperber @Nevon I addressed all the comments, please take another look

src/network/requestQueue/socketRequest.js
```diff
@@ -0,0 +1,134 @@
...
+   */
+  throwIfInvalidState({ accepted, next }) {
+    if (accepted.includes(this[PRIVATE.STATE])) {
+      return
+    }
+
+    const current = this[PRIVATE.STATE].toString()
+
+    throw new KafkaJSNonRetriableError(
+      `Invalid state, can't transition from ${current} to ${next.toString()}`
```
ianwsperber: So basically we only provide `next` for logging purposes, since we only validate `accepted`

tulios: Yes, just to improve the error message

PR #221, opened 2018-11-28
Handle receiving assignments for unsubscribed topics

Nevon:
If a consumer re-uses a group id from a consumer group with a different topic configuration we may receive member assignments for a topic that we're not subscribed to. We already attempted to handle this, but I don't think it was tested properly.

This also adds a new instrumentation event for the consumer, `consumer.crash`.

### DIFF
--- a/src/consumer/consumerGroup.js
+++ b/src/consumer/consumerGroup.js
@@ -166,7 +166,7 @@ module.exports = class ConsumerGroup {
     }
 
     // Check if the consumer is aware of all assigned partitions
-    for (let topic of assignedTopics) {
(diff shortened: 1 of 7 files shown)


src/consumer/consumerGroup.js
```diff
@@ -166,7 +166,7 @@ module.exports = class ConsumerGroup {
     }
 
     // Check if the consumer is aware of all assigned partitions
-    for (let topic of assignedTopics) {
+    for (let topic of keys(currentMemberAssignment)) {
```
tulios: Can you assign `keys(currentMemberAssignment)` to something like `safeAssignedTopics`? This will help us in the future

Issue #210, opened 2018-11-20
Consumer "read_uncommitted" isolation level


ianwsperber:
Support the "read_uncommitted" isolation level for transactional consumers. Since the default behavior is to "read_committed", we'll add the isolation level config option in this task.

- [x] Allow configuring isolation level at consumer level
- [x] Provide isolation level to `Fetch` call
- [x] Provide isolation level to `ListOffsets` call
- [x] Ensure that we do not ignore aborted transactions

ianwsperber:
@tulios Do you think it's better to provide an `isolationLevel=1` option to the consumer (0 or 1) or a `readUncommitted=false` option (true/false)? The former keeps us close to the Java API, the second is arguably more descriptive

ianwsperber:
Work on this is pretty much done, just waiting on #223 to get merged.

PR #223, opened 2018-11-29
EoS #202 Consumer "read_committed" default behavior

### LINKED ISSUE #202: Consumer "read_committed" isolation level
To support transactions consumers should have an option to read only committed messages. In the Java API this is accomplished using the "isolation_level" option. The naive translation would be to provide an `isolationLevel` flag to our `consumer()` call.

**Tasks**
- [x] Pass the isolation level to fetch requests
- [x] Pass isolation level to ListOffsets to retrieve the LSO ("last stable offset") (blocked by #201)
- [x] Do not return aborted commits to the user

See the [Java implementation](https://github.com/apache/kafka/blob/044350f28e4d0edcc0503b2bc2593c0d25948ed0/clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java#L1184) for a reference on how to ignore aborted commits. Requires maintaining a map locally of aborted PIDs.

Along with #197 & #201 this should be all the work necessary for transactional consumers. @tulios LMK if this all sounds right to you.

ianwsperber:
Resolves #202.

The meat of this is to update the `Batch` class to filter out aborted messages based on the returned `abortedTransactions` in the partition metadata. The `firstOffset`, `lastOffset` and `offsetLag` methods will reference the offsets for all messages received, regardless of whether they were filtered out. So now the consumer will by default filter out aborted messages - for #210 I'm planning to actually update the api to allow choosing between "read_committed" and "read_uncommitted".

To accomplish the filtering we needed a number of properties from the RecordBatch, so I decided to make all properties in the RecordBatch available on our individual messages via a "batchContext" property. I could foresee us needing to access the RecordBatch's properties again in the future, so I thought it best to make all its properties available here rather than continue to add new properties in a piecemeal fashion to the message object.

Had a bit of trouble with this one, there was a bug in the idempotent producer whereby it was sending the transactional id regardless of whether we were in a transaction, which tripped me up for a while as I tried to understand why the consumer wasn't receiving any messages!

~As an aside I've also been seeing some sporadic `CONCURRENT_TRANSACTIONS` errors, which have been difficult to reproduce. I would expect the `maxInFlightRequests` option to prevent this, but it hasn't seemed to stop it. I'll create an issue if I continue to encounter this problem.~ I think the issue was that "CONCURRENT_TRANSACTIONS" was marked as non-retriable, whereas I believe it [should be retriable](https://github.com/apache/kafka/blob/12f310d50e7f5b1c18c4f61a119a6cd830da3bc0/core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala#L153). I've updated the `retriable` flag and haven't seen the error again yet.


package.json
```diff
@@ -32,7 +32,7 @@
     "test:group:broker": "yarn test:local --testPathPattern 'src/broker/.*'",
     "test:group:admin": "yarn test:local --testPathPattern 'src/admin/.*'",
     "test:group:producer": "yarn test:local --testPathPattern 'src/producer/.*'",
-    "test:group:consumer": "yarn test:local --testPathPattern 'src/consumer/.*'",
+    "test:group:consumer": "yarn test:local --testPathPattern 'src/consumer/.*.spec.js'",
```
Nevon: This shouldn't be necessary 🤔 Did you see some issue?

src/consumer/__tests__/consumeMessages.spec.js
```diff
@@ -418,4 +418,207 @@ describe('Consumer', () => {
...
+      })
+
+      consumer = createConsumer({
+        cluster,
+        groupId,
+        maxWaitTimeInMs: 100,
+        logger: newLogger(),
+      })
+
+      jest.spyOn(cluster, 'refreshMetadataIfNecessary')
```
Nevon: I guess these spies aren't necessary

package.json
ianwsperber: @Nevon Yes, I added three fixtures defined as `.js` files, which caused jest to complain about files with no exported tests. Normally I'd define the fixtures as JSON but these objects have some values that need to be defined as buffers, which AFAIK would have been a pain in JSON. So I updated the test pattern to only look for `spec.js` files (which is the convention we use anyways).

src/protocol/error.js
```diff
@@ -320,7 +320,7 @@ const errorCodes = [
   {
     type: 'CONCURRENT_TRANSACTIONS',
     code: 51,
-    retriable: false,
+    retriable: true,
```
tulios: I'm planning to do a proper review tomorrow, but looking at the Kafka documentation this shouldn't be `true`.

https://kafka.apache.org/protocol.html#protocol_error_codes

tulios: Right, saw your comment now. Interesting

PR #228, opened 2018-12-04
Eos 210 consumer read uncommitted

### LINKED ISSUE #210: Consumer "read_uncommitted" isolation level

ianwsperber:
Resolves #210.

Add a `readUncommitted` flag to the consumer options which determines the isolation level for ListOffsets and Fetch requests. If we'd rather this option be `isolationLevel` or something of the sort that'd be fine as well.


src/index.js
```diff
@@ -95,11 +98,17 @@ module.exports = class Client {
     retry,
     allowAutoTopicCreation,
     maxInFlightRequests,
+    readUncommitted = false,
   } = {}) {
+    const isolationLevel = readUncommitted
```
tulios: I think this is the best option right now

tulios:
screen shot 2018-12-05 at 13 49 03

I think your test has a racing condition

ianwsperber:
Thanks for catching that @tulios. It appears that the control record is not always in the fetch respond immediately after we've committed/aborted the transaction. I believe this is because the transaction coordinator will write the control record to the leaders during the `EndTxn` request, but does not wait for the record to be replicated. I'm less familiar with the nuances of leader/replication behavior than you likely are, but that's my intuition for what's happening. Anyways I moved the fetch & assertion inside a `retry` statement, with the idea that we should _eventually_ see the control record. I haven't seen the failure since.

ianwsperber:
@tulios Can we merge this now?

Issue #207, opened 2018-11-16
Support committing offsets in transaction


ianwsperber:
#### Overview

Kafka transactions provide support for the "consume-transform-produce loop" by allowing transactional producers to commit offsets for their recently consumed messages. This allows a client to ensure that offsets are committed for consumed messages _only_ if the transaction is committed.

The spec recommends we enable support by adding a method "sendOffsets" (`sendOffsetsToTransaction`) to the transactional producer.  This method will accept a data structure mapping topic-partitions to offsets, then perform the requests `AddOffsetsToTxnRequest` and subsequently `TxnOffsetCommit`.

Currently the consumer will automatically commit offsets when reading messages. We will need to update the API to allow users to _disable_ this option, so that they may send the offsets themselves via the producer's new `sendOffsetsToTransaction` method.

Would love to hear your thoughts on how best to update the consumer to accommodate this need @tulios @Nevon

#### Tasks

- [x] Add method `sendOffsetsToTransaction`
- [x] Update consumer to allow disabling auto committing
- [x] Keep offsetManager in sync with sent offsets

#### Links
- [consume-transform-produce loop](https://cwiki.apache.org/confluence/display/KAFKA/KIP-98+-+Exactly+Once+Delivery+and+Transactional+Messaging#KIP-98-ExactlyOnceDeliveryandTransactionalMessaging-4.Theconsume-transform-produceloop) (steps 4.3 & 4.4)
- [Java Implementation](https://github.com/apache/kafka/blob/d71cb54672e63d2f0f4b999668336642a9a63a1d/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L673)

tulios:
> We will need to update the API to allow users to disable this option, so that they may send the offsets themselves via the producer's new sendOffsetsToTransaction method

Right, the idea behind `eachMessage` was to provide a convenient and batteries included method to consume messages. Do you think users will decide to commit in different situations there? The way I see it `eachMessage` will call `AddOffsetsToTxnRequest` if it succeeds and `TxnOffsetCommit` once the batch is over or based on the `autoCommit` configurations.

`eachBatch` has the capabilities to disable some of the automation so it should be fine to have something like that there.

WDYT?
and since you already have the use case for a transactional producer it will be nice to have your view of the problem as well, which type of code would you like to have?

sklose:
I like the idea of `eachMessage` providing this capability out of the box for simple use cases.

[...]

Issue #122, opened 2018-08-27
Add custom offset committing interface for consumer/runner


bwdeng:
**Feature**

Currently, the consumer only supports committing the resolved offset. It would be nice to support custom offset committing.

**Use case**

We are trying to do aggregation based on kafkajs, but the size of the working set is not fixed. For example, an input stream would look like this:

```
X0 -- X1 -- X2 -- X3 -- Y0 -- Y1 -- Z0 -- Z1 -- Z2 -- ...
```

We try to aggrege events of the same kind (agg([X0, X1, X2, X3]), agg([Y0, Y1]), agg([Z0, Z1, Z2, ...])), and only committing offset when the aggregated result has been saved. The problem is that until seeing a Y event, we don't know if all the X events have arrived. However, when we saw Y0, the resolved offset has already passed X3.

Having the ability to commit custom offsets would allow us to properly track the working progress. Thanks!

tulios:
Hi @bwdeng, I think you can solve your problem with the current feature set. If you switch to `eachBatch` you can use two properties in your favor. Set `eachBatchAutoResolve` to `false` so the consumer won't auto-resolve the offsets, and set `autoCommitThreshold` to 1 so you can control the commit flow. Example:

```javascript
{
  eachBatchAutoResolve: false,
  autoCommitThreshold: 1,
  eachBatch: async ({ batch, resolveOffset, heartbeat, commitOffsetsIfNecessary }) => {
    for (let message of batch.messages) {
      if (aggregate(message)) {
        lastOffset = message.offset
        await heartbeat()
      } else {
        await resolveOffset(lastOffset)
        await commitOffsetsIfNecessary()
      }
    }
  }
}
```

This is pseudo code and I'm assuming a lot, but let's go through it. You get the whole batch and can detect if you have the full or the partial set. You can force the commit with `commitOffsetsIfNecessary`. WDYT?

bwdeng:
Exactly, I was hoping that there's a way to commit the offset of a message outside of its original batch callback.

At first, I tried updating the offset through the admin interface. But got `The coordinator is not aware of this member` error. It seems like only the active member can update the offsets of their assigned partitions.

Currently, I have a flaky patch which exposes the underlying ConsumerGroup used by the Runner, and use it to update offsets:

```typescript
public async commit(topic: string, partition: number, offset: string): Promise<boolean> {
    const { groupId, generationId, memberId, coordinator } = this.consumerGroup; // the underlying consumer group
[...]

PR #232, opened 2018-12-10
EoS #207 SendOffsets

ianwsperber:
### Changes

**Features**
- New transactional method `sendOffsets`
- New consumer `run` option `autoCommit: boolean`
- New `eachBatch` callback option `uncommittedOffsets`

### Summary

This PR implements a new transactional method, `sendOffsets`, which allows us to conditionally commit offsets depending on the outcome of a transaction. Additionally we allow users to disable auto-committing of offsets, so that they may instead manually provide offsets to the `sendOffsets` method. We provide the a new method, `uncommittedOffsets`, in the `eachBatch` options, so that users may access offsets marked as resolved but not committed.

This review is somewhat preemptive, as I would like some feedback on the API changes thusfar, ~~and additionally suspect **there may be a race-condition in my test** "does not respect offsets sent by an aborted transaction".~~ **EDIT** I was incorrectly providing the autoCommit option to the consumer (which did cause a race condition).

If we agree on the interface for accessing uncommitted offsets, I think we should additionally add a new method to the consumer for committing arbitrary offsets, so that users have a way to manually commit offsets outside of transactions. I think that should be implemented in a separate ticket however.

A problem with the implementation as it exists is that the `sendOffsets` method will not communicate the result back to the offset manager, so resolved offsets will never be marked as committed. AFAIK the only real impact would be that we commit the same offsets unnecessarily, but it is misleading.

### DIFF
--- a/src/cluster/__tests__/committedOffsets.spec.js
+++ b/src/cluster/__tests__/committedOffsets.spec.js
@@ -7,10 +7,32 @@ describe('Cluster', () => {
     groupId = 'test-group-id'
   })
 
-  const testWithOffsets = offsets => {
-    it(`should return all committed offsets by group ${
-      offsets ? '' : '(no offset map provided)'
-    }`, async () => {
+  describe('if no offset map is provided', () => {
+    it('should return all committed offsets by group', async () => {
+      const cluster = createCluster({})
+      const topic = 'test-topic'
+      const partition = 0
+
+      expect(cluster.committedOffsets({ groupId })).toEqual({})
+
+      cluster.markOffsetAsCommitted({ groupId, topic, partition, offset: '100' })
+      cluster.markOffsetAsCommitted({ groupId: 'foobar', topic, partition, offset: '999' })
+
+      expect(cluster.committedOffsets({ groupId })).toEqual({ [topic]: { [partition]: '100' } })
+      expect(cluster.committedOffsets({ groupId: 'foobar' })).toEqual({
+        [topic]: { [partition]: '999' },
+      })
+    })
+  })
+
+  describe('if an offset map is provided', () => {
+    let offsets
+
+    beforeEach(() => {
+      offsets = new Map()
+    })
+
+    it('should return all committed offsets by group', async () => {
       const cluster = createCluster({ offsets })
       const topic = 'test-topic'
       const partition = 0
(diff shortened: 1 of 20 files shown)


src/consumer/runner.js
```diff
@@ -172,9 +174,14 @@ module.exports = class Runner {
         heartbeat: async () => {
           await this.consumerGroup.heartbeat({ interval: this.heartbeatInterval })
         },
+        /**
+         * Note that this method _does not_ heed the "autoCommit" option,
```
tulios: This is also used by `eachMessage` to enforce the autoCommit policies.

https://github.com/tulios/kafkajs/blob/b78461a4798907379f91b1435d19362cf3b54d55/src/consumer/runner.js#L159

ianwsperber:
@tulios There's already some retries on the `findGroupCoordinator` call within the `cluster` module, I've assumed that was sufficient. As you said may be best to see how things behave in a "production" setting first.

Ok, I'm going to play around with the code to see how I can get `sendOffsets` to keep the offset manager in sync. I'm worried that marking the offsets as committed when they've only been added to the transactions may also cause problems... but as along as the consume begins from the correct offset on subsequent runs it's fine.

ianwsperber:
@tulios Because the committed offsets live in the offset manager its pretty awkward sharing that state between the consumer and the producer. I'm going to attempt a refactor which hoists that state into the Kafka client, tracking commits per consumer id, so that we don't need to expose these details to the end user.

ianwsperber:
Completed a refactor to hoist state of committed transaction offsets into the Kafka client, but now we're getting the error `KafkaJSProtocolError: The producer attempted a transactional operation in an invalid state` (I believe when we call `transaction.commit()`) 😞

ianwsperber:
@tulios It turns out I was forgetting to await the call to `eosManager.sendOffsets`, which created another race condition 🙄 We're marking the offsets as "committed" when we call `sendOffsets` now. LMK if we're good to merge now!

src/cluster/__tests__/committedOffsets.spec.js
```diff
@@ -0,0 +1,46 @@
+const { createCluster } = require('testHelpers')
+
+describe('Cluster', () => {
+  let groupId
+
+  beforeEach(() => {
+    groupId = 'test-group-id'
+  })
+
+  const testWithOffsets = offsets => {
```
tulios: Since we only have 2 tests, let's split them into separate cases instead of using the function to generate the tests

tulios:
This PR also solves #122

ianwsperber:
@tulios For #122 to be solved we may still need to expose a `commit` method that allows providing the offsets to commit? The `commitOffsetsIfNecessary` method currently exposed in `eachBatch` will commit the most recent of all resolved offsets, which may not be what the user wants.

Issue #175, opened 2018-10-12
Expose low-level metrics


Nevon:
Currently there is no way for a user to collect low-level metrics on things like network calls (latencies etc.)

The way it's implemented in kafka-python is that they register a bunch of metrics in a local registry, and then expose a [pluggable interface](https://github.com/dpkp/kafka-python/blob/master/kafka/metrics/metrics_reporter.py#L6) that will be called when metrics are added/changed/removed, which would allow you to create a glue layer to for example Prometheus.

We could even include metrics on things that are technically already possible to collect, like batch size for example, so that the userland consumer code doesn't have to contain code to collect those metrics.

tulios:
We have decided to expose metrics through the instrumentation events; we can revisit this in the future.

PR #233, opened 2018-12-19
Add network instrumentation events

tulios:
This PR makes the request timeout configurable, and it introduces network instrumentation events to all clients (producers, consumers and admin). We want to write an easy to use prometheus/statsd/datadog plugin that collects metrics directly from KafkaJS, and we want to use the instrumentation events for that. Issue #175 describes what the official client does but after some research; we concluded that it was too much work and after talking to some people using the java client they mentioned that they don't use the built-in features and have separate integrations with the same providers.

I want to measure the impact of the request queue before defining a default value and making it enabled by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1249,6 +1249,34 @@ List of available events:
 
 * admin.events.DISCONNECT
 
+* admin.events.REQUEST
+  payload: {
+    `broker`,
+    `clientId`,
+    `correlationId`,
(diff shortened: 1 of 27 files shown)


ianwsperber:
Tests are failing with `Cannot find module '../network/InstrumentationEvents' from 'instrumentationEvents.js'`

PR #244, opened 2019-01-16
Validate consumer groupId

### LINKED ISSUE #243: Throw if groupId is missing on consumer creation
If you don't provide a `groupId` when creating a consumer, you won't get any error from the client, and only `Failed to find group coordinator` from the broker.

If `groupId` is missing when creating the consumer, we should throw a descriptive error.

Nevon:
This introduces a new devdependency, `jest-extended`. It's needed because jest doesn't natively support matching on both an error constructor as well as the message, so there was no way to differentiate between the missing groupId and the incorrect heartbeatInterval/sessionTimeout, other than manually catching and inspecting the error. See https://github.com/jest-community/jest-extended#tothrowwithmessagetype-message

Fixes #243

### DIFF
--- a/README.md
+++ b/README.md
@@ -567,7 +567,7 @@ async () => {
 
 ## <a name="consuming-messages"></a> Consuming messages from Kafka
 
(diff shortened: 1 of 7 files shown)


tulios:
@Nevon add something to the readme to highlight the need of the groupId.

Issue #36, opened 2018-02-20
React Native Support


nervouscat:
Hi There @tulios,

This Library looks very promising! 
I would like to ask you if your platform supports React Native as a client

Thank you in advance!

Nevon:
So the simple answer is no. `kafkajs` will not work in react native, as it uses Node modules such as `net`, which cannot be fully used in a browser context.

This begs the question though - what problem are you actually trying to solve by using kafkajs on the client side? We did a quick proof of concept by replacing `net` with a browserified version that relies on a proxy server, just because the idea was so crazy, and it actually worked. This still requires a server to act as a proxy, as browser don't offer support for raw sockets yet (though [there is a draft](https://www.w3.org/TR/raw-sockets/)), but it proves that it isn't _impossible_ to achieve.

joaomilho:
Brilliant

tommymsw:
We are currently encounter the same problem, our front-end is React , which need browser as client to consume kafka topics, and I think it is very nice if the kafkajs can support this function. We want to know do you have any plans to implement this feature? thanks very much.

zluo001:
Maybe it's crazy for running kafka client in browser, but it's really useful. Now days, all the solutions of produce or consume topics in browser require a proxy server. It makes the architecture become complicated and more latency of message delivery, while it led to consume more hardware resources also more money. It should be a tremendous progress in web client if this can be made. Thanks for your great job.

Nevon:
I personally think it's a cool idea, but this is not something that is being actively worked on.

We did a hacky PoC just to see if it's possible, and got it working. To elaborate on my previous comment, the problem is that the Kafka protocol is a custom binary protocol over TCP, which requires the client to have a raw socket connection. This is not a capability that the web has. This means that, as @zluo001 mentions, you need to run a proxy server that your client can establish a websocket connection to, which then proxies the messages to Kafka using regular TCP sockets.

For Android and iOS, you do have access to raw TCP sockets, AFAIK, although you'd have to expose that capability to react-native. For web it simply is not possible.

I think it's unlikely that we will support this use-case in core KafkaJS, but if someone who has a vested interest in this establishes and commits to a realistic plan for how to achieve this, I believe that we would just need to expose a pluggable interface for the network layer, which we could theoretically support. Given that, target-specific packages that just swap out the network layer could be built outside of KafkaJS core.

csnvrl:
[...]

PR #258, opened 2019-02-05
Added network connection instrumentation

KBInmarsat:
This adds a BROKER_CONNECTION event that is emitted whenever a new socket is established to a broker.

### DIFF
--- a/src/network/connection.js
+++ b/src/network/connection.js
@@ -6,6 +6,7 @@ const { KafkaJSConnectionError } = require('../errors')
 const { INT_32_MAX_VALUE } = require('../constants')
 const getEnv = require('../env')
 const RequestQueue = require('./requestQueue')
+const events = require('./instrumentationEvents')
 
 const requestInfo = ({ apiName, apiKey, apiVersion }) =>
   `${apiName}(key: ${apiKey}, version: ${apiVersion})`
@@ -53,6 +54,8 @@ module.exports = class Connection {
     this.broker = `${this.host}:${this.port}`
     this.logger = logger.namespace('Connection')
 
+    this.instrumentationEmitter = instrumentationEmitter
+
     this.ssl = ssl
     this.sasl = sasl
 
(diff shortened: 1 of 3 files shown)


KBInmarsat:
We need the socket object as we do a O/S specific setsockopt on the handle when connected to mark packets for network prioritisation so it's not something that should be in kafkajs.  I accept under normal circumstances you don't want to be exposing underlying objects.  The only other way would be to add to the API a way to specify a custom socket factory with the ability to override the default network\socket.js implementation, the disadvantage being it is a niche feature that would be exposed in the toplevel Kafka object API where as this instrumentation method is not.  If you prefer this way it can be implemented but if neither is acceptable we'ill just have to maintain our own fork as I can't see any other way of doing it.  Thanks..

Nevon:
I think the idea of a custom socket factory is actually a quite interesting one, as it could allow us to swap out the network layer to support targets other than NodeJS (this is fairly commonly requested, for example see #36). In the long run, I think it's the way to go.

@tulios, what do you think? I'm torn because I don't want to rush into an implementation of a swappable network layer, but I also don't want @KBInmarsat to have to maintain his own fork or to start to depend on an interface that we would eventually want to deprecate anyway (removing the socket from the instrumentation event).

tulios:
@KBInmarsat I like the idea of a factory function for the socket, I think you should be able to configure the socket in any way, and it's a clean solution. If you can keep this solution for a bit, we can come back with a proposal, something in the lines of what @Nevon mentioned, a way for you to provide a function that creates the socket.

Nevon:
I just submitted a PR for the `socketFactory` approach. #263 

I think it would solve your problem, @KBInmarsat. Let me know what you think.

KBInmarsat:
Thanks, I'ill try it out as soon as  I get chance. @Nevon

Nevon:
@KBInmarsat: Have you had a chance to try out the custom `socketFactory` approach in #263 yet? I've been using it for some experiments with a completely different network layer, and it's working out alright for me.

KBInmarsat:
I have not had a chance to yet but by coincidence I will be doing something with it tomorrow.

PR #263, opened 2019-02-06
Allow providing a socketFactory on client creation

Nevon:
Allows for lower-level configuration of the socket connection by
injecting a custom factory on instantiation. Default implementation is
in src/network/socketFactory.js, which is automatically injected by the
entrypoint.

I haven't documented it yet, but the idea is that if you need access to the socket or want to provide a socket implementation other than the default one, you implement a function like `({  host, port, ssl, onConnect  }) => Socket`, where `Socket` is an instance of an object compatible with [`net.Socket`](https://nodejs.org/api/net.html#net_class_net_socket)

Fixes #258

Contributes to #36

@KBInmarsat FYI


src/index.js
```diff
@@ -17,152 +17,155 @@ const PRIVATE = {
   OFFSETS: Symbol('private:Kafka:offsets'),
 }
 
-module.exports = class Client {
```
Nevon: All of this is just prettier deciding

tulios:
This PR is waiting for `v1.5.0` release

Nevon:
I'm hoping to spend some time on Operation Chrome Trident either this week or the next, so it'd be nice to get a 👍 and get this puppy merged, @tulios

tulios:
Yeah, I will merge this as soon as I release `1.5.1`

KBInmarsat:
Thanks for this, I've been using it in our code for the past few days an it works fine.

PR #283, opened 2019-02-13
Add default retry time to crash restart

### LINKED ISSUE #282: Missing default restart time on crashes due to retriable errors
#269 introduced a bug which is present in 1.5.0-beta.2 where if your consumer crashes with a retriable error, it will immediately restart the consumer with an error message like: `"[Consumer] Restarting the consumer in undefinedms"`

Nevon:
Fixes #282

### DIFF
--- a/src/consumer/index.js
+++ b/src/consumer/index.js
@@ -1,5 +1,6 @@
 const Long = require('long')
 const createRetry = require('../retry')
+const { initialRetryTime } = require('../retry/defaults')
 const ConsumerGroup = require('./consumerGroup')
 const Runner = require('./runner')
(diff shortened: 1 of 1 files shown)


src/consumer/index.js
```diff
@@ -203,11 +204,12 @@ module.exports = ({
       })
 
       if (e.name === 'KafkaJSNumberOfRetriesExceeded' || e.retriable === true) {
-        logger.error(`Restarting the consumer in ${e.retryTime}ms`, {
+        const retryTime = e.retryTime || retry.initialRetryTime || initialRetryTime
```
tulios: I think we need a growing `retryTime`, maybe we should embed this into RetriableError. We don't have to do this now, but this shouldn't be a fixed number.

PR #288, opened 2019-02-14
Validate message key and value before serialization

### LINKED ISSUE #281: Confusing error messaging when the key of the message isn't string
I was publishing a message like this:
```js
const kafka = new Kafka({
  clientId: 'my-app',
  brokers: ['localhost:9092']
});

const publish = async producer => {
  const d = { id: 100 };
  return producer.send({
    topic: 'data',
    messages: [{ key: d.id, value: JSON.stringify(d) }]
  });
};

const run = async () => {
  await producer.connect();
  const published = await publish(producer);
  console.log(published);
};

run();
```

I expect to see logs showing it's published, but actually I got the following error:
```
▸ node kafka.js
{"level":"ERROR","timestamp":"2019-02-13T15:53:30.372Z","logger":"kafkajs","message":"[Producer] \"value\" argument must not be a number","retryCount":0,"retryTime":357}
(node:29079) UnhandledPromiseRejectionWarning: TypeError: "value" argument must not be a number
    at Function.Buffer.from (buffer.js:186:11)
    at module.exports.string (/Users/leo/zhangchiqing/kafkajs/node_modules/kafkajs/src/producer/partitioners/default/murmur2.js:14:23)
    at /Users/leo/zhangchiqing/kafkajs/node_modules/kafkajs/src/producer/partitioners/default/index.js:34:25
    at messages.reduce (/Users/leo/zhangchiqing/kafkajs/node_modules/kafkajs/src/producer/groupMessagesPerPartition.js:7:23)
    at Array.reduce (<anonymous>)
    at module.exports (/Users/leo/zhangchiqing/kafkajs/node_modules/kafkajs/src/producer/groupMessagesPerPartition.js:6:19)
    at createProducerRequests (/Users/leo/zhangchiqing/kafkajs/node_modules/kafkajs/src/producer/sendMessages.js:42:38)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
(node:29079) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:29079) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

However, this change below fixes it, I just need to convert the `key` from a number into a string. 
```diff
-    messages: [{ key: d.id, value: JSON.stringify(d) }]
+    messages: [{ key: `${d.id}`, value: JSON.stringify(d) }]
```

The problem is that the error message was confusing. It seems like the `value` field was a number, but actually, it's the `key` field was number.

Environment: 
Node: v8.11.4
Version: kafkajs@1.4.7

Nevon:
Fixes #281

### DIFF
--- a/src/producer/index.spec.js
+++ b/src/producer/index.spec.js
@@ -76,6 +76,26 @@ describe('Producer', () => {
     )
   })
 
+  test('throws an error for messages with keys that cannot be converted to buffers', async () => {
+    producer = createProducer({ cluster: createCluster(), logger: newLogger() })
+
+    await expect(
+      producer.send({ acks: 1, topic: topicName, messages: [{ key: 1, value: 'value' }] })
+    ).rejects.toThrow(
+      `Invalid message for topic "${topicName}". "key" needs to be convertible to Buffer: {"key":1,"value":"value"}`
+    )
+  })
+
+  test('throws an error for messages with values that cannot be converted to buffers', async () => {
+    producer = createProducer({ cluster: createCluster(), logger: newLogger() })
+
+    await expect(
+      producer.send({ acks: 1, topic: topicName, messages: [{ value: 1 }] })
+    ).rejects.toThrow(
+      `Invalid message for topic "${topicName}". "value" needs to be convertible to Buffer: {"value":1}`
+    )
+  })
+
   test('allows messages with a null value to support tombstones', async () => {
     producer = createProducer({ cluster: createCluster(), logger: newLogger() })
     await producer.send({ acks: 1, topic: topicName, messages: [{ foo: 'bar', value: null }] })
--- a/src/producer/messageProducer.js
+++ b/src/producer/messageProducer.js
@@ -1,5 +1,6 @@
 const createSendMessages = require('./sendMessages')
 const { KafkaJSNonRetriableError } = require('../errors')
+const convertibleToBuffer = require('../utils/convertibleToBuffer')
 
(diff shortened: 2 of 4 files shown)


tulios:
I'm concerned about performance implications, calling `Buffer.from` 2x per message can impact the performance. Let's discuss a better solution.

Nevon:
Rather than validate the key and value, we'll coerce the key into a string if it's not already a buffer, same as we do for the message.

src/producer/index.spec.js
```diff
@@ -76,6 +76,26 @@ describe('Producer', () => {
...
+    )
+  })
+
+  test('throws an error for messages with values that cannot be converted to buffers', async () => {
+    producer = createProducer({ cluster: createCluster(), logger: newLogger() })
+
+    await expect(
+      producer.send({ acks: 1, topic: topicName, messages: [{ value: 1 }] })
+    ).rejects.toThrow(
+      `Invalid message for topic "${topicName}". "value" needs to be convertible to Buffer: {"value":1}`
```
WaleedAshraf: @Nevon If I'm not wrong, the value can be `null` also. In the case of the compacted topic, you need to send `null` message as a tombstone. https://kafka.apache.org/documentation/#design_compactionbasics

WaleedAshraf: I just noticed this was not merged.

PR #299, opened 2019-02-25
Abort old transactions on protocol error CONCURRENT_TRANSACTIONS

### LINKED ISSUE #296: Retry on CONCURRENT_TRANSACTIONS error for an idempotent transactional producer?
We are in the process of testing out different edge cases for the EOS implementation. One edge case we came across is: what happens if a service crashes before committing/aborting a transaction and it comes back up creating a new transaction before the previous transaction timeout elapsed.  We currently see a `CONCURRENT_TRANSACTIONS` error which crashes the service.

What would be the best way to handle this scenario:
1. the producer should have a way to abort the previous transaction so it doesn't have to wait the timeout period, 
2. kafkajs would automatically abort the previous transaction and retry the new transaction,
3. kafkajs doesn’t abort but retries the new transaction until success or 
4. the current behavior where it bubbles up as an error out of the producer and the caller of the producer has to do the retry itself (or handle the error in other ways)

Reproducer:

```
import * as ip from "ip";
import * as kafkajs from "kafkajs";
function sleep(ms: number) { return new Promise((resolve) => setTimeout(resolve, ms)); }

const waitTimeInSeconds = 60; // MODIFY THIS (60+ seconds required for transaction to timeout)

const id: number = Date.now();
const transactionalId = "transactional-id";
const host = process.env.HOST_IP || ip.address();
const broker: string = `${host}:9092`;
const topic = `test-topic-${id}`;
const acks = -1;
const maxInFlightRequests = 1;
const connectionTimeout = 1000;
const key = `sink-1-${id}`;
const payload1 = "1001";
const payload2 = "1002";
const messages = [
    { key, value: payload1 },
    { key, value: payload2 },
];
let adminInitialized = false;

async function doWrite(noCommit: boolean) {
    const client = new kafkajs.Kafka({
        clientId: "kafka-sink",
        brokers: [broker],
        connectionTimeout,
        maxInFlightRequests,
    });

    if (!adminInitialized) {
        const admin = client.admin();
        try {
            await admin.connect();
            await admin.createTopics({
                waitForLeaders: true,
                topics: [{ topic, numPartitions: 4, configEntries: [] }],
            });
        } finally {
            await admin.disconnect();
        }
        adminInitialized = true;
    }

    const producer = client.producer({
        // An idempotent producer enforces EoS messaging
        // idempotent: transactional ? true : false,
        idempotent: true,
        transactionalId,
    });
    await producer.connect();

    const transaction: kafkajs.Transaction = await producer.transaction();
    try {
        await transaction.send({
            acks,
            topic,
            messages,
        });
    } catch (e) {
        try {
            await transaction.abort();
        } catch (e) {
            throw e;
        }
        throw e;
    }

    if (!noCommit) {
        await transaction.commit();
        await producer.disconnect();
    }
}

async function doAll() {
    console.log("First doWrite");
    await doWrite(true);
    console.log(`Wait time of ${waitTimeInSeconds} seconds started: `);
    for (let ii = 0; ii < waitTimeInSeconds; ii++) {
        await sleep(1000);
        if ((ii + 1) % 5 === 0) {
            console.log(`${ii + 1} seconds passed`);
        }
    }
    console.log("Second doWrite");
    try {
        await doWrite(false);
    } catch (e) {
        console.log("ERROR:\n", e);
    }
    console.log("Done with doWrites");
}

doAll();
```

tulios:
Fixes #296 

This PR "aborts" old transactions when creating a new transaction, Kafka is responsible for the cleanup, so a simple retry is sufficient to terminate the old transaction (as indicated in the Java API https://github.com/apache/kafka/blob/201da0542726472d954080d54bc585b111aaf86f/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java#L1001-L1002).

I also took the opportunity to retry on other protocol errors, as indicated in the Java API.

### DIFF
--- a/src/producer/__tests__/concurrentTransaction.spec.js
+++ b/src/producer/__tests__/concurrentTransaction.spec.js
@@ -1,0 +1,52 @@
+const {
+  secureRandom,
+  newLogger,
+  createCluster,
+  testIfKafka_0_11,
+  createTopic,
+} = require('testHelpers')
+const createProducer = require('../index')
+
+describe('Producer > Transactional producer', () => {
+  let producer1, producer2, topicName, transactionalId, message
+
+  const newProducer = () =>
+    createProducer({
+      cluster: createCluster(),
+      logger: newLogger(),
+      idempotent: true,
+      transactionalId,
+      transactionTimeout: 100,
+    })
+
+  beforeEach(async () => {
+    topicName = `test-topic-${secureRandom()}`
+    transactionalId = `transactional-id-${secureRandom()}`
+    message = { key: `key-${secureRandom()}`, value: `value-${secureRandom()}` }
+
+    await createTopic({ topic: topicName })
+  })
+
+  afterEach(async () => {
+    producer1 && (await producer1.disconnect())
+    producer2 && (await producer2.disconnect())
+  })
+
+  describe('when there is an ongoing transaction on connect', () => {
+    testIfKafka_0_11('retries initProducerId to cancel the ongoing transaction', async () => {
+      // Producer 1 will create a transaction and "crash", it will never commit or abort the connection
+      producer1 = newProducer()
+      await producer1.connect()
+      const transaction1 = await producer1.transaction()
+      await transaction1.send({ topic: topicName, messages: [message] })
+
+      // Producer 2 starts with the same transactional id to cause the concurrent transactions error
+      producer2 = newProducer()
+      await producer2.connect()
+      let transaction2
+      await expect(producer2.transaction().then(t => (transaction2 = t))).resolves.toBeTruthy()
+      await transaction2.send({ topic: topicName, messages: [message] })
+      await transaction2.commit()
+    })
+  })
+})
--- a/src/producer/eosManager/index.js
+++ b/src/producer/eosManager/index.js
@@ -1,3 +1,4 @@
+const createRetry = require('../../retry')
 const { KafkaJSNonRetriableError } = require('../../errors')
 const COORDINATOR_TYPES = require('../../protocol/coordinatorTypes')
 const createStateMachine = require('./transactionStateMachine')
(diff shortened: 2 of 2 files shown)


tulios:
@sklose @plameniv Can you try this branch? I used your test case, and it works fine now.

plameniv:
I tried it with the replicator and it works fine. 

Only potentially confusing thing is that I still get a log of the error: 

`{"level":"ERROR","timestamp":"2019-02-25T18:55:31.535Z","logger":"kafkajs","message":"[Connection] Response InitProducerId(key: 22, version: 0)","broker":"10.51.18.152:9092","clientId":"kafka-sink","error":"The producer attempted to update a transaction while another concurrent operation on the same transaction was ongoing","correlationId":4,"size":20}`

Should there be some way to mark it as a retried error?

plameniv:
Can you point me to where in the code you are aborting the existing transaction?

tulios:
@plameniv, when the `CONCURRENT_TRANSACTIONS` error is retried, it cancels the previous ongoing transaction on the same `transactionId`, so I just included it to the list of retriable errors.
https://github.com/tulios/kafkajs/blob/d3d68cb1aa2f69ad2622627fdaed3a803c3cb84f/src/producer/eosManager/index.js#L15-L20

> Only potentially confusing thing is that I still get a log of the error

I think this is the correct behavior, even if we recover. The network layer logs the error, so unless it is something critical (or frequent), it's better to keep the behavior. Ideally, this shouldn't happen that often since producers would correctly abort transactions.

sklose:
thanks @tulios, we were confused because the code change looked like it is only retrying and we noticed the transaction went thru immediately now despite our 1 minute transaction timeout. I guess the broker does the abort on retry automatically and it's not the client's responsibility.

PR #313, opened 2019-03-12
Make the requestTimeout optional

tulios:
Add a new flag `enforceRequestTimeout` to turn on the request timeout system.

This change should allow us to have long run tests on the timeout system without impacting users. We received some issues that are complicated to investigate, so to unblock everybody it is better to make request timeouts optional.

Enabling request timeouts:

```javascript
const { Kafka } = require('kafkajs')

const kafka = new Kafka({
  // ...
  enforceRequestTimeout: true // default: false
  requestTimeout: 60000 // default: 30000
})
```


Nevon:
Having this default to `false` seems a bit odd and backwards incompatible. Wouldn't `true` mean that it behaves as before, since AFAIK we've always enforced request timeouts? Having the default be to wait forever seems bad.

tulios:
@Nevon it seems reversed but that was the way on < `1.5.0` versions, we had connection timeouts, but the client wasn't enforcing timeouts on each request. We introduced the individual timeouts on `1.5.0` but I'm not sure they are working as they should for everybody, that's why I'm disabling by default until we have confidence in the solution.

Nevon:
Oh that's interesting, but alright, then I'm with you.

PR #314, opened 2019-03-12
Add fetchTopicOffsets  method

patrykwegrzyn:
Hi there!! my first PR here.
Good job on the module , i love the interface, its been a pleasure to work with it.
I have tried every other kafka module  and this one is my fav :heart_eyes: .
I see `fetchTopicOffsets` to be very useful in situations when need to sync up topic up to the most recent offset.
A good example for that would be `ktable` implementation, let say emit `ready` event when  table is fully synced.
Let me know your thoughts


src/admin/index.js
```diff
@@ -173,6 +173,25 @@ module.exports = ({
...
+      throw new KafkaJSNonRetriableError(`Invalid topic ${topic}`)
+    }
+
+    const topicPartitions = await findTopicPartitions(cluster, topic)
+    const partitions = topicPartitions.map(partition => ({ partition }))
+    const broker = await cluster.findControllerBroker()
+    const offsets = await broker.listOffsets({ topics: [{ topic, partitions }] })
+
+    return offsets.responses
+      .find(response => response.topic === topic)
```
Nevon: What happens if the topic is not found by the broker? Are we guaranteed to get back a list containing the topic we asked for? I'm fine with us throwing an error, as long as it's not `TypeError: Cannot read property 'partitions' of undefined`.

src/admin/index.js
```diff
@@ -522,6 +541,7 @@ module.exports = ({
     getTopicMetadata,
     events,
     fetchOffsets,
+    fetchTopicOffsets,
```
Nevon: It doesn't matter much, but let's move this so that it's under `getTopicMetadata`, just so that they're grouped somewhat logically.

tulios:
Hi @patrykwegrzyn, thanks for the PR. I will take a look soon.

src/admin/index.js
```diff
@@ -173,6 +173,25 @@ module.exports = ({
     })
   }
 
+  /**
+   * @param {string} topic
+   */
+
+  const fetchTopicOffsets = async topic => {
```
Nevon: If we had a method that accepted a list of topics and returned a list of partition-offsets, what would we call that one?

I'm thinking about what the use cases for this would be, and I'm guessing a common usage would be to calculate lag. In that case, I would assume that it's not uncommon to be consuming more than one topic, in which case it would be nice to avoid people doing the following, since we could get information for all topics with a single call to the broker:

```javascript
const offsets = await Promise.all(async topics => admin.fetchTopicOffsets(topic))
```

Maybe we should have this method receive a list of topics instead? It feels a bit redundant to have something like:

```javascript
const fetchMultipleTopicOffsets = async (topics = []) => { // do stuff }
const fetchTopicOffsets = async topic => fetchMultipleTopicOffsets([topic])
```

src/admin/index.js
patrykwegrzyn: @Nevon good question i was blindly following [fetchOffsets](https://github.com/tulios/kafkajs/blob/master/src/admin/index.js#L181) as example i guess it suffers from the same issue

Nevon: Not really, because there it filters the array instead of using `find`, which means that even if the topic is not found, it still returns an array, which means that the `map` will always work, even if it just results in an empty array.

patrykwegrzyn: @Nevon not sure how this would make difference  tbh if topic is not found or don't exist `findTopicPartitions`  will throw error anyways ` This server does not host this topic-partition` this also apply to ` fetchOffsets`
as long as topic exist we should be good, topic will have at least 1 partition and it will satisfy arguments for `listOffsets` means we should always have `responses` with at least one partition ? unless `listOffsets` returns `[]` responses for what ever reason

Nevon: That's exactly what I wasn't sure of. If the methods will throw if the topic doesn't exist or can't be found, it's all good. I just wanted to make sure that we don't end up with a TypeError because of an invalid assumption. 👍

src/admin/index.js
patrykwegrzyn: @Nevon personally i am fan of atomic operations , ` Promise.all` and `fetchMultipleTopicOffsets` will do exactly the same background works to get the data.
its all depends on situation and what user needs.

Nevon: You're right. I didn't look very closely at what calls we would actually need to make, but they are indeed all scoped by a single topic, so in the end we would be doing the exact same work. If so, I think your API feels much nicer.

tulios:
@Nevon was faster than me; I will merge this after releasing `1.5.1`

PR #319, opened 2019-03-15
Update Consuming.md

abramjstamper:
Added async to demo source code where await was used without the async decorator


Nevon:
Good catch, and thanks for the patch!

Issue #303, opened 2019-03-01
Lock contention on high load


tulios:
From a comment on issue #177

> Since it only happens in specific load conditions, it's hard for me to reproduce it right away, but looking at the logs, I think it's from a producer.send call. You can find the > following snippet on how I'm initializing the producer and how I'm calling the send function.
> 
> ```javascript
> // Incoming data ... Can be considered something like this:
> for (let i = 0; i < 100000; i++)
>     queueMessage(message, 'specific-topic');
> 
> // Producer chunk ...
> const kafkaClient = new Kafka({
>     clientId: `SomeID`,
>     brokers: ['10.240.0.6:9092', '10.240.0.7:9092', '10.240.0.20:9092'],
>     connectionTimeout: 5000,
>     requestTimeout: 60000,
>     maxInFlightRequests: 200,
>     retry: {
>         initialRetryTime: 1000,
>         retries: 5
>     }
> });
> 
> const producer = kafkaClient.producer();
> 
> async function queueMessage(message, topic, retries = 0) {
>     try {
>         await producer.send({
>             topic,
>             // TODO -> KafkaJS probable bug? key=null should be present
>             messages: [{key: null, value: JSON.stringify(message)}],
>             acks: 1,
>             compression: kafkaUtils.SNAPPY
>         });
>     } catch (e) {
>         if (retries < 2)
>             setTimeout(() => queueMessage(message, topic, retries + 1), 2000 * (retries + 1));
>     }
> }
> ```
>
> I'm not using any Promise.all, and the flow of incoming messages is a stream of small messages, so I can't actually batch them together and send them, it's more like streaming a lot of > messages in realtime, for a messaging application.
> 
> And I'm using KafkaJS 1.4.7.

@AlirezaSadeghi I moved the investigation to a new issue

AlirezaSadeghi:
Hi @tulios 

Um, I'm still having all sorts of problems arising sometimes here and there in the processes, sometimes it's the lock contention error, sometimes it's the following errors:

```
{"level":"WARN","timestamp":"2019-03-11T01:21:22.932Z","logger":"kafkajs","message":"[RequestQueue] Response without match","clientId":"ps-8682-router","broker":"10.240.0.7:9092","correlationId":4278511}

{"level":"ERROR","timestamp":"2019-03-11T01:21:22.947Z","logger":"kafkajs","message":"[Producer] Request Metadata(key: 3, version: 4) timed out","retryCount":0,"retryTime":359}
```

[...]

Issue #257, opened 2019-02-04
Response without match


jafri:
Seeing lots of these errors
```
{"level":"WARN","timestamp":"2019-02-04T15:33:30.509Z","logger":"kafkajs","message":"[RequestQueue] Response without match","clientId":"client","broker":"localhost:29092","correlationId":145}
```

This is the command
```
          this.producer.send({
            topic: 'messages',
            idempotent: true,
            messages: [
              { value: JSON.stringify(message) }
            ],
          })
```

Sending 100K of these messages, and interestingly enough, I end up with anywhere between 110- 120K messages (retried 5 times).

tulios:
Hi @jafri, interesting. Are you using `1.5.0-beta.1`? And can you add here how are you posting these messages, is it a `Promise.all`, sequential, etc?

Thanks.

jafri:
Its a for loop using 1.5.0-beta.1, my newer tests show that messages are delivered properly when timeout is increased to 60 seconds.

I thought idempotent mode handles timeouts and prevents duplicates?

jafri:
```
for (let messages of messageBox) {
              this.producer.send({
                topic: 'messages',
                idempotent: true,
                messages: [
                  ...messages.map(message => ({ value: JSON.stringify(message) }))
                ],
              })
}
```

tulios:
I think this is a bug, the request queue should ignore timeouts when using `idempotent`. Can you increase the timeout for now, I will get this fixed on the next version (`1.5.0.beta.2`).

tulios:
Looking at your code now, can you add `await` to the `send` calls? This way you are generating a lot of concurrent promises

jafri:
I will look into it, but as the NodeJS app can easily handle the concurrent promises, I do believe there is an issue somewhere with the experimental idempotency guarantee.

tulios:
So, I ran some tests, and it looks fine. Since you are not awaiting `send` you are spawning a huge amount of promises at once (100k) and completely overloading the system. I used the example producer in the repo to test:

```javascript
const run = async () => {
  await producer.connect()
  for (let i = 0; i < 100000; i++) {
    await Promise.all([sendMessage(), sendMessage()])
  }
}

run().catch(e => console.error(`[example/producer] ${e.message}`, e))
```

`sendMessage` generates between 10 to 1000 messages, so I end up with ~1M published messages. 

```
[ { topicName: 'topic-test',
    partition: 1,
    errorCode: 0,
[...]

PR #323, opened 2019-03-27
Process a fixed number of lock releases per iteration on lock#release

tulios:
`lock#release` was blocking the whole process when we had a high number of 
locked resources (1000+), this PR should improve the results of #303 and #257


src/utils/lock.js
```diff
@@ -49,6 +50,6 @@ module.exports = class Lock {
   async release() {
     this[PRIVATE.LOCKED] = false
     const waitingForLock = Array.from(this[PRIVATE.WAITING])
-    return Promise.all(waitingForLock.map(acquireLock => acquireLock()))
+    return promiseAllInBatches({ batchSize: 20 }, waitingForLock)
```
Nevon: I understand what it's doing, but I'm trying to work through in my head why this would have an impact.

Say we have 1000 things that all want to acquire a lock that is currently locked. At that moment, we have 1000 functions in the wait list. Once the lock is released, we would call those 1000 functions. One of them would manage to acquire the lock and delete itself from the wait list. The other ones would do literally nothing. We now have 999 functions in the wait list. And so it repeats until all of them have either timed out or had a chance to acquire the lock.

With the new implementation, the only difference is that we only call the first 20 waiting functions instead of all 1000. This does save us some calls, but since we know only one of them can acquire the lock anyway, why don't we just pop the first one instead of calling 20 of them?

tulios: The difference is that we have 1000 `async` functions, which means we are creating a lot of promises and increasing memory and CPU usage. The system ends up in a huge pause and all the promises timeout once the event loop reaches the time queue, this change makes the system more stable even if the user is generating a lot of promises in userland code.

Nevon: Sure, that's true. Not to mention that when the lock releases the next time, you'll generate 998 promises, then 997, etc. Which is quite a large series. But the question remains, why even have 20 of them competing against each other? Why not just take a single one?

tulios: you have a point

tulios: @Nevon a3552cb is a lot more efficient now, we were creating yet another array with `Array.from` to convert the `Set`, now I am just using the iterator.

Nevon:
Looks good!

PR #324, opened 2019-03-27
Set up CI with Azure Pipelines [skip ci]

tulios:
Example:
https://dev.azure.com/tulios/kafkajs/_build/results?buildId=10


tulios:
@Nevon This is ready for review

The script that skips the tests for markdown files is not working, but it was failing on Travis as well. I'm checking with Microsoft on how to make it work with Azure pipelines, the problem is that the variable that they expose contains the merge commit and not the last commit of the branch, so we can't list the file changes.

Also, can you create an account on Azure DevOps so I can add you to the project?

azure-pipelines.yml
```diff
@@ -0,0 +1,199 @@
...
+
+variables:
+  - group: website_secrets
+  - name: KAFKA_VERSION
+    value: 1.1
+  - name: COMPOSE_FILE
+    value: docker-compose.1_1.yml
+
+####### Linter
+jobs:
```
Nevon: It's not something we need to do within this PR, but I wonder if we would see any benefit from not starting Kafka and Zookeeper in each job, but rather define them as container resources: https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#container-resource

It would be really neat if they could be re-used across each job, but I don't know if that's how that works, or if it'll boot all the resources in each job anyway.

Nevon:
Remove the travis config as well, since it's failing now anyway.

azure-pipelines.yml
tulios: Yeah, I considered that, but I already had a lot of changes. We can explore in another PR.

PR #325, opened 2019-03-29
feat: add admin.listTopics()

terebentina:
This PR adds a `listTopics` method to the admin client.

### DIFF
--- a/src/admin/index.js
+++ b/src/admin/index.js
@@ -66,6 +66,26 @@ module.exports = ({
     instrumentationEmitter.emit(DISCONNECT)
   }
 
+  const listTopics = async () => {
+    const retrier = createRetry(retry)
+
+    return retrier(async (bail, retryCount, retryTime) => {
+      try {
+        await cluster.refreshMetadata()
+        const broker = await cluster.findControllerBroker()
+        const metadata = await broker.metadata()
+        return metadata.topicMetadata.map(({ topic }) => topic)
+      } catch (e) {
+        if (e.type === 'NOT_CONTROLLER') {
+          logger.warn('Could not list topics', { error: e.message, retryCount, retryTime })
+          throw e
+        }
+
+        bail(e)
+      }
+    })
+  }
+
   /**
    * @param {array} topics
    * @param {boolean} [validateOnly=false]
@@ -554,6 +574,7 @@ module.exports = ({
   return {
     connect,
     disconnect,
+    listTopics,
     createTopics,
     deleteTopics,
     getTopicMetadata,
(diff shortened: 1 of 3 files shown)


src/protocol/requests/metadata/v4/request.js
```diff
@@ -12,6 +12,6 @@ module.exports = ({ topics, allowAutoTopicCreation = true }) => ({
   apiVersion: 4,
   apiName: 'Metadata',
   encode: async () => {
-    return new Encoder().writeArray(topics).writeBoolean(allowAutoTopicCreation)
+    return new Encoder().writeNullableArray(topics).writeBoolean(allowAutoTopicCreation)
```
terebentina: The protocol docs are saying: `If the topics array is null fetch metadata for all topics.`. It made sense to me to be specific here with the nullable array.

src/admin/index.js
```diff
@@ -66,6 +66,26 @@ module.exports = ({
     instrumentationEmitter.emit(DISCONNECT)
   }
 
+  const listTopics = async () => {
```
terebentina: Let me know if an optional regular expression argument makes sense here to filter and return just the topics matching that re. It can be handled outside the library as well.

tulios:
Hi @terebentina, thanks for both PRs. This is something I wanted to fix for a while; I will help you get this merged.

We already have a method called `getTopicMetadata`, which is the same but it also accepts a list of topics. The current implementation has a problem because it relies on the target group, which shouldn't be used for this purpose. Let's fix this method instead of adding a new one.

Your implemenation also suffers from the same problem, because the cluster is using the target topics internally, we can't rely on `refreshMetadata` for this function, we have to use whatever we get from `broker.metadata`. The controller broker works for this purpose but we could use any broker, so we shouldn't overload this specific broker, a better implementation could be:

```javascript
// on the cluster
async metadata({ topics: [] } = {}) {
  return this.brokerPool.withBroker(async ({ broker }) => this.retrier(async (bail, retryCount, retryTime) => {
    try {
      return broker.metadata(topics)
    } catch (e) {
      if (e.type === 'LEADER_NOT_AVAILABLE') {
        throw e
      }

      bail(e)
    }
  }))
}
```

terebentina:
@tulios I might not have time to look at this in April at all. So if you want to get crackin' with `getTopicMetadata` you should probably go ahead and do that and close my PRs.

tulios:
Sorry for the delay @terebentina, don't worry. My intention wasn't to pressure you; I will take this PR over and get it through. Thanks for the report.

tulios:
Similar code was merged on #331

PR #327, opened 2019-03-31
feat: add consume.patternSubscribe()

terebentina:
This PR adds a new consumer method to allow subscribing to multiple matching topics at once by using a wildcard character.
It doesn't watch the list of topics to auto-subscribe to new added topics - this is by design and, as far as I can tell, the same with the java consumer.

### DIFF
--- a/docs/Consuming.md
+++ b/docs/Consuming.md
@@ -24,7 +24,29 @@ await consumer.subscribe({ topic: 'topic-B' })
 // await consumer.subscribe({ topic: 'topic-C', fromBeginning: true })
 ```
 
-KafkaJS offers you two ways to process your data: `eachMessage` and `eachBatch`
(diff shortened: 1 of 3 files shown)


terebentina:
This is now updated to use the new `cluster.metadata()`

tulios:
Hi @terebentina, sorry for the delay. I've checked the Java client and that's the way to go, I will jump in and help you merge this. I have some suggestions, we can keep just `subscribe` and accept regexp instead, e.g:

```javascript
 await consumer.subscribe({ topic: /my-topic-(eu|us)$/ }) 
```

what do you think?

terebentina:
Hi @tulios, yes, both suggestions make sense. Actually, after having used patternSubscribe for a couple of weeks now, I already feel the need for a more powerful version so the regexp idea is especially attractive.
Let's go for it!

tulios:
Fixed in #346

Issue #320, opened 2019-03-20
Parallel message handling


yonigoyhman:
Hey, 

we are using 3rd party multi-threaded libraries and would like to be able to handle messages in parallel and only commit once a message was handled successfully.

These parallel handles might have different running time.

Going over your example, I have a few questions:

1. If a higher offset finishes processing before a lower one ,and we resolveOffset, will this offset be committed immediately (causing the all lower offsets that might have not finished processing be committed as well)?

2. If we fail processing a message, and use the retry mechanism, will the message be committed before actually processed successfully? What happens if process crashes before a successful retry?

3. If we pause a consumer due to back-pressure, and a re-balance occurs before we resume, will the consumer maintain its paused state?

Thanks.

tulios:
Hi @yonigoyhman, answering your questions:

> If a higher offset finishes processing before a lower one ,and we resolveOffset, will this offset be committed immediately (causing the all lower offsets that might have not finished processing be committed as well)?

`resolveOffset` will never commit; it only keeps a checkpoint of the offsets that should be committed. One thing to keep in mind is that you should never resolve offsets out of order, like, resolve 5, then 1, then 3. You can also only resolve the highest offset; you don't have to resolve every offset you process. The advantage of resolving every offset is that you can commit the results frequently, instead of only once. I'm assuming you are using `eachBatch`, so the best course of action is only to resolve once all processes finish.

> If we fail processing a message, and use the retry mechanism, will the message be committed before actually processed successfully? What happens if process crashes before a successful retry?

No. If it crashes, the resolved offsets will be committed, and the unresolved ones will come back again on the next batch. You should only resolve the offsets once they have been processed.

> If we pause a consumer due to back-pressure, and a re-balance occurs before we resume, will the consumer maintain its paused state?

Yes. Rebalances won't affect the paused state so that it will remain paused, but if the consumer crashes for some reason (usually connectivity issues, or issues on your callbacks), the consumer will be resumed. I think we can improve this and keep the consumer paused, but this is the current state.

We have a pattern for processing messages in parellel in our systems, we usually use a custom function called `promiseAllInBatches`, and the code looks like this:

```javascript
consumer.run({
  eachBatch: async ({ batch, resolveOffset, resolveOffset, commitOffsetsIfNecessary, heartbeat, isRunning }) => {
    const promises = batch.messages.map(message => () => processMessage(topic, partition, message))

    const shouldContinue = async (processedMessages, chunk, numberOfChunks) => {
[...]

PR #332, opened 2019-04-08
Concurrent message processing

### LINKED ISSUE #330: Partition aware concurrent `eachBatch`/`eachMessage`
Currently we are processing each batch in sequence, but there's no reason you couldn't process the batches concurrently, as long as you maintain order within each batch.

I would suggest adding a concurrency limit to the consumer, which would cause us to process a maximum of N batches or messages (depending on if the user is using `eachMessage` or `eachBatch`) concurrently.

This was initiated by @rdbenoit in https://github.com/tulios/kafkajs/issues/320#issuecomment-475396251

Nevon:
This implements a simple way to achieve partition-aware concurrency. This allows a consumer to consume messages from multiple partitions concurrently, while still ensuring ordering within each partition.

## Usage

```javascript
consumer.run({
  partitionsConsumedConcurrently: 3, // Default: 1
  eachMessage: async ({ topic, partition, message }) => {
      // This will be called up to 3 times concurrently
  }
})
```

In order to benefit from this kind of concurrency, the consumer needs to be subscribed to a topic with at least 2 partitions, and the work being done in `eachMessage`/`eachBatch` needs to be async (preferably I/O). I added a guideline in the docs to avoid people setting this to `Math.MAX_SAFE_INTEGER` and expecting their consumer to be super fast.

This is implemented by creating a kind of queue every time we finish a fetch. Each item in the queue is processed concurrently up to the specified limit. If one throws, we reject all the pending items. I did this to behave similarly to the existing behavior where we bail out of the loop on failure, but given that each item is a separate topic/partition, this is something we might be able to revisit in order to keep processing other partitions even if one of them fails.

Fixes #330.


src/utils/concurrency.js
```diff
@@ -0,0 +1,60 @@
...
+    waiting = []
+    semaphore = 0
+  }
+
+  const next = () => {
+    semaphore--
+    onChange(semaphore)
+
+    if (waiting.length > 0) {
+      const fn = waiting.pop()
```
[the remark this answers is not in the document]
Nevon: Fixed in 7069059

docs/Consuming.md
```diff
@@ -95,6 +95,25 @@ consumer.run({
...
+        // This will be called up to 3 times concurrently
+    },
+})
+```
+
+Messages in the same partition are still guaranteed to be processed in order, but messages from multiple partitions can be processed at the same time. If `eachMessage` consists of asynchronous work, such as network requests or other I/O, this can improve performance. If `eachMessage` is entirely synchronous, this will make no difference.
+
+The same thing applies if you are using [`eachBatch`](Consuming.md#each-batch). Given `partitionsConsumedConcurrently > 1`, you will be able to process multiple batches concurrently.
+
+A guideline for setting `partitionsConsumedConcurrently` would be `partitions < partitionsConsumedConcurrently < CPU cores`.
```
tulios: is this right? I think it is backwards right, `partitionsConsumedConcurrently` should be lower than `partitions`

Nevon: I rewrote this to actually make sense.

docs/Consuming.md
```diff
@@ -95,6 +95,25 @@ consumer.run({
...
+        // This will be called up to 3 times concurrently
+    },
+})
+```
+
+Messages in the same partition are still guaranteed to be processed in order, but messages from multiple partitions can be processed at the same time. If `eachMessage` consists of asynchronous work, such as network requests or other I/O, this can improve performance. If `eachMessage` is entirely synchronous, this will make no difference.
+
+The same thing applies if you are using [`eachBatch`](Consuming.md#each-batch). Given `partitionsConsumedConcurrently > 1`, you will be able to process multiple batches concurrently.
+
+A guideline for setting `partitionsConsumedConcurrently` would be that it should not be larger than the number of partitions consumed. Depending on whether or not your workload is CPU bound, it may also not benefit you to set it to a higher number than the number of logical CPU cores. A recommendation is to start with a low number and measure if increasing leads to higher throughput.
```
t-d-d: Is the number of CPU cores relevant for a Node application?  This is still single threaded processing.

t-d-d: Can I set it to a high number to always process all partitions in parallel?  Is the behaviour the same for any setting > # consumed partitions ?

Nevon: It depends on your workload. If the message processing is entirely synchronous, then this setting is completely irrelevant as all the work takes place on the main thread and will thus be executed sequentially regardless.

However, in most cases the work is not entirely synchronous. In many cases, we call remote APIs or read/write to a database, etc. In those cases, some of the work takes place outside of the main thread, such as when doing network I/O. If each message requires you to make a network call to some API, you would normally just sit around and wait while the remote service is doing some work, but with this concurrency model you can instead fire off multiple of these requests concurrently.

So then you might say that you would just set the concurrency limit to whatever number of partitions you have, in order to achieve maximum concurrency, but that may not actually increase throughput for multiple reasons. First, your downstream dependency might not appreciate getting that many concurrent requests and it's likely their performance may suffer. Even if that's fine, network or disk I/O is not infinitely parallelisable, and correlates with the number of cores you have.

While they're talking about connection pool sizing in this article, the principles are very much the same, and it's a really good high-level explanation: https://github.com/brettwooldridge/HikariCP/wiki/About-Pool-Sizing

So the complex answer is that it really depends on your workload and your dependencies. The recommendation is pretty conservative, because otherwise people will put a gigantic number in there thinking that "more is better" and end up getting lower throughput and potentially really bursty behavior. Throughput may very well increase with higher concurrency, given the right conditions, but it's better to start low and increase it over time while measuring the effect until you hit the sweet spot.

Nevon: Yes. You can set it to a number higher than the number of partitions, and the effective result will be the same as if you set it to the number of partitions.

tulios:
Fixes #320

Issue #316, opened 2019-03-13
Phantom Lag


sklose:
I am running into a weird issue where Kafka Manager is reporting a phantom lag for a consumer group:

![image](https://user-images.githubusercontent.com/546088/54296633-ab67fb80-458b-11e9-8a60-ce52669d1f82.png)

This happens when I reset the consumer group to earliest (-2) and then restart the consumer. The partitions that don't have any data in them never receive offset commits and the partition that shows a lag of 174 has a low watermark of 172 due to topic retention policy.

```
	Partition ID: 0
	High Watermark: 0
	Low Watermark: 0

	Partition ID: 1
	High Watermark: 172
	Low Watermark: 172

	Partition ID: 2
	High Watermark: 0
	Low Watermark: 0

	Partition ID: 3
	High Watermark: 0
	Low Watermark: 0

	Partition ID: 4
	High Watermark: 877
	Low Watermark: 726

	Partition ID: 5
	High Watermark: 0
	Low Watermark: 0

	Partition ID: 6
	High Watermark: 0
	Low Watermark: 0

	Partition ID: 7
	High Watermark: 0
	Low Watermark: 0
```

I am wondering if it would make sense for kafkajs to commit the low watermark for partitions it hasn't received any data from to prevent this edge case?

tulios:
I think it does, I've seen other libraries doing this. We should also improve `admin#resetOffsets` to resolve the offsets and commit.

tulios:
@sklose I can include this issue to the `1.6.0` release, we want to bump most of the protocol APIs and perform some clean-up

Nevon:
Fixed by #333

PR #333, opened 2019-04-08
fetchTopicOffsets should return low watermarks

sklose:
enhances the admin client's `fetchTopicOffsets` function to return the low + high watermarks for the topic. this is related to #316 

I kept the existing `offset` field so it's backwards compatible and added `high` + `low` (offset and high are equivalent)


src/admin/index.js
```diff
@@ -190,15 +190,36 @@ module.exports = ({
...
+        ])
+
+        const { partitions } = high.pop()
+        const offsets = partitions.map(({ partition, offset }) => ({
+          partition,
+          offset,
+          high: offset,
+        }))
+
+        for (const { partition, offset } of low.pop().partitions) {
```
Nevon: Took me a minute to understand that this was mutating `offsets`. How about moving this into the map above instead? Something like:

```javascript
const { partitions: lowPartitions } = low.pop()
const offsets = partitions.map(({ partition, offset }) => ({
  partition,
  offset,
  high: offset,
  low: lowPartitions.find(({ partition: lowPartition }) => lowPartition === partition).offset
}))
```

sklose: good point, fixed

Nevon:
Great, thanks!

PR #338, opened 2019-04-10
Renamed restricted keyword parameter

jacktuck:
Using a restricted keyword is considered bad practice and also causes an issue In [Terser](https://github.com/terser-js/terser):

```
ERROR in handlers/versions/activate-version.js from Terser
Unexpected token: operator (=) [./node_modules/kafkajs/src/producer/eosManager/transactionStateMachine.js:16,34][handlers/versions/activate-version.js:48083,54]
```

I created an issue on Terser project too but perhaps they handle it correctly already.
https://github.com/terser-js/terser/issues/329

### DIFF
--- a/src/producer/eosManager/transactionStateMachine.js
+++ b/src/producer/eosManager/transactionStateMachine.js
@@ -13,7 +13,7 @@ const VALID_STATE_TRANSITIONS = {
 module.exports = ({ logger, initialState = STATES.UNINITIALIZED }) => {
   let currentState = initialState
 
-  const guard = (object, method, { legalStates, async = true }) => {
+  const guard = (object, method, { legalStates, async: isAsync = true }) => {
     if (!object[method]) {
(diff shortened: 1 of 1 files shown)


Nevon:
Thanks for the contribution! I'm quite confident that this is a bug in terser, as `async` is not a reserved keyword, but we can make this change regardless as a workaround.

jacktuck:
@Nevon Thanks for the quick reply! Having this workaround merged would be great as I'm currently using my Fork.

jacktuck:
Thanks so much for the merge. Would it be possible to get an npm release as well? :)

Nevon:
Since we have some other unreleased changes in master I would like to first run master in production for a little while, but we should be able to cut a new version pretty soon. For now, you could pin your dependency to 77a7e03d443018503d8b40cffeb2d3bf582018de if you want to help test it out.

Nevon:
@jacktuck `v1.7.0` is out, including your fix!

PR #343, opened 2019-04-16
Update protocol APIs to latest versions of Kafka 1

tulios:
This PR update all the protocols used up to the latest version available in Kafka 1, I have excluded the update of `produce` and `fetch` since they are a bit more complicated, the update will come on another PR.

Some APIs changed a bit but nothing critical, the highlight is `JoinGroup`, which now accepts a `rebalanceTimeout`.


src/protocol/requests/joinGroup/index.js
```diff
@@ -1,4 +1,10 @@
 const NETWORK_DELAY = 5000
+
+const requestTimeout = ({ rebalanceTimeout, sessionTimeout }) => {
```
Nevon: This can be re-used in v0 as well, since `rebalanceTimeout` will be undefined.

PR #346, opened 2019-04-19
Add consumer pattern subscribe

### LINKED ISSUE #336: Consume wildcard topics
Something librdkafka and java clients support.

await consumer.subscribe({ topic: 'data-source-tables.*' });

And keeps up to date when new topics added :)

tulios:
Update `consumer#subscribe` to support regexp, e.g.:

```javascript
await consumer.subscribe({ topic: /topic-(eu|us)-.*)/i })
```

@terebentina I decided to create a new PR since the code would change 100%, from `patternSubscribe` to `subscribe`.

This PR is the outcome of the discussions in PR #327 and fixes #336


src/cluster/index.js
```diff
@@ -159,6 +159,24 @@ module.exports = class Cluster {
     }
   }
 
+  /**
+   * @public
+   * @param {string[]} topics
+   * @return {Promise}
+   */
+  async addMultipleTargetTopics(topics) {
```
Nevon: Doesn't matter much, but why not just name this `addTargetTopics`?

src/cluster/index.js
```diff
@@ -159,6 +159,24 @@ module.exports = class Cluster {
     }
   }
 
+  /**
```
Nevon: Make `addTargetTopic` delegate to `addMultipleTargetTopics`, since the implementation is exactly the same.

src/cluster/index.js
tulios: I think the version with the plural is hard to spot and prone to typos, I was experimenting with different alternatives before I settled for this name.

Issue #349, opened 2019-04-23
Handle offline replicas in metadata response


Nevon:
`offline_replicas` in the v5 metadata response is an array, but we're reading it as a single int32. This happens to work as long as there are no offline replicas, but if there are any, this crashes with a `RangeError`:

https://github.com/tulios/kafkajs/blob/6cd8c988852ceb08a43c698c23ea34f9a760906a/src/protocol/requests/metadata/v5/response.js#L41-L48

PR #350, opened 2019-04-23
Fix crash on offline replicas in metadata v5 response

Nevon:
`offlineReplicas` is an array of int32, not a single one.


tulios:
Fixes #349

PR #358, opened 2019-05-08
Make murmur2 implementation compatible with Java version

### LINKED ISSUE #357: KafkaJS and java producer partitioning returning different results
Producing message with the same key gets assigned to different partitions depending if app is using java or nodejs client. 

This creates problems when using kafka streams for joining those data - **meeting the co-partitioning requirement fails** even when message keys are the same.

I created an isolated test code which confirmed it. Simple java and nodejs app with partitioning using murmur2 code extracted from the latest kafkajs and java producer. The code looks the same, but results differ. Even checked that input bytes are equal. ¯\_(ツ)_/¯

Test runner result matches the partition number I'm seeing in Kafka broker...

Any idea why and can it be fixed? 

Our current workaround is to repartition all kafkajs based messages with java producer.. 

```
let key = '100:48069';
let numPartitions = 12;

node bytes {"type":"Buffer","data":[49,48,48,58,52,56,48,54,57]}
node murmur 1713501762
node hash 1713501762
node partition 6

java bytes 49 48 48 58 52 56 48 54 57 
java murmur 1009543857
java hash 1009543857
java partition 9
```
 
[partitioning test code.zip](https://github.com/tulios/kafkajs/files/3155784/partitioning.test.code.zip)

Nevon:
Fixes #357

There were two main issues. The first one being that Java does integer division, so `length4` would be `2.25` in JS vs `2` in Java, which meant that we would do an additional iteration in the loop compared to the Java version.

The second, and far more annoying issue, is that Java's int and Javascripts Number type behave quite differently, so we'd get wildly different results during multiplication of large numbers. I changed the implementation to use `Long`, which solved the problem. I'm not very happy about the part where I convert back to an int to do a zero fill right shift, but I found that Long's `k.shiftRightUnsigned(r)` behaves differently to `k >>> r` (Java), but maybe I misunderstood something.

I haven't updated the existing murmur2 tests, since they will now all fail. This feels like a super breaking change, so I'm not sure how we should roll it out. I'm opening this PR for the sake of discussion.

@pimpelsang

### DIFF
--- a/src/producer/partitioners/defaultJava/Test.java
+++ b/src/producer/partitioners/defaultJava/Test.java
@@ -1,0 +1,75 @@
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Based on:
+ * https://github.com/apache/kafka/blob/3cdc78e6bb1f83973a14ce1550fe3874f7348b05/clients/src/main/java/org/apache/kafka/clients/producer/internals/DefaultPartitioner.java#L69
+ * https://github.com/apache/kafka/blob/3cdc78e6bb1f83973a14ce1550fe3874f7348b05/clients/src/main/java/org/apache/kafka/common/utils/Utils.java
+ * 
+ * Used to validate the Javascript implementation of the Java-compatible murmur2
+ * hashing function
+ * 
+ * javac Test.java && java Test
+ */
+
+public class Test {
+    public static void main(String[] args) {
+        List<String> keys = Arrays.asList("0", "1", "128", "2187", "16384", "78125", "279936", "823543", "2097152",
+                "4782969", "10000000", "19487171", "35831808", "62748517", "105413504", "170859375");
+
+        for (String key : keys) {
+            byte[] keyBytes = key.getBytes();
+
+            int murmur = murmur2(keyBytes);
+            System.out.println("\"" + key + "\": " + murmur);
+        }
+
+    }
+
+    /**
+     * Generates 32 bit murmur2 hash from byte array
+     * 
+     * @param data byte array to hash
+     * @return 32 bit hash of the given array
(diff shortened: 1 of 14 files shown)


tulios:
I agree @Nevon, we can't change the current murmur implementation without breaking compatibility. I think we should create a `murmur2Java` implementation and a `defaultJava` partitioner that uses the new algorithm and then document that default doesn't follow the same logic as the Java partitioner, but it's easy to switch to `defaultJava`. When I read the issue I was sure that it was a number overflow, Long got into the project quite late. I think the murmur implementation was already in place and never got reviewed.

Nevon:
Sounds good. I'll go for that solution. Will update the PR.

Nevon:
Not sure why Github thinks I live in the future (my previous comment was actually posted before Tulio's...)

I've exposed both the default partitioner and the Java compatible one now, and documented how to use them and why you might want to:

```javascript
const { Partitioners } = require('kafkajs')
kafka.producer({ createPartitioner: Partitioners.JavaCompatiblePartitioner })
```

src/producer/partitioners/defaultJava/Test.java
```diff
@@ -0,0 +1,75 @@
+import java.util.Arrays;
+import java.util.List;
```
tulios: Can you ignore this file on the `.npmignore`, otherwise it will be in the bundle

Nevon: I thought I did. I accidentally fucked up git a bit and had to do some reflog madness, so it got lost 🙈

Nevon: Fixed in 12c92c0

PR #361, opened 2019-05-13
Destroy socket after closing connection

### LINKED ISSUE #360: ENOBUFS error
Hi everyone,
I noticed when I start many consumption/productions of message.
After +4000 messages, I get systematically this error :
```
error: [Kafka] KafkaJSNumberOfRetriesExceeded: Connection error: connect ENOBUFS 127.0.0.1:9094 - Local (undefined:undefined)
error: [Kafka] [Kafka-Connection] Connection error: connect ENOBUFS 127.0.0.1:9094 - Local (undefined:undefined)
error: [Kafka] [Kafka-BrokerPool] KafkaJSConnectionError: Connection error: connect ENOBUFS 127.0.0.1:9094 - Local (undefined:undefined)
error: [Kafka] [Kafka-BrokerPool] Failed to connect to seed broker, trying another broker from the list: Connection error: connect ENOBUFS 127.0.0.1:9094 - Local (undefined:undefined)
```
It loops on this error until I stop the program.
When I restart it, the problem disappear... until it reach again ~4-5k messages.

First I was thinking it was because my 1 broker fails because of the load, so I added 2 additionnal brokers. But even with 3 brokers I get the error for the 3 urls !

I think it's related to TCP temporary ports. When I type `netstat -n` I get many and many lines like that :
>  TCP    xx.xx.xx.xx:53851     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53853     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53854     xx.xx.xx.xx:9092      ESTABLISHED
  TCP    xx.xx.xx.xx:53856     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53857     xx.xx.xx.xx:9092      ESTABLISHED
  TCP    xx.xx.xx.xx:53858     xx.xx.xx.xx:9092      ESTABLISHED
  TCP    xx.xx.xx.xx:53859     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53860     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53861     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53862     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53863     xx.xx.xx.xx:9094      ESTABLISHED
  TCP    xx.xx.xx.xx:53864     xx.xx.xx.xx:9094      ESTABLISHED
  TCP    xx.xx.xx.xx:53871     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53872     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53873     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53874     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53875     xx.xx.xx.xx:9092      ESTABLISHED
  TCP    xx.xx.xx.xx:53876     xx.xx.xx.xx:9094      ESTABLISHED
  TCP    xx.xx.xx.xx:53877     xx.xx.xx.xx:9092      ESTABLISHED
  TCP    xx.xx.xx.xx:53878     xx.xx.xx.xx:9092      ESTABLISHED
  TCP    xx.xx.xx.xx:53879     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53885     xx.xx.xx.xx:9092      ESTABLISHED
  TCP    xx.xx.xx.xx:53888     xx.xx.xx.xx:9092      ESTABLISHED
  TCP    xx.xx.xx.xx:53890     xx.xx.xx.xx:9094      ESTABLISHED
  TCP    xx.xx.xx.xx:53891     xx.xx.xx.xx:9092      ESTABLISHED
  TCP    xx.xx.xx.xx:53892     xx.xx.xx.xx:9094      ESTABLISHED
  TCP    xx.xx.xx.xx:53893     xx.xx.xx.xx:9092      ESTABLISHED
  TCP    xx.xx.xx.xx:53898     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53899     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53903     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53904     xx.xx.xx.xx:9094      ESTABLISHED
  TCP    xx.xx.xx.xx:53906     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53907     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53909     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53910     xx.xx.xx.xx:9093      ESTABLISHED
[...]
  TCP    xx.xx.xx.xx:53912     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53913     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53914     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53915     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53916     xx.xx.xx.xx:9093      ESTABLISHED
  TCP    xx.xx.xx.xx:53923     xx.xx.xx.xx:9093      ESTABLISHED

I counted more than 39k lines.

In order to help you, I get similar error when I was calling too much an API : my ports allowed were saturated. I solve it by setting a keepAlive HTTP Agent (with Infinity free sockets).
Is that a similar issue with kafkaJs? Maybe you already know how to solve it?

Thanks in advance

Nevon:
Fixes #360

Just to be clear, I don't like this, but I'm not sure what the alternative is.

This diagram from wikipedia explains pretty well how closing a connection works:

Diagram showing a 4-way connection closing

What I think is happening in #360 is that the rate of connections being opened is higher than the rate of connections being closed. That is, when we call `Socket.end`, that initiates the connection closing handshake dance by sending a `FIN` packet. Once the server has acknowledged it, the client waits for some time until it actually closes the connection and hands the file descriptor back to the OS. I think the reason for this is that if you want to avoid some delayed packet being received by a subsequent connection.

`Socket.destroy` ignores this and just completely closes the connection immediately.

I also tried just awaiting the callback to `Socket.end`, hoping that it would get invoked when the handshake is done and the connection is actually closed, but that did not solve the problem. I'm not sure if it's just because the rate of opened connections is too high or if it's invoked just after sending the initial `FIN` packet.

My feeling right now is that the proper solution to this is to add a connection pool that bounds the total number of sockets somehow, so that you have to wait for a socket to become available before you can establish a new connection.

FYI @vykimo

### DIFF
--- a/src/network/connection.js
+++ b/src/network/connection.js
@@ -2,7 +2,7 @@ const createRetry = require('../retry')
 const createSocket = require('./socket')
 const createRequest = require('../protocol/request')
 const Decoder = require('../protocol/decoder')
-const { KafkaJSConnectionError } = require('../errors')
+const { KafkaJSConnectionError, KafkaJSTimeout } = require('../errors')
 const { INT_32_MAX_VALUE } = require('../constants')
 const getEnv = require('../env')
 const RequestQueue = require('./requestQueue')
@@ -195,8 +195,25 @@ module.exports = class Connection {
 
     this.logDebug('disconnecting...')
     this.connected = false
-    this.socket.end()
     this.socket.unref()
+
+    await new Promise((resolve, reject) => {
+      const timeout = setTimeout(
+        () =>
+          reject(
+            new KafkaJSTimeout(`Failed to close connection within ${this.connectionTimeout}ms`)
+          ),
+        this.connectionTimeout
+      )
+
+      this.socket.end(null, null, () => {
+        clearTimeout(timeout)
+        this.socket.destroy()
+        resolve()
+      })
+    })
+
+    this.socket = null
     this.logDebug('disconnected')
     return true
   }


Nevon:
I also experimented with listening for the `close` event before resolving `disconnect`, which also did not solve the problem.

vykimo:
A Problem I noticed too is that when openning rate is too high, packets seems never closing (or maybe they are but the openning is so high that we can't notice it) and then, overflowing happens.

Nevon:
This is a great article explaining the problem: http://www.serverframework.com/asynchronousevents/2011/01/time-wait-and-its-design-implications-for-protocols-and-scalable-servers.html

src/network/connection.js
```diff
@@ -195,8 +195,25 @@ module.exports = class Connection {
 
     this.logDebug('disconnecting...')
     this.connected = false
-    this.socket.end()
     this.socket.unref()
+
+    await new Promise((resolve, reject) => {
+      const timeout = setTimeout(
```
tulios: `setTimeout` is bad; that's the reason the request timeout is disabled by default. CPU will rise, and the process will be unresponsive depending on the number of close operations.

Nevon: Don't worry, this PR will not be merged. The current solution is not an option. I just opened the PR to start a discussion.

Nevon:
I'm quite convinced at this point that the problem is ephemeral port exhaustion on the client side due to the `WAIT_TIME`. Basically, even though we open and close the socket within the scope of a single message being processed, the socket can stick around for several minutes before it's released. I don't think that forcing the socket to be destroyed and re-used by another connection is an option, since we might get delayed packets ending up in the wrong place.

One way that some applications work around this is by having multiple IP addresses on the source destination. Each source ip gives you a number of available ports, but since the problem is the rate at which sockets are opened, that would only delay the problem.

Even if we could somehow allow the client to create an huge number of connections, eventually the destination would run out of ports, so that's not a solution.

So then we're back to the idea of limiting the creation of sockets. Instead of just directly creating a socket, we could enforce a limit and have the client wait for a socket to be released before we open a new one. This has two issues to solve:

1. Can we know when a socket is properly closed? I believe that the `close` event represents this, but I would need to verify it in code.
2. How do we know how many sockets we can create? This is configurable on the host, and differs across OSs. Maybe we can have a reasonable default and allow the user to override it through configuration if needed.

I'll do some experimentation with this approach when I have some time. I think it's the only reasonable solution outside of userland.

@vikymo: As a workaround for now, you could rate limit your consumer yourself until you hit an equilibrium. There are ways to modify how long the `WAIT_TIME` on the socket is, but you won't be able to get it so low that you can consume at top speed anyway.

Nevon:
I wrote a little test script to see how the socket events work, but I can't really seem to replicate the issue. I wanted to see if we would get the `close` event after the `TIME_WAIT` period, so we could use it to keep track of the number of still open sockets.

```javascript
const net = require('net')
const ip = require('ip')

const KEEP_ALIVE_DELAY = 60000
const host = ip.address()
const port = 9092

const log = event => (...args) =>
  console.log(`[${event} | ${new Date().toISOString()}]: ${JSON.stringify(args)}`)

const socket = net.connect(
  { host, port },
  log('connect')
)

socket.setKeepAlive(true, KEEP_ALIVE_DELAY)
socket.on('data', log('data'))
socket.on('end', log('end'))
socket.on('error', log('error'))
socket.on('timeout', log('timeout'))
socket.on('close', log('close'))
socket.on('drain', log('drain'))
socket.on('lookup', log('lookup'))
socket.on('ready', log('ready'))

const errorTypes = ['unhandledRejection', 'uncaughtException']
const signalTraps = ['SIGTERM', 'SIGINT', 'SIGUSR2']

errorTypes.map(type => {
  process.on(type, e => {
    try {
      console.log(`process.on ${type}`)
      console.error(e)
      socket.end()
      process.exit(0)
    } catch (_) {
      process.exit(1)
    }
  })
})

signalTraps.map(type => {
  process.once(type, () => {
    try {
      socket.end()
    } finally {
      process.kill(process.pid, type)
    }
  })
})

process.on('SIGHUP', () => {
  log('SIGHUP')('Closing socket')
  socket.end()
})
```

All I'm doing is establishing a connection to Kafka (doesn't really matter what it is - I just wanted something on the other end) and then on `SIGHUP` I end the connection.

```
[connect | 2019-05-15T08:33:06.113Z]: undefined
[ready | 2019-05-15T08:33:06.115Z]: undefined
[SIGHUP | 2019-05-15T08:33:17.686Z]: "Closing socket"
[end | 2019-05-15T08:33:17.690Z]: undefined
[close | 2019-05-15T08:33:17.690Z]: false
```

If I check what sockets are open for my PID using `ps aux | grep socket-test | grep -v grep | tr -s [:blank:] | cut -d' ' -f2 | tr -d '[:space:]' | xargs nettop -p`, I can see the connection is established and then disappears pretty much immediately after sending `SIGHUP`.

Using TCP dump, I can also see that the final `FIN` packets are sent just about immediately (The `.` flag I think is just heartbeats. `F` is `FIN`):

```
tcpdump -i lo0 -n tcp and port 9092 and 'tcp[tcpflags] & (tcp-fin|tcp-rst|tcp-ack) != 0'
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on lo0, link-type NULL (BSD loopback), capture size 262144 bytes
10:33:09.701798 IP 10.3.220.236.9092 > 10.3.220.236.52013: Flags [.], ack 2653344820, win 6379, length 0
10:33:09.701865 IP 10.3.220.236.52013 > 10.3.220.236.9092: Flags [.], ack 1, win 6379, options [nop,nop,TS val 582780015 ecr 582776435], length 0
10:33:10.807175 IP 10.3.220.236.9092 > 10.3.220.236.52013: Flags [.], ack 1, win 6379, length 0
10:33:10.807289 IP 10.3.220.236.52013 > 10.3.220.236.9092: Flags [.], ack 1, win 6379, options [nop,nop,TS val 582781118 ecr 582776435], length 0
10:33:11.869113 IP 10.3.220.236.9092 > 10.3.220.236.52013: Flags [.], ack 1, win 6379, length 0
10:33:11.869181 IP 10.3.220.236.52013 > 10.3.220.236.9092: Flags [.], ack 1, win 6379, options [nop,nop,TS val 582782177 ecr 582776435], length 0
10:33:13.076164 IP 10.3.220.236.9092 > 10.3.220.236.52013: Flags [.], ack 1, win 6379, length 0
10:33:13.076197 IP 10.3.220.236.52013 > 10.3.220.236.9092: Flags [.], ack 1, win 6379, options [nop,nop,TS val 582783380 ecr 582776435], length 0
10:33:14.167058 IP 10.3.220.236.9092 > 10.3.220.236.52013: Flags [.], ack 1, win 6379, length 0
10:33:14.167131 IP 10.3.220.236.52013 > 10.3.220.236.9092: Flags [.], ack 1, win 6379, options [nop,nop,TS val 582784464 ecr 582776435], length 0
10:33:15.269185 IP 10.3.220.236.9092 > 10.3.220.236.52013: Flags [.], ack 1, win 6379, length 0
10:33:15.269215 IP 10.3.220.236.52013 > 10.3.220.236.9092: Flags [.], ack 1, win 6379, options [nop,nop,TS val 582785564 ecr 582776435], length 0
10:33:16.284796 IP 10.3.220.236.9092 > 10.3.220.236.52013: Flags [.], ack 1, win 6379, length 0
10:33:16.284835 IP 10.3.220.236.52013 > 10.3.220.236.9092: Flags [.], ack 1, win 6379, options [nop,nop,TS val 582786578 ecr 582776435], length 0
10:33:17.285907 IP 10.3.220.236.9092 > 10.3.220.236.52013: Flags [.], ack 1, win 6379, length 0
10:33:17.285941 IP 10.3.220.236.52013 > 10.3.220.236.9092: Flags [.], ack 1, win 6379, options [nop,nop,TS val 582787578 ecr 582776435], length 0
10:33:17.687335 IP 10.3.220.236.52013 > 10.3.220.236.9092: Flags [F.], seq 1, ack 1, win 6379, options [nop,nop,TS val 582787979 ecr 582776435], length 0
10:33:17.687380 IP 10.3.220.236.9092 > 10.3.220.236.52013: Flags [.], ack 2, win 6379, options [nop,nop,TS val 582787979 ecr 582787979], length 0
10:33:17.688757 IP 10.3.220.236.9092 > 10.3.220.236.52013: Flags [F.], seq 1, ack 2, win 6379, options [nop,nop,TS val 582787980 ecr 582787979], length 0
10:33:17.688821 IP 10.3.220.236.52013 > 10.3.220.236.9092: Flags [.], ack 2, win 6379, options [nop,nop,TS val 582787980 ecr 582787980], length 0
^C
20 packets captured
26 packets received by filter
0 packets dropped by kernel
```

I would have expected the socket to stick around for a few minutes before seeing the final `FIN` and `ACK`, but that's not what I'm seeing, so I must be misunderstanding something or I'm not testing it properly.

Nevon:
Closing this PR. It's good background information, but I'm not actively working on this at the moment.

PR #362, opened 2019-05-13
Add in typescript declaration file.

talam:
I've added in a typescript declaration file and also updated the transactions documentation to correctly  reflect the setup process for transactional producer.

### DIFF
--- a/docs/Transactions.md
+++ b/docs/Transactions.md
@@ -17,20 +17,13 @@ You initialize a transaction by making an async call to `producer.transaction()`
 > - The producer must wait for acknowledgement from all replicas (acks=-1)
 > - The producer must have unlimited retries
 
-Configure the producer client with `maxFlightRequests: 1` to guarantee EOS.
+Configure the producer client with `maxInFlightRequests: 1` and `idempotent: true` to guarantee EOS. Configuring the two options will enable the settings mentioned above.
 
 ```javascript
 const client = new Kafka({
   clientId: 'transactional-client',
   brokers: ['kafka1:9092', 'kafka2:9092'],
 })
-const producer = client.producer({ maxInFlightRequests: 1 });
-```
-
-Setting `idempotent` to `true` will correctly configure the producer
-to use unlimited retries and enforce acks from all replicas
-
-```javascript
 const producer = client.producer({ maxInFlightRequests: 1, idempotent: true })
 ```
 
--- a/docs/Transactions.md
+++ b/docs/Transactions.md
@@ -14,24 +14,24 @@ You initialize a transaction by making an async call to `producer.transaction()`
 > Note: Kafka requires that the transactional producer have the following configuration to _guarantee_ EoS ("Exactly-once-semantics"):
 >
 > - The producer must have a max in flight requests of 1
-> - The producer must wait for acknowledgement from all replices (acks=-1)
-> - The producer must have unlimitted retries
+> - The producer must wait for acknowledgement from all replicas (acks=-1)
+> - The producer must have unlimited retries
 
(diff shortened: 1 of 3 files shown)


tulios:
@talam thanks for the PR

index.d.ts
```diff
@@ -0,0 +1,590 @@
...
+    clientId?: string;
+    connectionTimeout?: number;
+    authenticationTimeout?: number;
+    requestTimeout?: number;
+    enforceRequestTimeout?: boolean;
+    retry?: RetryOptions;
+    socketFactory?: ISocketFactory;
+    logLevel?: logLevel;
+    logCreator?: logCreator;
+    allowExperimentalV011?: boolean;
```
tulios: You can remove `allowExperimentalV011`; this is enabled by default since v1.4.0. The reason it is still here is that we have a lot of useful tests checking the migration from the 0.10 format to 0.11 and we haven't decided what to do with them, but we should remove this flag from the public interface.

index.d.ts
```diff
@@ -0,0 +1,590 @@
...
+    None = 0,
+    GZIP = 1,
+    Snappy = 2,
+    LZ4 = 3,
+    ZSTD = 4,
+}
+
+export const CompressionCodecs: {
+    [CompressionTypes.GZIP]: () => any;
+    [CompressionTypes.Snappy]: () => never;
```
tulios: Users of kafkajs can provide an implementation for the other codecs, how can we describe this here? (maybe any for all of theme)

index.d.ts
talam: removed.

index.d.ts
talam: changed this to a `var` that will allow others to implement their own codes instead of a `const`

docs/Transactions.md
```diff
@@ -14,24 +14,24 @@ You initialize a transaction by making an async call to `producer.transaction()`
 > Note: Kafka requires that the transactional producer have the following configuration to _guarantee_ EoS ("Exactly-once-semantics"):
 >
 > - The producer must have a max in flight requests of 1
-> - The producer must wait for acknowledgement from all replices (acks=-1)
-> - The producer must have unlimitted retries
+> - The producer must wait for acknowledgement from all replicas (acks=-1)
+> - The producer must have unlimited retries
 
-Configure the client with `maxFlightRequests: 1` to guarantee EOS.
+Configure the producer client with `maxFlightRequests: 1` to guarantee EOS.
```
Nevon: The producer needs to have both `maxInFlightRequests: 1` and `idempotent: true` to guarantee EOS, so I would recommend merging this sentence with the one below ("Setting `idempotent` to `true`...") and getting rid of this code block, so that you just have one sentence and one code block with everything you need to configure your producer for EOS.

Nevon: Also, it's `maxInFlightRequests`, not `maxFlightRequests`

Nevon:
This is an amazing effort. Thank you so much!

docs/Transactions.md
talam: adjusted the docs, thanks.

tulios:
I think Azure pipelines is having some issues; I will get this build green and merge. Thanks.

micaelmbagira:
Nice work!
@tulios any idea when this will be released?

tulios:
@micaelmbagira I will probably bundle this change with some other issues so it might take a little while. You can try it out by using the commit number:

```sh
yarn add https://github.com/tulios/kafkajs.git#054ef5042a99cca6af97bd90e43facadcc01dd3e
```

PR #374, opened 2019-05-31
Add forceRefresh option to getTopicMetadata

### LINKED ISSUE #205: Add option to make admin.fetchOffsets resolve the offsets
`await admin.fetchOffsets` might return -1 or -2 depending on the state of the consumer group. We could add an option to force it to resolve the offsets, so you get the earliest or the latest offset.

proposal:
```javascript
await admin.fetchOffsets({ groupId, topic, resolveOffsets: true })
```

expelledboy:
Not sure how to unit test

Resolves #205


tulios:
Hi @expelledboy , `getTopicMetadata` is deprecated, you can take a look at `fetchTopocMetadata`. It does the same thing without the issues of previous method.

https://kafka.js.org/docs/admin#a-name-fetch-topic-metadata-a-fetch-topic-metadata

expelledboy:
@tulios Cool I can move the option, but do you like the solution?

tulios:
This won't be necessary since the new method always fetch metadata, so it doesn't use the cluster cache. Do you think we should document the depreciation better? I haven't included a log message but perhaps that's a better solution.

tulios:
Issue #205 is about resolving consumer groups offsets, the might be assign special values, like -1 or -2.

expelledboy:
@tulios does it always fetch metadata? Please see the latest commit, perhaps I am barking up the wrong tree then.

tulios:
@expelledboy yes, it will always fetch the metadata

Nevon:
Closing as inactive.

PR #377, opened 2019-06-05
Link to custom logger page

mpfeil:
The previous link was a non-existing anchor. Changed it to the Custom Logger docs.


Nevon:
Thanks for the fix!

Issue #378, opened 2019-06-05
Add way to commit offset with Consumer


JaapRood:
In various other clients, [like the official KafkaConsumer](https://kafka.apache.org/10/javadoc/?org/apache/kafka/clients/consumer/KafkaConsumer.html), arbitrary offsets can be committed through an interface like `consumer.commit`. So far, the only way to do this in KafkaJS is through `commitOffsetIfNecessary` function exposed as part of the `eachBatch` mechanism.

This limitation prevents the committing of offsets before a message has been consumed, which is a time and place where in stream processing tasks it's desirable to do so. While in most cases a `consumer.seek` operation is desirable, there are situations where you just want to set the progress made and not consume any messages at all. 

Another example is actually present in [the admin client's 'setOffsets' method](https://github.com/tulios/kafkajs/blob/2685df7a888c8400fe9378989dbc1ad29b76ac9f/src/admin/index.js#L292-L342), where a seek and then processing of messages must happen in order for the auto commit to eventually commit the right offset.

## Proposed solution

Introduce a `consumer.commit` or `consumer.commitOffsets` method, like the official `KafkaConsumer` does. Like `consumer.seek`, it could limit it's calling to only after `consumer.run` has established the internal `consumerGroup`.

JaapRood:
In [a thread on slack](https://kafkajs.slack.com/archives/CF6RFPF6K/p1559726888001700), @tulios  expressed some concerns about how the fetching mechanism might be affected:

>  Another problem is that the consumer loop only fetches the offsets in certain conditions, it always increments the offsets and commits, so if you commit outside of the loop it might be quickly overridden

But also seemed confident the following would fix that:

> when resuming a consumer it should load the offsets again

JaapRood:
As with #373 I'm happy to take give this a go, but will be delayed in doing so as we're still trying to see how much of our stream processing tasks we can get running against KafkaJS, before going back and fixing this. I'll be sure to announce it here when I get around to it, hoping others will too if they get around to it.

Hann:
Hi. Is there a release plan included this feature? 
I found the feature on the docs but It seems to be unreleased. 
 https://kafka.js.org/docs/consuming#a-name-manual-commits-a-manual-committing

JaapRood:
The release isn't out yet, but it should be in the next one. Meanwhile, you can depend on the commit sha  (see https://docs.npmjs.com/cli/install). There is no build step for the repo and it's how I run earlier releases.

[...]

PR #381, opened 2019-06-06
Fetch offsets on consumer resume

tulios:
This PR contributes to #378; it makes the consumer re-fetch the offsets for the recently resumed topics.

@JaapRood with this PR you can keep a reference of `commitOffsetsIfNecessary` and commit with the paused consumers, once they resume the offsets will be re-fetched.


src/consumer/__tests__/pause.spec.js
```diff
@@ -179,7 +179,7 @@ describe('Consumer', () => {
       const [pausedTopic, activeTopic] = topics
       consumer.pause([{ topic: pausedTopic }])
 
-      waitForConsumerToJoinGroup(consumer)
+      await waitForConsumerToJoinGroup(consumer)
```
Nevon: 🤦‍♂

src/consumer/__tests__/seek.spec.js
```diff
@@ -95,8 +95,7 @@ describe('Consumer', () => {
       consumer.run({ eachMessage: async event => messagesConsumed.push(event) })
       consumer.seek({ topic: topicName, partition: 0, offset: 1 })
 
-      waitForConsumerToJoinGroup(consumer)
-
+      await waitForConsumerToJoinGroup(consumer)
```
Nevon: I'm amazed that these tests ever passed.

src/consumer/subscriptionState.js
```diff
@@ -23,4 +25,15 @@ module.exports = class SubscriptionState {
...
   }
+
+  /**
+   * @returns {Array<string>} resumed topics
+   */
+  resumed() {
+    return Array.from(this.resumedTopics.values())
+  }
+
+  ackResumed() {
```
Nevon: This feels like it could be error prone, but I don't have a better suggestion. It just feels a bit arbitrary when the resumed topics are no longer considered resumed. I can't even really put into words what makes a topic not resumed anymore.

Nevon:
Good boyscouting of the tests!

src/consumer/subscriptionState.js
tulios: Yeah, resume and pause aren't async operations, so I can't do it there. Then I needed to flag that some topics were just resumed, and this should only happen once. The code is small and straightforward to read; we can revisit this later.

PR #389, opened 2019-06-13
Include fetched offset metadata retrieved with admin.fetchOffsets

JaapRood:
Kafka supports a nullable string of `metadata` to be stored with a committed offset. KafkaJS actually supports this with it's protocol definitions, but `admin.fetchOffsets` filters it out just before returning the retrieved offsets to us.

### DIFF
--- a/src/admin/__tests__/fetchOffsets.spec.js
+++ b/src/admin/__tests__/fetchOffsets.spec.js
@@ -43,7 +43,7 @@ describe('Admin', () => {
         topic: topicName,
       })
 
-      expect(offsets).toEqual([{ partition: 0, offset: '-1' }])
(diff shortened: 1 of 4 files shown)


JaapRood:
As you might see from the tests: it only asserts metadata of `null` is forwarded. While `consumerGroup` will pass on any metadata and write it, that method isn't exposed anywhere at the moment. `admin.setOffsets` actually spawns a `Consumer` and relies on the auto commit mechanism to commit a resolved offset after seeking to it. See #378.

Nevon:
Can you make sure to update the return type for `fetchOffsets` as well? https://github.com/tulios/kafkajs/blob/master/index.d.ts#L272

PR #392, opened 2019-06-13
Allow offset metadata to be written as part of OffsetCommit requests

JaapRood:
Following #389, where we can retrieve the `nullable string` metadata with an offset, it would make a lot of sense if it could be written, too.


JaapRood:
The unit tests still pass, but it would probably be a good idea to cover the case where `metadata` is actually specified. Not sure how that's to be expressed / structured with how the fixtures are defined, though.

JaapRood:
I'm not too sure why those consumer tests would fail as a result of the changes made. Flaky tests, or actually related?

Nevon:
Flakiness 😢

PR #393, opened 2019-06-13
Resume consuming from currently resolved offset instead of committed offset

JaapRood:
In running against the version that includes #381 (which this partially reverts) our tests highlighted an issue with the concept of resuming from the committed offset upon resuming consuming for a partition. This prevents the `pause` and `resume` methods to be used for controlling the consumption flow / rate. In addition, it assumes that broker stored offsets are used and a message is committed in this way before pausing the fetching of messages can be resumed.

The documentation for the official `KafkaConsumer` highlights under "Usage Examples" how pause and resume are useful for managing the consumption rate (implementing back-pressure while allowing concurrent processing, etc.) In addition, it also highlights why keeping your commits out of Kafka can be a good idea too (e.g. writing it atomically to a store where the results are written).


JaapRood:
@tulios I remember you adding this in #381 for addressing something related to #378, but I'm not sure exactly what. I suggest we figure that out as part of that issue separately and prevent this from being broken in the meantime :)

JaapRood:
Given the changes made (not related to Admin), I wouldn't be surprised if these test failures were due to flakiness 🤔.

tulios:
@JaapRood you are right on the resume/seek case, I missed the fact that you control your offsets, and it makes sense to support that since it's standard behavior.

I will re-run the admin tests

Issue #383, opened 2019-06-09
Large batches take exponentially longer to encode


colinfindlay-nz:
Hi,

We're currently seeing issues with very large message batch sizes.
If we send a message[] of approx 1,000 messages - it's taking about 200 milliseconds to encode. With 10,000 messages it's taking a lot longer - as much as 12 seconds in my rough tests.

I believe the problem is with how the RecordBatch encoder iterates over arrays:
```
  if (compression === Compression.None) {
    batchBody.writeArray(records)
  } else {
    const compressedRecords = await compressRecords(compression, records)
    batchBody.writeInt32(records.length).writeBuffer(compressedRecords)
  }
```

```
const compressRecords = async (compression, records) => {
  const codec = lookupCodec(compression)
  const recordsEncoder = new Encoder()

  for (let record of records) {
    recordsEncoder.writeEncoder(record)
  }

  return codec.compress(recordsEncoder)
}
```

In both instances this ends up in `encoder.writeEncoder(value)` - which performs a simple `Buffer.concat` - unfortunately this doesn't scale when callled in a loop, resulting in large amounts of buffer cloning.

A quick proof-of-concept hack to take an array and write it as a single concat, rather than many showed a performance improvement of about 1200% 

Is it possible to improve this?

Nevon:
Yes, absolutely! The encoder very naively copies buffers constantly, which is very inefficient. I've been wanting to replace the buffer usage with something that over-allocates buffer space, allowing us to add to the buffer without necessarily cloning the whole thing, but I haven't found the time to drive the change to completion.

If you believe you can turn your proof-of-concept hack into a decent PR, then that's a good quick win.

tulios:
WeHi @colinfindlay-nz, it was a conscious decision to use `Buffer.concat` on the encoder implementation, we wanted to nail the first implementation, and it was easier to achieve that with `concat`. We are aware that the performance can be improved, and most of the code that manipulates buffers are concentrated on the `Encoder` class. We had a discussion some months ago about tackling this, but we decided that it wasn't the time.

We have an extensive code coverage on the protocol parts, so if you want to give it a try. We will improve the performance of the encoder; it's just a matter o priority. We have some requirements on the new authentication methods so that we might look at this again after that.

colinfindlay-nz:
[...]

PR #394, opened 2019-06-14
Update encoder to write arrays as single buffer.concat where possible

colinfindlay-nz:
https://github.com/tulios/kafkajs/issues/383

This is not an all-encompassing performance improvement - I've allowed RecordBatches to use a more efficient encoding method while trying to retain all the existing behaviour.
Unit tests all seem to pass - and overall performance is about 10x faster for record arrays of 1000 or more


tulios:
Great work @colinfindlay-nz, thanks a lot!

PR #397, opened 2019-06-14
Prevent the consumption of messages for topics paused while fetch is in-flight

JaapRood:
We ran into this bug, where if you pause a topic in the time between a requesting a fetch and receiving a response for it, the received messages are consumed, despite being the topic having been paused. A classic race condition, easily fixed by filtering for paused topics once again at response time.

### DIFF
--- a/src/consumer/__tests__/consumeMessages.spec.js
+++ b/src/consumer/__tests__/consumeMessages.spec.js
@@ -12,6 +12,7 @@ const {
   newLogger,
   waitFor,
   waitForMessages,
+  waitForNextEvent,
   testIfKafka_0_11,
   waitForConsumerToJoinGroup,
   generateMessages,
@@ -654,7 +655,7 @@ describe('Consumer', () => {
     consumer.pause([{ topic: topicName }])
     await producer.send({ acks: 1, topic: topicName, messages }) // trigger completion of fetch
 
-    await sleep(200)
+    await waitForNextEvent(consumer, consumer.events.FETCH)
 
     expect(offsetsConsumed.length).toEqual(messages.length)
   })
--- a/src/consumer/__tests__/consumeMessages.spec.js
+++ b/src/consumer/__tests__/consumeMessages.spec.js
@@ -609,6 +609,57 @@ describe('Consumer', () => {
     })
   })
 
+  it('discards messages received when pausing while fetch is in-flight', async () => {
+    consumer = createConsumer({
+      cluster: createCluster(),
+      groupId,
+      maxWaitTimeInMs: 1000,
+      logger: newLogger(),
+    })
+
+    const messages = Array(10)
+      .fill()
+      .map(() => {
+        const value = secureRandom()
+        return { key: `key-${value}`, value: `value-${value}` }
+      })
+    await producer.connect()
+    await producer.send({ acks: 1, topic: topicName, messages })
+
+    await consumer.connect()
+
+    await consumer.subscribe({ topic: topicName, fromBeginning: true })
+
+    const sleep = value => waitFor(delay => delay >= value)
+    let offsetsConsumed = []
+
+    const eachBatch = async ({ batch, heartbeat }) => {
+      for (const message of batch.messages) {
+        offsetsConsumed.push(message.offset)
+      }
+
+      await heartbeat()
+    }
+
+    consumer.run({
+      eachBatch,
+    })
+
+    await waitForConsumerToJoinGroup(consumer)
+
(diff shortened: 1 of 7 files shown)


src/consumer/__tests__/consumeMessages.spec.js
```diff
@@ -609,6 +609,57 @@ describe('Consumer', () => {
...
+    consumer.run({
+      eachBatch,
+    })
+
+    await waitForConsumerToJoinGroup(consumer)
+
+    await waitFor(() => offsetsConsumed.length === messages.length, { delay: 50 })
+    await sleep(50)
+
+    // Hope that we're now in an active fetch state? Something like FETCH_START might help
```
Nevon: This looks like it's gonna be extremely flaky. We should indeed add a new instrumentation event.

src/consumer/__tests__/consumeMessages.spec.js
```diff
@@ -609,6 +609,57 @@ describe('Consumer', () => {
...
+    const seekedOffset = offsetsConsumed[Math.floor(messages.length / 2)]
+    consumer.pause([{ topic: topicName }])
+    await producer.send({ acks: 1, topic: topicName, messages }) // trigger completion of fetch
+
+    await sleep(200)
```
Nevon: Rather than sleeping an arbitrary amount and hoping we've consumed the message by then, use the `waitFor` utility.

JaapRood: The point is that no message has been consumed, that nothing has happened, so that makes a `waitFor` a bit trickier! Happy to hear if you've got any suggestions for that :)

src/consumer/__tests__/consumeMessages.spec.js
JaapRood: An instrumentation event would definitely be better, however in all past test runs since the first version of this was introduced (in #367) I haven't found a trace of failing once. That said, if Kafka responds a bunch slower through doing some compaction or whatever, the timing is most likely hosed.

JaapRood:
Should have some time to look at making the test a bit more robust later today :)

src/consumer/__tests__/consumeMessages.spec.js
Nevon: Ah, I should have read it more properly. Perhaps there's an event you could wait for having happened, such as `END_BATCH_PROCESS` to make sure that a fetch has happened.

src/consumer/__tests__/consumeMessages.spec.js
```diff
@@ -609,6 +609,57 @@ describe('Consumer', () => {
...
+      eachBatch,
+    })
+
+    await waitForConsumerToJoinGroup(consumer)
+
+    await waitFor(() => offsetsConsumed.length === messages.length, { delay: 50 })
+    await sleep(50)
+
+    // Hope that we're now in an active fetch state? Something like FETCH_START might help
+    const seekedOffset = offsetsConsumed[Math.floor(messages.length / 2)]
```
Nevon: This is unused.

src/consumer/__tests__/consumeMessages.spec.js
Nevon: These things tend to become an issue in CI, even if they're reasonably reliable locally.

src/consumer/__tests__/consumeMessages.spec.js
JaapRood: Can you tell I duplicated the test of `consumer.seek` at all? 😅

Nevon: 🙈🙉🙊

src/consumer/__tests__/consumeMessages.spec.js
JaapRood: Ended up going with `FETCH`, which seems to capture the use case correctly. Looked into `END_BATCH_PROCESS`, which worked for the unfixed implementation, but never triggered in the first as the filtered response was an empty batch, which get skipped!

src/consumer/__tests__/consumeMessages.spec.js
JaapRood: I added `FETCH_START`, so far without a payload, as the only use case for it now doesn't require one to function. It's mirrored after `GROUP_JOIN`, with subject first, action second. Was a bit of a judgement call, as the BATCH ones do list the verb first. I reckoned subject first should help group them nicely when listing them, which seems as good a reason for something rather arbitrary as any other :)

JaapRood:
@Nevon should be good for another look!

Nevon:
Looks good to me!

PR #400, opened 2019-06-17
Log error message on connection errors (fixes #399)

### LINKED ISSUE #399: Duplicated error log and wrong formatting
I was trying to set up a custom logger and I have found that connection errors in the broker pool are logged with the error instance instead of the proper message. Also, in some cases the error was being logged twice.

I'm new to the codebase so I don't know if this is intentional or not. I'll make a PR with the fix in case it is useful.

Cheers!

rcoedo:
This PR should fix issue #399.

### DIFF
--- a/src/cluster/brokerPool.js
+++ b/src/cluster/brokerPool.js
@@ -72,8 +72,6 @@ module.exports = class BrokerPool {
         await this.seedBroker.connect()
         this.versions = this.seedBroker.versions
       } catch (e) {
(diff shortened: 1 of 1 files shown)


rcoedo:
These errors seem unrelated to my changes, am I missing something?

Thanks!

tulios:
Thanks for the PR @rcoedo, I think it's just flakiness. I will re-run the failing group.

PR #411, opened 2019-06-26
fix: use correct mechanism name for broker compatibility

### LINKED ISSUE #410: AWS SASL Mechanism Name is incorrect
@Nevon 
The [broker module](https://github.com/STACK-Fintech/kafka-auth-aws-iam) expects the mechanism name to be `AWS` and not `AWS-IAM`. The subsequent commit that was added to [my PR](https://github.com/tulios/kafkajs/pull/402) before it landed moves Kafka.js out of alignment.

The problem stems from the configuration for the [Kafka Docker Image](https://hub.docker.com/r/wurstmeister/kafka/). The script they use to parse environment variables into configuration entries does not handle hyphens correctly.

Configuration options like `listener.name.outside.aws-iam.sasl.server.callback.handler.class=com.stack.security.auth.aws.internal.AwsIamServerCallbackHandler` aren't handled correctly by the container's environment variable parsing script, which means you'll never be able to configure the broker to use it.

Admittedly, I haven't tested it using an ordinary deployment, so it's possible that this would work, but this still makes it impossible to use with the dockerized version.

kjdelisle:
The broker module expects `AWS` as the mechanism name.

Fixes #410

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -41,6 +41,7 @@ Refer to [TLS create secure context](https://nodejs.org/dist/latest-v8.x/docs/ap
 Kafka has support for using SASL to authenticate clients. The `sasl` option can be used to configure the authentication mechanism. Currently, KafkaJS supports `PLAIN`, `SCRAM-SHA-256`, `SCRAM-SHA-512`, and `AWS-IAM` mechanisms.
 
 ### PLAIN/SCRAM Example
+
 ```javascript
 new Kafka({
   clientId: 'my-app',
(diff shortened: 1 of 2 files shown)


Nevon:
Sorry about that! I wanted to avoid the AWS name because it seemed probable that AWS would eventually release some other identity product, so it seemed prudent to specify IAM.

kjdelisle:
@Nevon Believe me, I know the frustration. I initially created the plugin with `AWS-IAM` as the mechanism, and when that ran aground, I tried `AWS_IAM` and had the same problems!

PR #413, opened 2019-06-27
TypeScript Definition Topic RegExp

mkaufmaner:
Allows a consumer to subscribe to topics using RegExp.


tulios:
I am about to release 1.9.3, I can include this fix.
Thanks!

Issue #396, opened 2019-06-14
Consumer can not pause and resume individual partitions


JaapRood:
While the Consumer has `pause` and `resume` methods, these can only be used on the topic level. Since the partition is the unit of concurrency Kafka has been designed around, it would make sense to control the consumption rate for each partition individually.

## Proposed solution

Accept a `partition` or `partitions` argument as part of `consumer.pause` and `consume.resume`.

JaapRood:
It's Friday past 6pm local time, so this issue is less detailed than I would normally make them. Happy to expand on it later if that's required!

tulios:
We kept this feature out consciously; we needed resume and pause, so we implemented just what we need back them, we knew we would eventually come back and add the remaining features. We were waiting for the need, so the time has come. 😄

JaapRood:
> [...] so we implemented just what we need back them, [...]

👏👏🙌

We'll get along just fine :)

JaapRood:
Fixed by #417.

PR #417, opened 2019-07-05
Allow Consumer pausing and resuming consumption of individual partitions

JaapRood:
To address #396.

### DIFF
--- a/src/consumer/subscriptionState.js
+++ b/src/consumer/subscriptionState.js
@@ -41,8 +41,9 @@ module.exports = class SubscriptionState {
       } else if (Array.isArray(partitions) && !pausedForTopic.all) {
         partitions.forEach(partition => pausedForTopic.partitions.delete(partition))
       } else if (Array.isArray(partitions) && pausedForTopic.all) {
-        // TODO: consider whether we should actively track active topics, rather than paused ones, as to avoid this,
-        // or perhaps a "whitelist" and "blacklist" of either, to allow for pausing toppars we haven't had assigned yet
+        // TODO: Remove this after we've moved member-assignment state into here
+        // (https://github.com/tulios/kafkajs/issues/427) as that should allow us to correctly resume
+        // partitions after pausing an entire topic (by tracking resumed partitions as well as paused).
         throw new KafkaJSNonRetriableError(
           'Can not resume specific partitions of topic when entire topic was paused before'
         )
--- a/src/consumer/subscriptionState.js
+++ b/src/consumer/subscriptionState.js
@@ -64,19 +64,17 @@ module.exports = class SubscriptionState {
    * @returns {Array<TopicPartitions>} topicPartitions Example: [{ topic: 'topic-name', partitions: [1, 2] }]
    */
   paused() {
-    return Array.from(
-      Object.values(this.topicPartitions).map(({ topic, partitions }) => {
-        return {
-          topic,
-          partitions: Array.from(partitions.values()),
-        }
-      })
-    )
+    return Object.values(this.pausedPartitionsByTopic).map(({ topic, partitions }) => {
+      return {
+        topic,
+        partitions: Array.from(partitions.values()),
+      }
+    })
   }
 
   isPaused(topic, partition) {
     let paused = this.pausedPartitionsByTopic[topic]
 
-    return paused && (paused.all || paused.partitions.includes(partition))
+    return !!(paused && (paused.all || paused.partitions.includes(partition)))
   }
 }
--- a/src/consumer/subscriptionState.js
+++ b/src/consumer/subscriptionState.js
@@ -1,3 +1,5 @@
+const { KafkaJSNonRetriableError } = require('../errors')
+
 module.exports = class SubscriptionState {
   constructor() {
     this.pausedPartitionsByTopic = {}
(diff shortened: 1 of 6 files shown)


JaapRood:
I did a first draft of the required to see what the implications would be for implemention, public API, etc. In doing so I've identified some friction with the current public API of `consumer.pause` and `consumer.resume`. 

Both already accept an array `partitions` to be specified with the topic name. The issue arises is what happens when omitting this `partitions` array, or including an empty one. The rest of the changes so far suggests that if the partitions would be empty, nothing would be paused, but so far the behaviour has been the opposite: the entire topic is paused.

The question here is whether we have to break compatibility to make this change or not, or whether we can find some reasonable way to maintain the current behaviour.

I'd propose the following:

- when `typeof partitions === undefined` pause all current and future (can change through rebalance events) partitions 
- when `Array.isArray(partitions)` pause only those partitions, no partitions when empty
- otherwise throw an error

JaapRood:
The `pause.spec.js` tests are passing without any changes to the tests, so that must mean we're doing something right :)

However, the implementation also brought an issue to light with the tracking of only the paused partitions. When having paused "all" partitions through calling `consumer.pause([{ topic }])`, you then can't resume any partitions specifically. That's because the `SubscriptionState` does not have access to the partitions that have actually been assigned to the consumer group. 

This could be solved by not only keeping a collection of paused topics, but also resumed. An alternative would be for the `SubscriptionState` to own the state describing what topics and partitions are assigned (currently in `consumerGroup.topics` and `consumerGroup.memberAssignment`), which given it's name suggests that might be the right thing to do anyway. That approach would also clean up `consumerGroup.fetch` which now has to join the paused with the assigned manually (although it's the only place where that needs to happen, so perhaps that's fine).

I'll have a look at that next, as well as what other pre-existing tests are failing. From there, we can start to add tests covering the new behaviour.

tulios:
Nice, and I agree with your proposal. I can try to get an early review today.

JaapRood:
The version I just pushed fixes and tests the basic implementation of pausing and resuming single partitions. 

It does not yet cover the case where we `pause` by only topic, to then `resume` a specific partition, which is currently unsupported. I'm not sure whether it's worth the refactoring required at this point to make that work, as I don't think it's proven this is an actual use case. I reckon you either choose topics as your unit or partitions in this case.

With the meat of the changes here, it's ready for review while I get some other tests passing, like the subscription state unit tests.

JaapRood:
The PR should be in pretty good shape now, with fixed unit tests, a happy linter and documentation for the enhancement. It seems like the current failure of the tests is unrelated to the change and probably attributable to flakiness!

JaapRood:
Just fyi, #421 will impact this, as it exposes the internals of `SubscriptionState` that have changed in this version. I'd recommend making these changes first, to then resolve how to expose it.

src/consumer/subscriptionState.js
```diff
@@ -1,26 +1,73 @@
...
+    topicPartitions.forEach(({ topic, partitions }) => {
+      const pausedForTopic = this.pausedPartitionsByTopic[topic] || { topic, partitions: new Set() }
+
+      if (typeof partitions === 'undefined') {
+        pausedForTopic.partitions.clear()
+        pausedForTopic.all = false
+      } else if (Array.isArray(partitions) && !pausedForTopic.all) {
+        partitions.forEach(partition => pausedForTopic.partitions.delete(partition))
+      } else if (Array.isArray(partitions) && pausedForTopic.all) {
+        // TODO: consider whether we should actively track active topics, rather than paused ones, as to avoid this,
```
JaapRood: We'll probably want to either do something about this, or document it elsewhere. I don't think it's a proven use case yet, so throwing for now is fine.

Nevon: I feel like we should probably make a note of this in an issue or something and bundle it into some future API breaking change instead, because the current behavior is pretty confusing (but totally the right call, given backwards compatibility).

JaapRood: Do you mean the API difference between calling with or withouts `partitions` specified? This comment was about the error we have to throw when we've previously paused all topics by not specifying `partitions`, to then try to resume a single partition. My idea so far has been that people use either full topic pausing, or per partition, but if this thrown Error is too confusing, we can go with bringing the current member assignment into `SubscriptionState`, which would allow mixing the two. 

> This could be solved by not only keeping a collection of paused topics, but also resumed. An alternative would be for the SubscriptionState to own the state describing what topics and partitions are assigned (currently in consumerGroup.topics and consumerGroup.memberAssignment), which given it's name suggests that might be the right thing to do anyway. That approach would also clean up consumerGroup.fetch which now has to join the paused with the assigned manually (although it's the only place where that needs to happen, so perhaps that's fine).

Maybe that's just the way to go anyway, as it would probably clean things up a bit, too. I'll have a look at how much is involved by having a small go at it.

Nevon: Yeah, I was talking about the general problem of calling the public API with or without partitions. I'm not so sure that the assumption that people will either pause entire topics or individual partitions is valid, but I'm perfectly fine with us making the decision to go with it until we discover otherwise.

If you want to look into moving assigned partitions into SubscriptionState, you're very welcome to. I agree that it's likely the way to go, but I don't think it necessarily has to be done within the scope of this PR.

tulios:
Just a note to myself, update the typescript types before releasing this.

JaapRood:
@tulios how can I help with that in the future? Must say I haven't really used Typescript apart from some small experiments, but would like to offer complete PR's in the future!

tulios:
@JaapRood we are also getting used with the typescript flow, so don't worry. I made this mistake several times by now. 😓 

I am planning to add a PR template with a checklist of things we should be aware

PR #419, opened 2019-07-06
On ready callback function parameter added

carlosbajo:
#### OnReady(ready: boolean)

Hi, I was working a while in a project with kubernetes and other tools and discovered that I need a flag or something when the consumer group has joined to kafka because my api start to work when kafka start at all.

this feature works like this.
```javascript
  const consumer = kafka.consumer({ groupId: "testgroup", onReady: ready });
  function ready(ready) {
    console.log('Hi, the consumer has joined :)');
  }
```

I really need a feature like this on the library, if my changes don't meet the requirements I would like to open an issue.

Thank you for reading.


tulios:
Hi @carlosbajo, this is already possible with the instrumentation events, example:

```javascript
consumer.on(consumer.events.GROUP_JOIN, async e => {
  console.log({
    timestamp: e.timestamp,
    groupId: e.payload.groupId,
    memberId: e.payload.memberId,
    memberAssignment: e.payload.memberAssignment,
    // etc
  })
})
```

Take a look at https://kafka.js.org/docs/instrumentation-events#a-name-consumer-a-consumer

carlosbajo:
> Hi @carlosbajo, this is already possible with the instrumentation events, example:
> 
> ```js
> consumer.on(consumer.events.GROUP_JOIN, async e => {
>   console.log({
>     timestamp: e.timestamp,
>     groupId: e.payload.groupId,
>     memberId: e.payload.memberId,
>     memberAssignment: e.payload.memberAssignment,
>     // etc
>   })
> })
> ```
> 
> Take a look at https://kafka.js.org/docs/instrumentation-events#a-name-consumer-a-consumer

una disculpa por no revisar bien la doc, :p. thx tulios

PR #425, opened 2019-07-12
feat: allow sasl option to be provider function

### LINKED ISSUE #423: AWS Credential Refresh Issue
Hello again!

Normally for something like this, I would've written a pull request first, but I wanted to coordinate on this to make sure I'm not hacking through existing patterns and whatnot.

The problem I'm currently seeing with the new AWS authentication is that it won't work for session-based credentials (which expire and require refresh). Originally, I thought that the consuming code would need to be responsible for handling these refreshes, but after some investigative work I've realized that Kafka.js is caching these credentials at various layers.

Even after writing a custom getter and injecting it into the options, I've noticed that it never gets called more than once, which means its output is being cloned on the first call.

Is there an established methodology for this that I'm not seeing, or would there need to be some refactoring involved to hoist those credentials out into some sort of provider that can be updated on-the-fly?

As of right now, the only alternative I can see is disposing of the Client on each call -- there aren't any errors that my consuming code can capture directly (please let me know if this is wrong), which means that I can't trigger a new Client creation on authentication failure.

Given the expense of creating all of the broker connections again, that's not something I think we'd want to do.

kjdelisle:
fixes #423 

You may now use a function for the sasl property of Client
configuration, as long as it returns an object of the same format.

Note: I think there's probably a type definition in here somewhere that will need updating to allow this.

### DIFF
--- a/src/broker/saslAuthenticator/index.js
+++ b/src/broker/saslAuthenticator/index.js
@@ -28,7 +28,12 @@ module.exports = class SASLAuthenticator {
   }
 
   async authenticate() {
-    const mechanism = this.connection.sasl.mechanism.toUpperCase()
+    let mechanism
+    if (typeof this.connection.sasl === 'function') {
+      mechanism = this.connection.sasl().mechanism.toUpperCase()
+    } else {
+      mechanism = this.connection.sasl.mechanism.toUpperCase()
+    }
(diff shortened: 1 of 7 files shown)


kjdelisle:
I'll open a new one later; I need to keep experimenting and I don't want to keep spamming CI jobs at the moment.

Issue #427, opened 2019-07-15
Move member assignment state into SubscriptionState


JaapRood:
While working on #417 we ran into limitations due to `SubscriptionState` not owning the current member assignment state, which instead is held by `ConsumerGroup`. It seems to make more sense for the subscription state to hold all state about the current subscription, including the current assignment.

In #417 it caused an issue where, after pausing a full topic through `consumer.pause`, individual partitions could then not be resumed. That's because it couldn't resolve the 'relative' definition of "all" partitions of a topic with the current assignment, to a concrete list of partitions to fetch or not fetch from. In #421 we'll run into that same resolve issue again.

JaapRood:
Looking through open issues, it dawned on me that these changes will also help #373, where we'll look for manual assignments rather than subscriptions. Another reason for me to get started on this.

PR #421, opened 2019-07-08
feat: add consumer.paused() to get the list of paused topics

terebentina:
Closes #415

PR #429, opened 2019-07-15
Use SubscriptionState to track member assignment and how that's affected by pause / resume logic

JaapRood:
To address #427.

### DIFF
--- a/src/consumer/consumerGroup.js
+++ b/src/consumer/consumerGroup.js
@@ -166,8 +166,7 @@ module.exports = class ConsumerGroup {
       memberAssignment: decodedAssignment,
     })
 
-    let currentMemberAssignment = decodedAssignment
-    const assignedTopics = keys(currentMemberAssignment)
+    const assignedTopics = keys(decodedAssignment)
     const topicsNotSubscribed = arrayDiff(assignedTopics, topicsSubscribed)
 
     if (topicsNotSubscribed.length > 0) {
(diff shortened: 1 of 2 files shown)


JaapRood:
Starting from what I imagine the enhanced API of `SubscriptionState` to look like, it seems definitely like a net gain to me in terms of clarity. While logic living between two files is always a bit of a trade off, I think it now becomes a lot more explicit what other logic governs the member assignment and how it should be used.

From trying to fit the consistent API of `SubscriptionState`, accepting `Array<TopicPartitions>`, it became apparent that the code that calls it should keep track of it's newly received assignment in a slightly different format. All consumer tests passing should prove that this should be fine.

So far, while the `SubscriptionState` assignment is being set as part of `consumerGroup.sync`, it's not used anywhere yet. I'll look at that next, which should give good insight whether these drafted changes actually have the desired effect.

JaapRood:
I'm happy to consider this ready for merging with the progress made.

`subscriptionState.active` is now used to determine which partitions to fetch for, taking into consideration all pausing and resuming logic. The `paused` method now renders which of the _assigned partitions_ are paused, whether an entire topic or just the individual partitions were paused, making it ready for public consumption (see #421).

Now that `subscriptionState` is aware of the assigned partitions, we could also add the tracking of explicitly resumed partitions, removing an entire type of `Error` that can be thrown. Basically you can now do this:

```js
consumer.pause([{ topic: 'topic1' }]) // this pauses all current and future partitions for this topic

// This resumes consumption of these specific 2 partitions, while the rest remains paused. 
// Threw error before
consumer.resume([{ topic: 'topic1', partitions: [1,2]}]) 
```

I had a look whether there were other things of `ConsumerGroup` that would be better housed in `SubscriptionState`, but everything else is too closely related to the internals of the consumer group to qualify.

JaapRood:
Looks like tests are failing because of an actual problem for once 😂. On it!

JaapRood:
Turned out that the `runner` was accessing `memberAssignment` as well. Instead of exposing that once again, I added `consumerGroup.assigned`, which proxies `subscriptionState.assigned`, just like it does with paused. A bit of format conversion was needed, to not break the public API of the `GROUP_JOIN` event.

src/consumer/consumerGroup.js
```diff
@@ -220,13 +217,19 @@ module.exports = class ConsumerGroup {
...
       topicConfigurations: this.topicConfigurations,
       instrumentationEmitter: this.instrumentationEmitter,
-      memberAssignment: this.memberAssignment,
+      memberAssignment: currentMemberAssignment.reduce(
+        (partitionsByTopic, { topic, partitions }) => ({
+          ...partitionsByTopic,
+          [topic]: partitions,
+        }),
+        {}
+      ),
```
JaapRood: Hmm, turns out that not only `SubscriptionState` needs to own member assignments, as the `OffsetManager` requires the same. It seems such a natural fit, as it really is subscription state, but perhaps it belongs to the `ConsumerGroup` after all? 🤔. Or perhaps instead, we should pass the result of `this.subscriptionState.assigned()` as a parameter to `OffsetManager` and have it accept the `Array<TopicPartitions>` type, dealing with it's own internal representation separately?

JaapRood: Had another look at this, leaving me to think it's right the way it is, for now.

Naturally thinking, which partitions are assigned really is part of the domain of subscription state, where as for an offset manager it seems more like something to use. As far as I'm concerned, the ownership model is right this way.

Adapting `OffsetManager` to accept `Array<TopicPartitions>` is possible, but for the moving of code from `ConsumerGroup` to `OffsetManager`, we'd have to at least change a whole bunch of tests too, as the format of the `memberAssignment` parameter changes. If we'd actually want to eliminate this bit of code all together, we'd have to refactor all of OffsetManager in addition to the tests.

tulios:
@JaapRood can you update the typescript types with the new methods and any signature changes? I can do it if you can't.

JaapRood:
@tulios if you're in the position to do so swiftly, go right ahead. I haven't touched TS in a while and not past some experiments, so would have to take some more time than I have available for it today! Happy to include it in future PR's, though :)

tulios:
I have updated the types; I will merge after a green build.

Issue #395, opened 2019-06-14
Consumer.seek commits an offset, causing possible data and metadata loss


JaapRood:
In debugging one of our stream processing tasks we noticed that `consumer.seek` commits an offset in order to affect the consumer's fetching position. This mixes controlling the consumer position with recording progress and can lead to offset-metadata getting lost.

Offset metadata can be very useful when working with replicating to and from local stores, especially for achieving exactly-once processing, and [is used by KafkaStreams](https://docs.google.com/document/d/1pGZ8xtOOyGwDYgH5vA6h19zOMMaduFK1DAB8_gBYA2c) for exactly this purpose (emphasis mine):

> - When committing a task, after the state stores have all been flushed and before produce write the committed offset, we can choose to trigger the persistent state stores' checkpoint mechanism to make a checkpoint for each persistent stores.
> - When writing the committed offset into the input-offset-topic, also **pass in the reference (e.g. local state directory / files locations) of all the persistent store checkpoints into the offset metadata as part of the state triplet**.
> - Then upon resuming, if the fetched last transaction contains the state store checkpoints information, we can skip the restoration step of the state stores from changelog but use the checkpointed image directly.

When seek commits an offset, this metadata is subsequently lost.

Another example is where you produce to the same topic that you're consuming (eating your own tail), which is useful for things like implementing uniqueness constraints ([described well here](https://queue.acm.org/detail.cfm?id=3321612)). There it can often be useful to seek back 1 from the last committed offset, to kickstart the process (as if you don't, no messages are ever produced for you to continue from), especially when doing things like replicating from a store to a Kafka topic.

As far as I can tell, the Java `KafkaConsumer` does not commit the offset, but only changes the offset in memory so the next fetch happens from there.

## Proposed solution

Eventually, I think we'd want to be in a situation where:

- `consumer.seek` only affects what message is fetched next, not how progress has been made thus far.
- `consumer.commit` can be used to record progress, possible after a seek, if that's required. See #378.

However, this would constitute a breaking change, despite the committing of the offset is currently a bit of a side effect. So perhaps a flag passed to `consumer.seek` would be a way to opt-in to the new behaviour, as a way to verify the approach in real applications. That way if the API is to be broken in a new major version, we prevent the need for a revert.

JaapRood:
[...]

Issue #373, opened 2019-05-29
Add manual assignment of TopicPartitions to Consumer


JaapRood:
In using Kafka for stream processing it's common practice to have to manually assign topic-partitions for consumption, rather than subscribing and receiving an assignment through the group.

For example, when processing messages from a partition the consumer subscribes to, you might have to keep some state. To make sure this state is durable, it's often replicated to a changelog, also implemented by a Kafka topic. That way, when the consumer is designed another partition or crashes, the state can be restored. To make sure the right processing state is restored before processing continues, the changelog is written to a topic with the same partition for which we're processing, a process called **copartitioning**. By simply consuming the state-replication-topic from the same partition for which you're processing messages, you're guaranteed to restore the right state. 

In the above example, subscribing to a ConsumerGroup won't work: you've already been assigned a partition for the input topic and you need exactly that same partition number. There's other examples, too, like replicating a changelog as a local Cache on every node, ready to be queried through HTTP. In that case, you want to consume _all_ partitions, rather than just be assigned a couple.

The documentation of the official Java client [describes it](https://kafka.apache.org/10/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#manualassignment) as well:

> In the previous examples, we subscribed to the topics we were interested in and let Kafka dynamically assign a fair share of the partitions for those topics based on the active consumers in the group. However, in some cases you may need finer control over the specific partitions that are assigned. For example:
> - If the process is maintaining some kind of local state associated with that partition (like a local on-disk key-value store), then it should only get records for the partition it is maintaining on disk.
> - If the process itself is highly available and will be restarted if it fails (perhaps using a cluster management framework like YARN, Mesos, or AWS facilities, or as part of a stream processing framework). In this case there is no need for Kafka to detect the failure and reassign the partition since the consuming process will be restarted on another machine.

## Proposed solution

[Like the Java `KafkaConsumer` does](https://kafka.apache.org/10/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#assign-java.util.Collection-), allow to call `consumer.assign` instead of `consumer.subscribe` with an array of topic partitions.

To make this practical, implementing `consumer.assignments()` might be necessary too, returning the list of topic partitions assigned to the consumer (through either subscription or manual assignment).

JaapRood:
[...]

PR #436, opened 2019-07-16
Add way to commit offsets with Consumer

### LINKED ISSUE #378: Add way to commit offset with Consumer

JaapRood:
Closes #378. 

> In various other clients, like the official KafkaConsumer, arbitrary offsets can be committed through an interface like consumer.commit. So far, the only way to do this in KafkaJS is through commitOffsetIfNecessary function exposed as part of the eachBatch mechanism.
> 
> This limitation prevents the committing of offsets before a message has been consumed, which is a time and place where in stream processing tasks it's desirable to do so. While in most cases a consumer.seek operation is desirable, there are situations where you just want to set the progress made and not consume any messages at all.

### DIFF
--- a/docs/Consuming.md
+++ b/docs/Consuming.md
@@ -150,11 +150,44 @@ consumer.run({
 
 Having both flavors at the same time is also possible, the consumer will commit the offsets if any of the use cases (interval or number of messages) happens.
 
-`autoCommit`: Advanced option to disable auto committing altogether. If auto committing is disabled you must manually commit message offsets, either by using the `commitOffsetsIfNecessary` method available in the `eachBatch` callback, or by [sending message offsets in a transaction](Transactions.md#offsets). The `commitOffsetsIfNecessary` method will still respect the other autoCommit options if set. Default: `true`
+`autoCommit`: Advanced option to disable auto committing altogether. Instead, you can [manually commit offsets](#manual-commits). Default: `true`
+
+## <a name="manual-commits"></a> Manual committing
+
+When disabling [`autoCommit`](#auto-commit) you can still manually commit message offsets, in a couple of different ways:
+
+- By using the `commitOffsetsIfNecessary` method available in the `eachBatch` callback. The `commitOffsetsIfNecessary` method will still respect the other autoCommit options if set.
+- By [sending message offsets in a transaction](Transactions.md#offsets). 
+- By using [the `commitOffsets` method](#commit-offsets) of the consumer.
+
+The `consumer.commitOffsets` is the lowest-level option and will ignore all other auto commit settings, but in doing so allows the committed offset to be set to any offset and committing various offsets at once. This can be useful, for example, for building an processing reset tool. It can only be called after `consumer.run`. Committing offsets does not change what message we'll consume next once we've started consuming, but instead is only used to determine **from which place to start**. To immediately change from what offset you're consuming messages, you'll want to [seek](#seek), instead.
+
+```javascript
+consumer.run({
+    autoCommit: false,
+    eachMessage: async ({ topic, partition, message }) => {
+        // Process the message somehow
+    },
+})
+
+consumer.commitOffsets([
+  { topic: 'topic-A', partition: 0, offset: '1' },
+  { topic: 'topic-A', partition: 1, offset: '3' },
+  { topic: 'topic-B', partition: 0, offset: '2' }
+])
+```
+
+Note that you don't *have* to store consumed offsets in Kafka, but instead store it in a storage mechanism of your own choosing. That's an especially useful approach when the results of consuming a message is written to a datastore that allows atomically writing the consumed offset with it, like for example PostgreSQL. When possible it can make the consumption fully atomic and give "exactly once" semantics that are stronger than the default "at-least once" semantics you get with Kafka's offset commit functionality. 
+
+The usual usage pattern for offsets stored outside of Kafka is as follows:
+
+- Run the consumer with `autoCommit` disabled.
+- Store a message's `offset + 1` in the store together with the results of processing. `1` is added to prevent that same message from being consumed again.
+- Use the externally stored offset on restart to [seek](#seek) the consumer to it.
 
 ## <a name="from-beginning"></a> fromBeginning
 
-The consumer group will use the latest committed offset when fetching messages. If the offset is invalid or not defined, `fromBeginning` defines the behavior of the consumer group. This can be configured when subscribing to a topic:
+The consumer group will use the latest committed offset when starting to fetch messages. If the offset is invalid or not defined, `fromBeginning` defines the behavior of the consumer group. This can be configured when subscribing to a topic:
 
 ```javascript
 await consumer.subscribe({ topic: 'test-topic', fromBeginning: true })
--- a/docs/Consuming.md
+++ b/docs/Consuming.md
@@ -158,7 +158,7 @@ When disabling [`autoCommit`](#auto-commit) you can still manually commit messag
 
 - By using the `commitOffsetsIfNecessary` method available in the `eachBatch` callback. The `commitOffsetsIfNecessary` method will still respect the other autoCommit options if set.
 - By [sending message offsets in a transaction](Transactions.md#offsets). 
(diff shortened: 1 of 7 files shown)


JaapRood:
With the first steps made to explore how this would look, there's a couple of things to mention:

- Rather than going with the shape of `consumerGroup.commitOffsets`, I defined the type as `Array<TopicPartitionOffsetAndMetadata>`, not unlike the KafkaConsumer does. I changed it to this, as the signature of `{ topics: {[ topic, partitions: [{ partition, offset, metadata }] ]}` was just really cumbersome, instantly showed by just writing the basic specs. Instead, we accept `[{ topic, partition, offset, metadata }]`.
- It was quite awkward trying to figure out how to assert this interface is working. The `commitOffsets` is tested more deeply in specs for the offset manager, so I wasn't concerned with being overly exhaustive. However, I found myself having to setup a consumption loop, while the point of this PR is showing how that shouldn't really be necessary. Sometimes you just want to set the point of progress, in like application reset, or using some custom batching logic, etc. Requiring a consumption loop basically makes this similar to `consumer.seek` (especially as seek currently commits offset #395). It kind of ties into #373 as well, which might be a better time to address this. So far, it's just used instead of auto commit (even though there's other ways of doing that).
- We don't actually assert in the test yet that the committing works properly, it's just a draft to explore so far 😬.

JaapRood:
Looking at the code, the other thing I'm worried about given the current implementation is that `commitOffset` call would cause basically a `seek` to be performed, where I'm pretty convinced it shouldn't. Committing an offset is basically just producing a message (they have their own request in the protocol, but they are stored on a kafka partition, hence why it works transactionally as well). It should only be used to consider from where to resume processing when starting an assignment, like after a rebalance or manual assignment. In other words, a stored `committedOffset` _should_ basically be the seed of the local `consumedOffset` when determining where to fetch and consume the partition from. As highlighted by #395, seeking should be changing that `consumedOffset`, but not touch the `committedOffset`.

I'll have to dig deeper to see if that's the effect it's having, see if I can write a test to verify it either way. You could say that could be a separate issue and PR, but as this concerns a public API, changing it's behaviour would incur a breaking change, so it's probably something we want to figure out right from the beginning.

JaapRood:
Having looked closer at how `OffsetManager`, `ConsumerGroup` and `Cluster` interact, I'm happy to say that I don't believe calling `commitOffset` performs a seek, with the consumption continuing where it should 😬. While the opposite _is_ true, keeping #395 as a valid issue, it shouldn't impact these changes. I'll add a test to verify as much, just to be sure we don't conflate committing and seeking behaviours now or in the future!

JaapRood:
Some work remains to be done on the documentation side, but apart from that, I think this is looking pretty good and ready for review.

tulios:
This PR looks good, but I think it is still missing something. What happens if you try to commit outside of the run loop and you get a protocol error back, for example, `REBALANCE_IN_PROGRESS`. In the current setup, the consumer will catch the error and execute the rebalance, but since the error occurred outside of the loop, you have to "notify" the consumer that a rebalance is needed. This is also the case for other errors, so maybe some errors have to trigger the "recover" phase of the consumer.
WDYT?

JaapRood:
You make a good point, something I hadn't considered thus far. Thinking out loud for a bit:

Technically speaking, you'd say that both calls to commit offsets (or send heartbeats) happen in the context of the `ConsumerGroup`, which is the entity that should be able to reason about these kinds of semantics (atleast from a conceptual point of view). Given that, why would there be any difference about who called this method and how to recover from errors?

Looking at where these protocol errors are handled now, it seems to be at the Runner level. I guess that has the benefit that any protocol errors inside a `eachBatch` or `eachMessage` is also caught. However, if we only want to respond to protocol errors that originate from our `ConsumerGroup` in the first place, shouldn't those already have been dealt with before they even reach that message processing context?

@tulios not having really look at the reality of it in code, my recommendation would be to handle protocol errors that concern the consumer group inside `ConsumerGroup`, regardless of where the calls for `heartbeat` or `commitOffsets` come from. Would be great to hear what you think about that rationale.

JaapRood:
Looking at the actual code, things seem a bit more complicated, especially if we want to avoid an all-out refactor. The `Runner` dealing with run-time state, errors and retries seems like a good idea, so I'm going in the direction of trying to run the commit offsets in context of it.

JaapRood:
I took the approach of adding a `commitOffsets` to `Runner` and calling that from the public consumer API. This allows runner to handle the errors around rebalancing or lost members (yet to be verified by test coverage) and the operation to be retried. It still needs some love, but this is mostly to discuss whether this it the right approach.

From the perspective of code organisation, I can imagine the same working for `consumer.heartbeat`, although the retry logic would be somewhat different (there's no longer a need to send a heartbeat after a group has just rebalanced or we lost our membership).

The thing I'm wondering if retrying after a rebalance is actually the behaviour we want to see from committing offsets. A commit would be sensitive to the assignment context it's made in, so if that context changes, isn't the correct thing to bail on the attempt instead? A new assignment would then pick up the previously committed offset and "retry" through processing the messages again. To achieve exactly-once, transactions should be used.

JaapRood:
@tulios I think that covers it, ready for another look from you!

tulios:
@JaapRood sorry for the delay, the changes should work. Final question 😄 
Let's say I have my consumer running and I'm not committing with the new method within the run loop, what should be the expectation when using the new `commitOffset`, if I commit offset 3 should the consumer fetch from offset 4, or is it acceptable that the two things might not be in-sync, and the consumer will actually fetch from offset 2 because fetch was triggered before `commitOffset` actually finished.

JaapRood:
> @JaapRood sorry for the delay

@tulios no worries, that's what async working is for 👍 .

Fetching and committing do not have to be in sync, as far as I've been able to gather through experience, my exploration above in this PR and #395. They just happen to be most of the time. It should be through the use of `consumer.seek` that a user can keep them in sync, by performing that after the commit. `Consumer.seek` in turn, should not commit an offset, merely set the run-time 'consumed' offset to produce the right 'next' offset to fetch from (see #395). The 'committed' offset only interacts with 'consumed' offset in determining from where to start fetching a new assignment when not explicitly seeked too.

Hopefully that holds water, happy to further discuss those semantics if need be!

tulios:
Nice, that's also my understanding. I just wanted to make sure we were aligned.

ThisIsMissEm:
What's the timeline for this reaching a release?

(I'd been looking at the documentation & thought it was already released)

tulios:
Hi @ThisIsMissEm, we usually run all the changes in production before we release to the public. We merged the feature recently so it might take at least a week for us to release, in the meanwhile, you can depend on the commit hash (`npm install https://github.com/tulios/kafkajs.git#2d5d83097ed347d466f7a97969dbfe1f03189db4`)

The current setup is suboptimal, and we are already addressing the situation, once #441 is merged every merge to master will create a pre-release version on GitHub with the change.

ThisIsMissEm:
@tulios okay, thanks!

PR #440, opened 2019-07-23
Update dev dependencies

tulios:
This PR should address most of the security warnings we've received.


src/network/connection.js
```diff
@@ -207,6 +207,12 @@ module.exports = class Connection {
    */
   authenticate({ authExpectResponse = false, request, response }) {
     this.authExpectResponse = authExpectResponse
+
+    /**
+     * TODO: rewrite removing the async promise executor
+     */
+
+    /* eslint-disable no-async-promise-executor */
```
Nevon: Catch on line 239 and reject

tulios: 21f549e

tulios:
`1.9.3` released :tada:
https://www.npmjs.com/package/kafkajs/v/1.9.3

PR #441, opened 2019-07-23
Add pre-release on merge to master

tulios:
This PR adds a new step to the pipeline to publish a pre-release of master.

The versioning follows this convention:

If the master branch is on `1.9.3` the next time we merge it will release `1.10.0-beta.0`, the next merge `1.10.0-beta.1` and so on. Once we bump master to `1.10.0`, it will start releasing `1.11.0-beta.0`. If we just release a patch, it will continue to make beta releases using the next minor, so master on `1.9.4` should release beta `1.10.0-beta.2`, assuming the previous example.

I am never sure if the YML changes will work on Azure pipelines, so if you like the idea I will try this manually before I merge

@JaapRood it's here 😄


JaapRood:
This is great, makes it that much easier to use a pre-release version, with the version number nicely documenting in our projects what it could be upgraded to. I do wonder how you ever publish patch versions for a current minor? Say `1.10.0` just got released, is broken and requires a `1.10.1`?  Wouldn't some hotfix PR then instantly create the wrong pre-release version?

azure-pipelines.yml
```diff
@@ -192,6 +192,32 @@ jobs:
...
+      vmImage: 'Ubuntu 16.04'
+    steps:
+      - task: NodeTool@0
+        inputs:
+          versionSpec: '8.x'
+      - bash: ./scripts/betaRelease.js
+        displayName: build beta release
+      - task: Npm@1
+        inputs:
+          command: custom
```
Nevon: Maybe we should tag with the new beta version so that you can find which commits are in the beta release. Would help to generate changelogs etc if we really want to as well.

tulios:
The hotfix would release `1.10.1` and `1.11.0-beta.0`, I can make the pre-release dependent on the release, so we never create both if that's the case. I am also planning to add the commit hash into the bundle so you can always check what's in the release, or create a tag on github. I will try to refine this a bit.

tulios:
`1.10.0-beta.0` released :tada:

scripts/commentBackOnPR.js
```diff
@@ -0,0 +1,115 @@
...
+const commentOnPR = async () => {
+  const commitMessage = execa
+    .commandSync('git log -1 --pretty=%B', { shell: true })
+    .stdout.toString('utf-8')
+    .trim()
+
+  const PR_NUMBER_REGEXP = /^Merge pull request #(?<prNumber>[^\s]+)/
+  const { groups } = commitMessage.match(PR_NUMBER_REGEXP)
+
+  if (!groups.prNumber) {
```
Nevon: The return value from `match` can be undefined, which would cause this to crash

tulios:
This PR is complete now; it will only generate beta versions for merges changing `index.js`, `package.json`, `src/*` or `types/*`. It will also create a tag with the beta version.

It's hard to test pipeline changes, so we have to merge and fix as we go.

PR #444, opened 2019-07-24
Expose consumer paused partitions

tulios:
I made a small change to the subscription state, @JaapRood can you take a look? I am filtering out paused topics without active partitions on the consumer. The method will always return an empty array if the consumer group isn't initialized, I think it makes more sense than throwing the error on this case.

CC @terebentina


tulios:
I am excited to see if the pre-release is going to work. 😄

JaapRood:
[No such luck? ](https://github.com/tulios/kafkajs/commit/4710f5c8b40dade22cf645e638e43d02de364ff7) 😂

tulios:
yml pipelines are hard :/

PR #445, opened 2019-07-24
feat: allow sasl option to be provider function

kjdelisle:
Request handlers now check if SASL is a provider function, and
call it on every use of the encode functions.


kjdelisle:
As it turns out, you can't `Object.assign` the credentials created by the [AWS SDK](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Credentials.html) as they're doing clever things to avoid enumeration of the `secretAccessKey`, which is why this PR wasn't working for me.

Now I just need to get this PR to pass tests... :)

kjdelisle:
@tulios PTAL

docs/Configuration.md
```diff
@@ -95,6 +95,24 @@ A complete breakdown can be found in the IAM User Guide's
...
+```javascript
+function getCredentials() {
+  return credentials; // Assume this value is being updated by some other function.
+}
+
+new Kafka({
+  clientId: 'my-app',
+  brokers: ['kafka1:9092', 'kafka2:9092'],
+  // authenticationTimeout: 1000,
+  sasl: getLatestCredentials,
```
Nevon: This should be `getCredentials`.

src/broker/saslAuthenticator/index.js
```diff
@@ -28,7 +28,12 @@ module.exports = class SASLAuthenticator {
   }
 
   async authenticate() {
-    const mechanism = this.connection.sasl.mechanism.toUpperCase()
+    let mechanism
+    if (typeof this.connection.sasl === 'function') {
+      mechanism = this.connection.sasl().mechanism.toUpperCase()
```
Nevon: Let's not duplicate the uppercasing. Instead, let's evaluate the function to get the sasl configuration if needed and then operate as normal:

```javascript
const saslConfiguration = typeof this.connection.sasl === 'function' ? this.connection.sasl() : this.connection.sasl
const mechanism = saslConfiguration.mechanism.toUpperCase()
```

Nevon:
I've been through the code, and now I'm seeing that this "check if it's a function and evaluate it" is creeping into a lot of places. I think this will be hell to deal with.

I would instead suggest that we create a getter on the `Connection` class that evaluates `sasl` if needed and always returns an object. Something like:

```javascript
class Connection {
 get sasl() {
    return typeof this.sasl === 'function' ? this.sasl() : this.sasl
  }
}
```

That way there would be almost no changes to any other code.

kjdelisle:
@Nevon

I initially attempted to make the change exclusively on the connection object, but the protocol request functions are building the encode functions with the parameters stored outside of their scope.

>I've been through the code, and now I'm seeing that this "check if it's a function and evaluate it" is creeping into a lot of places. I think this will be hell to deal with.

I wrestled with the idea of using static AWS credentials and rotating them manually, but I think support for this idea goes a long way in helping encourage good practices like the use of short-lived, automatically rotated/provided credentials.

I'm going to tinker with using the getter on the Connection class, but I think the changes to those protocol request handlers will still be necessary.

kjdelisle:
@tulios I've updated the code based on @Nevon 's feedback, and also modified the test slightly to assert that the provider function is in fact being called multiple times (by tainting the test provider function with a call counter).

If you're okay with using [Sinon.js](https://sinonjs.org/), I can rewrite that test to use a spy instead. I decided to err on the side of caution and not include a new dev dependency just to make that simpler.

tulios:
@kjdelisle I am a bit delayed on this PR, I will try to review it today.

kjdelisle:
@tulios We're looking to launch our stuff in production soon, and I need this code to avoid using static credentials. If this isn't going to land, I'll have to start bundling my fork instead. :sob:

tulios:
Sorry about that, I dropped the ball here. I will get it done.

Nevon:
I opened #492 to add to this, as we need a way to know _when_ to re-authenticate. I'm assuming that in your solution you are simply periodically re-authenticating or something, but the proper solution would be for the broker to tell you when you need to re-authenticate (which is already supported by Kafka. See my issue)

We can still land this change separately, but looking ahead, we'll need to adapt it slightly to make sure that you could re-authenticate from within the sasl provider function. In practice, it means that rather than the sasl function being a plain synchronous function, it needs to return a promise that resolves with the credentials.

docs/Configuration.md
```diff
@@ -95,6 +95,24 @@ A complete breakdown can be found in the IAM User Guide's
 It is **highly recommended** that you use SSL for encryption when using `PLAIN` or `AWS`,
 otherwise credentials will be transmitted in cleartext!
 
+### SASL Provider Functions
+
+Alternatively, you may set the `sasl` property to a function that returns a compatible configuration
+object, which allows dynamic refresh of credentials at runtime.
+
+```javascript
+function getCredentials() {
```
Nevon: This would have to become something like:

```
async function getCredentials() {
  const credentials = await getToken()
  return credentials
}
```

src/broker/saslAuthenticator/awsIam.js
```diff
@@ -10,18 +10,16 @@ module.exports = class AWSIAMAuthenticator {
 
   async authenticate() {
     const { sasl } = this.connection
-    if (!sasl.authorizationIdentity) {
+    const props = sasl
```
Nevon: Is there a reason that you are re-assigning `sasl` to this `props` variable? I can't see any reason to do it, so please remove it and just use `sasl` directly.

src/broker/saslAuthenticator/plain.js
```diff
@@ -10,7 +10,8 @@ module.exports = class PlainAuthenticator {
 
   async authenticate() {
     const { sasl } = this.connection
-    if (sasl.username == null || sasl.password == null) {
+    const props = sasl
```
Nevon: Same

src/broker/saslAuthenticator/scram.js
```diff
@@ -128,9 +128,10 @@ class SCRAM {
   async authenticate() {
     const { PREFIX } = this
     const { host, port, sasl } = this.connection
+    const props = sasl
```
Nevon: Same

src/network/connection.js
```diff
@@ -90,6 +90,14 @@ module.exports = class Connection {
     this.shouldLogBuffers = getEnv().KAFKAJS_DEBUG_PROTOCOL_BUFFERS === '1'
   }
 
+  get sasl() {
```
Nevon: Since `_sasl` can now be an async function, this will also need to become async, I'm afraid.

This unfortunately means that the getter approach doesn't make much sense anymore (sorry about that. Didn't think ahead). Instead, we should just make sure that `sasl` is always an async function, so something like:

```javascript
this.sasl = async () => typeof sasl === 'function' ? sasl() : sasl
```

This means that any place where we're accessing `connection.sasl` will need to change to invoke and await:

```javascript
const sasl = await connection.sasl()
```

Another option is to keep the getter, and change it to return a promise:

```javascript
get sasl() {
  return new Promise((resolve, reject) => {
    if (typeof this._sasl === 'function') {
      try {
        resolve(this._sasl())
      } catch (e) {
        reject(e)
      }
    } else {
      resolve(this._sasl)
    }
  })
}
```

Either way, all call sites would need to be updated to await the promise, so:

```javascript
const sasl = await connection.sasl
```

In my mind that feels less intuitive, so I wouldn't go with the getter.

src/protocol/sasl/awsIam/request.js
```diff
@@ -2,10 +2,16 @@ const Encoder = require('../../encoder')
 
 const US_ASCII_NULL_CHAR = '\u0000'
 
-module.exports = ({ authorizationIdentity, accessKeyId, secretAccessKey, sessionToken = '' }) => ({
+module.exports = sasl => ({
   encode: async () => {
+    const props = sasl
```
Nevon: More `props`.

src/protocol/sasl/plain/request.js
```diff
@@ -19,10 +19,11 @@ const Encoder = require('../../encoder')
 
 const US_ASCII_NULL_CHAR = '\u0000'
 
-module.exports = ({ authorizationIdentity = null, username, password }) => ({
+module.exports = sasl => ({
   encode: async () => {
+    const props = sasl
```
Nevon: Same

testHelpers/index.js
```diff
@@ -59,6 +59,19 @@ const saslConnectionOpts = () =>
     },
   })
 
+let callCount = 0
```
Nevon: This callcount is now shared with every test that imports this module. It should at least be scoped, i.e:

```javascript
const saslConnectionDynamicOpts = () => {
  let callCount = 0

  const baseSaslConnectionOpts = saslConnectionOpts()
  const sasl = async () => Object.assign(
    {},
    baseSaslConnectionOpts.sasl,
    callCount: ++callCount
  )

  return Object.assign({}, baseSaslConnectionOpts, { sasl })
}
```

Nevon:
I'm working on adapting this to the new requirements. It's a little bit tricker than I had hoped, and I would like to get ##496 in first, as they touch a bit of the same code.

What I realized when I started working on this is that we need a separation between getting the current sasl configuration and renewing the sasl configuration. My plan is to introduce a `SaslProvider` inside the `Connection` (or maybe inside `Broker` and having it pass it down to the `Connection`).

The `SaslProvider` will expose two functions: `sasl(): {SaslConfiguration}` and `renew(): {Promise}`. `renew` is basically just wrapping the `sasl` option provided by the user in an async function. When invoked, it will store the returned value. `SaslProvider.sasl` is just a getter for the current `SaslConfiguration`.

Almost all places will just call `SaslProvider.sasl`, same as today, but during the initial authentication and subsequent reauthentications we can call `renew`.

The reason for this is that there are quite a few places that actually make use of the sasl configuration, but we don't want to renew it in all those places. If the user provides a static `SaslConfiguration`, then `renew` will basically do nothing (just return the same value as before).

kjdelisle:
There was a bit of a change on my end and we're going to be rolling out some other infrastructure before this gets launched so it might make more sense to close this PR and roll out something more robust as opposed to my hacky PR. :P

If anything, I'll need to spend some time on the Java SASL module to ensure that it can respond appropriately with session expiration and such.

kjdelisle:
Opened https://github.com/STACK-Fintech/kafka-auth-aws-iam/issues/5 for the Kafka plugin work on my end.

PR #458, opened 2019-08-08
fix(client): throw an error if broker list is empty

miton18:
When no brokers are provided, return an understandable error


Nevon:
Since #460 has the same fix, but includes tests, I'm going to merge that one (once it's been updated to throw a non-retriable error). But regardless, I very much appreciate the contribution! ❤️

Issue #454, opened 2019-08-06
Better error message when broker list is empty


miton18:
When, by a misconfiguration, you try to instantiate a `new Kafka({ brokers: [] })`

The error is not very easy to understand:
```
TypeError: Cannot read property 'split' of undefined
```

caused by https://github.com/tulios/kafkajs/blob/master/src/cluster/connectionBuilder.js#L24

we can easily check `brokers.length` before doing anything

miton18:
Proposal #458

Nevon:
Fixed in #460.

PR #460, opened 2019-08-08
Issue #454 proposal - Adding validation to connectionBuilder

capucho:
Proposal for issue #454

### DIFF
--- a/src/cluster/connectionBuilder.js
+++ b/src/cluster/connectionBuilder.js
@@ -1,4 +1,11 @@
 const Connection = require('../network/connection')
+const { KafkaJSConnectionError } = require('../errors')
+
+const validateBrokers = brokers => {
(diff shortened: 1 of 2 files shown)


src/cluster/connectionBuilder.js
```diff
@@ -1,4 +1,11 @@
 const Connection = require('../network/connection')
+const { KafkaJSConnectionError } = require('../errors')
+
+const validateBrokers = brokers => {
+  if (!brokers || brokers.length === 0) {
+    throw new KafkaJSConnectionError(`Failed to connect: expected brokers array and got nothing`)
```
Nevon: `KafkaJSConnectionError` is a retriable error (it inherits `KafkaJSError` which is retriable by default). Since retrying wouldn't change the list of brokers, this should be non-retriable.

PR #462, opened 2019-08-12
make NodeJS REPL get correct randomBytes()

Dieken:
NodeJS REPL imports crypto by default, this breaks the original logic.


Nevon:
Thanks for the fix!

PR #470, opened 2019-08-16
Prevent changes in the subscriptions for running consumers

ankon:
Such changes would be ignored by the consumer, so we better fail very on.

---
This is mainly to avoid frustration for users, an issue for being able to change the subscriptions for a running consumer will come as well


tulios:
Pre-release `1.11.0-beta.5` was published with this change

PR #473, opened 2019-08-18
Add `consumer.heartbeat()`

### LINKED ISSUE #387: Heartbeats can not be sent manually until messages have been consumed
This issue is of the same nature as #378, where it's not possible to perform heartbeats before having consumed any messages, as that possibility is only exposed as part of `eachBatch`.

It's common within stream processing tasks to pause the fetching of messages and not process any until some work is performed, like restoring a processing state store. Especially in that case, it's easy for this preparation work to outlast the `sessionTimeout`, causing a rebalance and restarting the process all over again. This can lead to no processing ever happening, with the consumers stuck in a rebalance loop.

The only chance the assignment has is to send heartbeats, indicating that it hasn't died, preventing the group coordinator from issuing a rebalance.

## Proposed solution

Introduce a `consumer.heartbeat` method. Like consumer.seek, it could limit it's calling to only after consumer.run has established the internal consumerGroup.

The Java `KafkaConsumer` doesn't have such a method, but that's because they use calling `consumer.poll` in combination with `max.poll.interval.ms`. In KIP-62, where this was added, [it's actually cited a possibility](https://cwiki.apache.org/confluence/display/KAFKA/KIP-62%3A+Allow+consumer+to+send+heartbeats+from+a+background+thread#KIP-62:Allowconsumertosendheartbeatsfromabackgroundthread-RejectedAlternatives) in the granularity of calling a heartbeat manually is needed.

> **Add a separate API the user can call to indicate liveness:** We considered adding a heartbeat() API which the user could use from their own thread in order to keep the consumer alive. This also solves the problem, but it puts the burden of managing that thread (including shutdown coordination) on the user. Although there is some advantage to having a separate API since it allows users to develop their own notion of liveness, we feel must users would simply spawn a thread and call heartbeat() in a loop. We leave this as a possible extension for the future if users find they need it.

Given that KafkaJS _does_ have such an heartbeat API (actually making it a lot easier to deal with this), calling it on the consumer seems to make sense!

Nevon:
Fixes #387

I was looking into how to implement rate limiting if using the `eachMessage` interface and realized that there needs to be a way to heartbeat manually. Seems we have the same need in #387, so I took a punt at implementing it.

### DIFF
--- a/docs/Consuming.md
+++ b/docs/Consuming.md
@@ -185,6 +185,22 @@ The usual usage pattern for offsets stored outside of Kafka is as follows:
 - Store a message's `offset + 1` in the store together with the results of processing. `1` is added to prevent that same message from being consumed again.
 - Use the externally stored offset on restart to [seek](#seek) the consumer to it.
 
+## <a name="manual-heatbeat"></a> Manual heartbeat
+
+In some circumstances, it may be necessary to heartbeat outside of the `eachBatch` callback, such as for example while restoring some external state before starting to process messages, or even from within `eachMessage` if processing a single message can sometimes take longer than the configured `sessionTimeout`.
+
+To manually commit, use `consumer.heartbeat()`:
+
+```js
+// To heartbeat, the consumer must have been initialized and be running
+await consumer.connect()
+consumer.run({ eachMessage: async () => {} })
+
+await consumer.heartbeat()
+```
+
+> Note: Heartbeating is how the brokers know the state of your consumer. Be careful to only heartbeat when your consumer is healthy.
+
 ## <a name="from-beginning"></a> fromBeginning
 
 The consumer group will use the latest committed offset when starting to fetch messages. If the offset is invalid or not defined, `fromBeginning` defines the behavior of the consumer group. This can be configured when subscribing to a topic:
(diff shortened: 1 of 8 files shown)


docs/Consuming.md
```diff
@@ -185,6 +185,22 @@ The usual usage pattern for offsets stored outside of Kafka is as follows:
...
+
+```js
+// To heartbeat, the consumer must have been initialized and be running
+await consumer.connect()
+consumer.run({ eachMessage: async () => {} })
+
+await consumer.heartbeat()
+```
+
+> Note: Heartbeating is how the brokers know the state of your consumer. Be careful to only heartbeat when your consumer is healthy.
```
JaapRood: Important note! 👌

tulios:
As we discussed F2F, with this change, the recovery logic is replicated three times, so it's worth to refactor. We also need to notify the process loop that the consumer is rebalancing, etc. We need to re-work somethings to get this in.

Nevon:
I've been thinking about how to do this. Refactoring the recovery logic so that it's not replicated is probably not too hard, but I'm not sure about how or where to notify the fetch loop.

The tricky bit is that both commitOffsets and heartbeat can be called when the fetch loop is not even running, in which case it should raise an error, but we also don't want to raise the error twice (once in the caller and once in the fetch loop). There's also a lot of places in the fetch loop where we might want to interrupt - at the beginning of a fetch, in between each batch, in between each message, etc. and I'm not sure where it makes sense to do so. I guess the point is that we want to complete the whole `heartbeat`/`commitOffsets` call, including error recovery and retries, before continuing with whatever the fetch loop is doing, to avoid triggering multiple joins or whatever.

JaapRood:
I think somehow we're going to end up with a finite-state-machine here, seen that we have a single state of consumer group (joined, rebalancing, unsubscribed, etc) that both governs and can be triggered by multiple instigators. Concretely, I think that means adding a 'state' flag in the runner, which each action dependent on this state, on every attempt, can query this state before either starting or handling the outcome of the action. That approach should allow for concurrent actions, rather than limiting the runner to only performing then sequentially.

More concretely, I can imagine a `retry(actionFn) {}` kind of method, which could wrap the action specific behaviour in the retry mechanism that checks for the runner's state and deals with the errors. But then I haven't tried, so not sure if that's practical 😅.

goriunov:
Is this going in any time soon ? Have a problem when one of the messages takes longer to process than session timeout it cases disconnect on the server side.

tulios:
@goriunov if you use `eachBatch` you have access to the `heartbeat` function. This PR needs some more work before we merge, we still want to move forward but it will take a while.

xozzslip:
In my case I want to consume 100MB batch and it is fetching quite long time. While it is fetching Kafka does not get any heartbeats from consumer. So now the only way to resolve this is to set heartbeat timeout to one minute or so.

liukun:
Can't we just send heartbeat outside of the loop by `setTimeout` or `setInterval`?

JaapRood:
> Can't we just send heartbeat outside of the loop by setTimeout or setInterval?

You totally could, and the broker would accept it, but as soon as you call `heartbeat` outside of the processing loop, it stops being a good indicator of health of that loop. The heartbeat is at the heart (🙌) of Kafka's failure detection mechanism, which in turn is vital for the fault tolerance behaviour it tries to implement.

> In my case I want to consume 100MB batch and it is fetching quite long time. 

Is that an individual message of 100 mb, or lots of messages? While Kafka's messages can be any sequence of bytes, it's not optimise for massive file transfers like this. I reckon if I was doing something like file uploads to Kafka, I'd be splitting them up into chunks and re-assemble them at the consumption end.

Nevon:
I'm closing this PR, as it's grown very stale and will need to be refactored.

Issue #451, opened 2019-08-02
Performance transparency


IdanAdar:
Can someone publish workload performance for kafka.js compared to other packages?

Nevon:
Duplicate of #398.

Issue #398, opened 2019-06-16
How about the performance compared with node-rdkafka?


leolcao:
I am searching some kafka client libraries in node.js, this library seems awesome, and lightweight, and i am only worry about the performance, so does anyone do some benchmark yet?

I am curious about it is not a wrapper from native librdkafka, it just use node.js 'net' and 'tls' to connect kafka cluster.

Nevon:
We have done benchmarks in isolation just to see what kind of throughput we could achieve, and we are running it in production under heavy load, but we've never done any direct comparisons with other libraries.

So far, it has been fast enough for our needs, so we haven't put any serious effort into optimizations. I would welcome a real comparison between the different libraries to see how we stack up.

JaapRood:
The library `node-rdkafka` binds to `librdkafka` is has got quite a few performance optimisations, mostly in the form of batching messages in both producing and consuming. It's actually done up to such a level that some stuff is pretty hard to work with. Pretty much everything is queued to then be executed in the background. That makes things like producing and committing "synchronously" a lot more difficult, as you need to call other apis and watch for delivery reports, etc. Pretty much all developer experience is sacrificed for message throughput

If necessary, the same strategies could be applied to `kafkajs`, at which point I reckon you could get pretty far in pure throughput compared to `librdkafka`. However, given at the level that Node works, I imagine it would never get that far down the performance vs. usability trade-off. 

Reality is that you probably won't need it to. The official Java Kafka client doesn't employ strategies up to that extreme point either, and is still used in production everywhere by big tech.

We haven't moved our highest volume topics from node-rdkafka to kafkajs yet, but not for a worry about performance. Unless you already have a big deployment and know exactly what kind of throughput you'll be facing, I wouldn't worry too much about it either.

leolcao:
In recent days, I tried kafkajs, also node-rdkafka. The scenario what we need is writing a kafka sink connector which can save data from kafka to mongodb, it need detecting the new kafka topics, and call the mongodb native driver to write data. The result is we choose node-rdkafka. It can support to 100 topics in 150 ~ 200ms interval in the same time, in other words, it can consume 500 ~ 600 topics in 1 second. kafkajs seems the much lower performance to subscribe many topics.

funduck:
## Write
I made a comparison  
[bench_kafka_drivers.txt](https://github.com/tulios/kafkajs/files/3500573/bench_kafka_drivers.txt)

### Usage difference
#### node-rdkafka
Config
```
[...]

PR #475, opened 2019-08-18
Benchmark encoding with expandable buffer

Nevon:
I've been toying with the idea of writing a buffer wrapper that avoids all the copying we're doing when encoding/decoding, so to see if it's worth it, I did some benchmarking:

1. The original encoder, using standard Buffer
2. "EncoderV2", which uses my custom Buffer wrapper

```
$ node benchmarks/encoder.js
Encode with native buffer x 79,612 ops/sec ±1.85% (89 runs sampled)
Encode with expanding buffer x 117,255 ops/sec ±2.59% (85 runs sampled)
Fastest is Encode with expanding buffer
```

In this test, I am just calling each of the Encoder methods once with some arbitrary data. It might not be 100% representative of a normal use-case, but it should give us a hint at least.

As you can see, the result on my particular machine is that the expandable buffer implementation is maybe ~50% faster. I didn't spend a lot of time on implementing the expandable buffer (as you can see from the complete lack of unit tests), so it might be that I'm doing something dumb somewhere and it could be even faster.

The general idea is quite simple. `ExpandableBuffer` is just a class that holds an internal `Buffer` instance and a `_readOffset` and `_writeOffset`. It is in fact pretty similar to the `Encoder` itself. On any write, we check if there's enough space in the allocated internal buffer. If there isn't, we create a new buffer that's `bufferSize * 3 / 2 + 1` bytes (or whatever the minimum is to be able to fit what we're writing). The point is that by over-allocating, we can reduce the number of copies we do.

I didn't bother making `ExpandableBuffer` have all the functionality of a regular `Buffer`, and instead just implemented what we were actually using.


JaapRood:
> The point is that by over-allocating, we can reduce the number of copies we do.

In my limited experiences optimising for allocations, that's a legit way to go. In our Avro encoder we over allocate, to then later expand if we were unlucky to run out of space in an edge-case. Even then we go from a worst-case scenario 2 allocations instead of a default of 2. Looks like a pretty neat implementation of that!

> It might not be 100% representative of a normal use-case, [...] I didn't bother making ExpandableBuffer have all the functionality of a regular Buffer, and instead just implemented what we were actually using.

This sort of highlights my main concern with this: it increases the code footprint, with more tests required, more work required if using some part of Buffer that wasn't ported, more unsafe allocations and complexity. All of that _could_ very well be worth it, but what's missing from the equation is the indication that this is somehow the bottleneck in apps / tasks written with KafkaJS. Has anyone measured KafkaJS performance in an production application that indicates this is a good place to spend complexity budget?

Not that I have any say in the matter of course 😅🤐.

Nevon:
> Has anyone measured KafkaJS performance in an production application that indicates this is a good place to spend complexity budget?

No, and that would be a good next step. I only did localized performance testing initially, because it's a lot easier to set up, but it would be awesome to run the kind of tests done in #451 to see if there's a measurable difference.

> This sort of highlights my main concern with this: it increases the code footprint, with more tests required, more work required if using some part of Buffer that wasn't ported, more unsafe allocations and complexity.

I'm not so worried about it because it's very contained complexity. It doesn't affect anything outside of the Encoder/Decoder and is easily testable. The kind of complexity I'm scared of is the kind that is spread out across the codebase and requires a bunch of context to understand.

JaapRood:
> I'm not so worried about it because it's very contained complexity.

That's definitely a good point, it's pretty much local to just the Encoder, right? I'd recommend not adding it as it's own 'util' module then, as doing so makes it seem like the rest of the codebase could (and maybe even should) be using that. Maybe something a few can remember, for a while, that it's only to be used by the encoder, but as a project grows (which it definitely seems to be doing 👏🎉) something to be easily missed at some point.

In terms of spending a 'complexity budget' that definitely drives the cost of it down, but would still  be great if there's something concrete on the other side of that equation, right? The tests in #398 is better than nothing at all, but they are awfully synthetic. 

I guess why I bring that up is that In our (also completely anecdotal!) experience, with the kind of stream processing, we run into limits of I/O (whether Kafka, fs, other services or postgres) before anything else and don't even get close to raw numbers like that. We'd have to go full "kappa" architecture, where all your state is replicated down locally (probably from Kafka), so all processing can happen without any further network trips, if not completely in-memory. However, there's a bunch of other limitations that prevent  that from being workable with KafkaJS at the moment, so would think it's wise to spend complexity budget there first and hoping one doesn't get into 'debt' in the first place 😅.

Nevon:
> That's definitely a good point, it's pretty much local to just the Encoder, right?

I suspect it would also be useful in `src/network/connection.js` (originally I thought the Decoder, but the Decoder itself is fine). When we read data from the socket, we copy over the entire buffer every time until we reach the amount of data needed to read the response (which can be rather large). But I haven't run any tests on that.

> In terms of spending a 'complexity budget' that definitely drives the cost of it down, but would still be great if there's something concrete on the other side of that equation, right?

Of course. I'm all for doing performance tests before taking on complexity, but I might need some help figuring out how to test it.

> I guess why I bring that up is that In our (also completely anecdotal!) experience, with the kind of stream processing, we run into limits of I/O (whether Kafka, fs, other services or postgres) before anything else and don't even get close to raw numbers like that.

Same, but those factors are kind of irrelevant when it comes to the performance of the client. If you have external services slowing down your processing, there's nothing we can do about that, but if we're spending a bunch of CPU time copying buffers back and forth, it means you'll need to run on more or beefier machines than you should need to.

JaapRood:
> Same, but those factors are kind of irrelevant when it comes to the performance of the client. If you have external services slowing down your processing, there's nothing we can do about that, but if we're spending a bunch of CPU time copying buffers back and forth, it means you'll need to run on more or beefier machines than you should need to.

Not sure what you mean here. If network I/O is the bottleneck of processing, then by definition the time spent on additional memory allocations is inconsequential for the end result. If bound by network, it would be doing fewer allocations for buffer copying by virtue of being back-pressured elsewhere, so actually meaning you could run on a machine with less CPU capacity. It's what the whole running the cloud on commodity (virtualised) hardware argument is about, right?

> Of course. I'm all for doing performance tests before taking on complexity, but I might need some help figuring out how to test it.

For our client work we'll be moving our more performance sensitive processing to KafkaJS (basically as soon as manual assignment #373 is in, but we might try to work around it). We have some benchmarks performing an actual live workload there for node-rdkafka. I can make sure we perform the same benchmark against KafkaJS as well and contribute that back! With a bit of luck can I share a runtime profile as well, which should help give further insight :)

Nevon:
> Not sure what you mean here. If network I/O is the bottleneck of processing, then by definition the time spent on additional memory allocations is inconsequential for the end result. If bound by network, it would be doing fewer allocations for buffer copying by virtue of being back-pressured elsewhere, so actually meaning you could run on a machine with less CPU capacity. It's what the whole running the cloud on commodity (virtualised) hardware argument is about, right?

Yes and no. You're right that of course if you are bottlenecked by some external system or network IO, then the additional time spent on memory allocation might only account for a small part of the end result. You may not be able to increase throughput beyond a certain point (the maximum throughput being bounded by external factors), but you may be able to keep the _same_ throughput with fewer resources if we are more efficient.

> I can make sure we perform the same benchmark against KafkaJS as well and contribute that back! With a bit of luck can I share a runtime profile as well, which should help give further insight

That would be really awesome. We do have some high-throughput (I guess. I don't have much to compare to) cases as well, but we haven't set it up to collect any detailed metrics.

> The tests in #398 are better than nothing at all, but they are awfully synthetic.

What about those tests is it that you feel is unrealistic? Nevermind the comparison with node-rdkafka, where it might be the case that it's testing two different things, but if we were comparing KafkaJS with a different version of KafkaJS, don't you think those tests are pretty reasonable? I'm trying to think how I can develop a benchmark suite that we could run periodically to test improvements/detect regressions.

JaapRood:
> What about those tests is it that you feel is unrealistic? Nevermind the comparison with node-rdkafka, where it might be the case that it's testing two different things, but if we were comparing KafkaJS with a different version of KafkaJS, don't you think those tests are pretty reasonable? 

It's unrealistic in the same way every synthetic benchmark is: it makes the gears spin but the input / output isn't connected to anything that would be a in-production use case. That's not to say there isn't some value in it: in this case it would definitely give you some idea whether the encoder cog of the machine is spinning with greater efficiency, so if that's the goal, it's fine. 

However, to be informed in whether the increases in efficiency have actually led to a machine  that performs it's work better when fully connected up, it doesn't tell you much or could even lead you astray. Back when I was still doing automotive engineering we had definitely optimized something in isolation before and had to make it stronger, only to realise later the part was strong enough before, but now the overall assembly got heavier, which actually has a detrimental effect on overall handling. Maybe not the strongest analogy, but ever since then something I've been weary of. 

> I'm trying to think how I can develop a benchmark suite that we could run periodically to test improvements/detect regressions.

That's great! While the best would be running and measuring things against production, that's obviously not as easily repeatable. In the end benchmarking will always be a trade-off between synthetic and real-world, with some hybrid.

Concretely, I'd say we'd have to have some test sets of data and processes that _represent_ real-world test cases, giving the synthetic tests some work to perform. Best would be anonymised production data, but even procedurally generated example data is generally better than just letting the cogs spin free without any resistance :)

Nevon:
To no one's surprise, there's a bug somewhere in my buffer implementation. Swapping it out as the default implementation of the Encoder fails lots of tests. 

I'm gonna close this PR for now. I was hoping to get a working version that I could toss into some end-to-end benchmarks, but I don't have the time and energy to do it right now.

PR #477, opened 2019-08-19
types: Improve definition of SASLOptions by adding a SASLMechanism type

ThisIsMissEm:
Previously if you tried to assign an environment variable to the mechanism option, then you'd get a typescript error of "cannot assign string to 'plain' | 'scram-sha-256' | 'scram-sha-512' | 'aws'" and the only workaround was to declare a type which copied the string union.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -33,8 +33,10 @@ export type ISocketFactory = (
   onConnect: () => void
 ) => net.Socket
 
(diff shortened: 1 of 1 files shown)


ThisIsMissEm:
Turns out this may not be needed, as you can apparently do `foo as SASLOptions["mechanism"]` — though this isn't documented much anywhere.

tulios:
I think this is still valid; it's a good improvement.
Thanks.

tulios:
Pre-release `1.11.0-beta.9` was published with the changes.
Thanks.

Issue #469, opened 2019-08-15
The producer attempted a transactional operation in an invalid state


eliw00d:
So, using the example given [here](https://kafka.js.org/docs/transactions), I tried to set up a transactional producer.

client:
```javascript
const client = new Kafka({
  clientId: 'my-client',
  brokers: ['my-broker'],
})
const producer = client.producer({ 
  maxInFlightRequests: 1, 
  idempotent: true,
  transactionalId: 'my-transaction'
})
```
transaction:
```javascript
const  transaction = await producer.transaction()

try {
  await transaction.send({ topic, messages })

  await transaction.commit()
} catch (e) {
  await transaction.abort()
}
```
As long as I send and commit without issue, it works. However, if any error is thrown, and abort is called, I get the following error:
```
[Connection] Response EndTxn(key: 26, version: 0)
    broker: "my-broker"
    clientId: "my-client"
    error: "The producer attempted a transactional operation in an invalid state"
    correlationId: 1
    size: 10
```
```
    {
      "type": "KafkaJSProtocolError",
      "message": "The producer attempted a transactional operation in an invalid state",
      "stack":
          KafkaJSProtocolError: The producer attempted a transactional operation in an invalid state
              at createErrorFromCode (/Users/eliw00d/Documents/GitHub/my-project/node_modules/kafkajs/src/protocol/error.js:536:10)
              at Object.parse (/Users/eliw00d/Documents/GitHub/my-project/node_modules/kafkajs/src/protocol/requests/endTxn/v0/response.js:24:11)
              at Connection.send (/Users/eliw00d/Documents/GitHub/my-project/node_modules/kafkajs/src/network/connection.js:306:35)
              at process._tickCallback (internal/process/next_tick.js:68:7)
      "name": "KafkaJSProtocolError",
      "retriable": false,
      "code": 48
    }
```

To better illustrate the issue, I even tried:
```javascript
const transaction = await producer.transaction()
await transaction.abort()
```
with the same result.

Is this a bug or am I missing something?

eliw00d:
Actually, I realized that this is probably intentional since there would be nothing to abort if send was never called. However, in our case, the try/catch wraps a database transaction and we want to ensure that if the database transaction fails we abort or cancel the producer transaction as well. 

Maybe this could be a feature request for something like `transaction.isAbortable` that we can call before `transaction.abort` to avoid this error? Like so:
```javascript
const  transaction = await producer.transaction()
try {
  // insert code that will create the messages here
[...]

PR #478, opened 2019-08-22
fix: correct value of ABORTING transactionState

eliw00d:
Fixes https://github.com/tulios/kafkajs/issues/469.


eliw00d:
This does not actually fix the issue entirely, I will have to do some more research.

tulios:
Good catch, thanks!

tulios:
Pre-release `1.11.0-beta.10` was published with the changes.

PR #497, opened 2019-09-13
Added confluent links to doc

Durairaj:
Confluent documents and blogs is definitely an useful resource when people trying to dig inito kafka. so i think its a valuable addition to the blog


Nevon:
I don't think the Confluent blog is particularly useful for developing KafkaJS (note, KafkaJS, not developing _with_ KafkaJS). However, the Confluent docs can be quite useful.

PR #505, opened 2019-09-23
Remove normalisation of the password when using SCRAM auth

### LINKED ISSUE #506: Incorrect normalisation of SASL SCRAM password
Stumbled across this when trying to authenticate with kafka using SASL SCRAM. It took quite a bit of debugging but narrowed down some problematic characters that were common across broken password. The characters being `,` and `=`.
Turns out KafkaJS incorrectly normalised the password using along with the username. However the RFC referenced in the source code https://tools.ietf.org/html/rfc5802#section-5.1, as far as I am aware, only calls for the normalisation of the username.

I have opened the following PR to solve this issue:  #505

ramesius:
KafkaJS was incorrectly normalised the SCRAM password, resulting in `,` and `=` characters being encoded causing authentication failures.

According to the cited RFC in the source code (https://tools.ietf.org/html/rfc5802#section-5.1), just the username needs to be normalised.

Closes #506

### DIFF
--- a/scripts/createScramCredentials.sh
+++ b/scripts/createScramCredentials.sh
@@ -10,8 +10,8 @@ find_container_id() {
 }
 
 USERNAME=${USERNAME:='testscram'}
-PASSWORD_256=${PASSWORD_256:='testtestscram256'}
-PASSWORD_512=${PASSWORD_512:='testtestscram512'}
+PASSWORD_256=${PASSWORD_256:='testtestscram=256'}
+PASSWORD_512=${PASSWORD_512:='testtestscram=512'}
 
 docker exec \
   $(find_container_id) \
--- a/src/broker/saslAuthenticator/scram.js
+++ b/src/broker/saslAuthenticator/scram.js
@@ -296,7 +296,7 @@ class SCRAM {
    */
   encodedPassword() {
     const { password } = this.connection.sasl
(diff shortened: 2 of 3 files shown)


ramesius:
Java client for reference: https://github.com/axbaretto/kafka/blob/master/clients/src/main/java/org/apache/kafka/common/security/scram/ScramFormatter.java#L101

Nevon:
Thank you very much for your contribution! I just want to reproduce this issue myself, in order to see that the proposed fix works, and then I'll merge it. :)

ramesius:
Thanks. I think I want to add some tests for special characters in the password but want some direction on how I should include something like that 👍

Nevon:
What I did was to simply change the password in [scripts/createScramCredentials.sh](https://github.com/tulios/kafkajs/blob/master/scripts/createScramCredentials.sh#L13-L14) to include `=`, update the passwords to match in [testHelpers/index.js](https://github.com/tulios/kafkajs/blob/master/testHelpers/index.js#L62-L80), and then run the existing tests. Without your fix, the tests failed, and with the fix, they pass. 👍

ramesius:
Perfect, thanks for that

Nevon:
This fix will be available in pre-release version `1.11.0-beta.12` as soon as [the build](https://dev.azure.com/tulios/kafkajs/_build/results?buildId=483) is done.

Issue #403, opened 2019-06-20
Consumer gets stuck when it receives a control batch


kkurten:
We have a simple data pipeline where events are consumed from topic A and written to topic B by __transactional__ (Java) Streams application. A nodejs microservice then consumes events from topic B (using kafkajs) and writes events to one of our APIs. 

We noticed that when kafkajs receives a control batch (https://kafka.apache.org/documentation/#controlbatch) that contains only a single control record the consumer gets permanently stuck. Since the batch contains only a control record, Kafkajs filters out that record from the batch and the batch becomes empty. Since we are not actually consuming any messages the offset is not advanced. The next fetch will receive the same control batch and the same thing happens again. At this point our consumer is basically in a busy loop, constantly fetching the same data and not advancing anywhere (cpu usage also spikes quite a lot). We have a lot of microservices that use kafkajs and this issue seems to only affect topics that contain control batches.

I have not experienced this issue with Java consumer but I'm not sure if this is actually a bug in kafkajs or "working as intended". We are using currently eachMessage handler in our kafkajs microservice so we don't have much control over how e.g. offsets are managed. Should we just use eachBatch in these kind of scenarios to manually advance the offset or is there perhaps something in the kafkajs implementation that could be improved to avoid this situation completely(?).

#### Kafka version
2.2

#### Kafkajs version
1.8

#### Kafka consumer group offsets
```
TOPIC           PARTITION          CURRENT-OFFSET   LOG-END-OFFSET     LAG             CONSUMER-ID                                                  HOST                        CLIENT-ID
test-topic-1    0                  38187            56935              18748           test-microservice-b0093482-f946-4af0-b632-7a31b9c8011a       127.0.0.1/127.0.0.1         test-microservice
test-topic-1    1                  33034            57693              24659           test-microservice-b0093482-f946-4af0-b632-7a31b9c8011a       127.0.0.1/127.0.0.1         test-microservice
test-topic-1    2                  35574            55538              19964           test-microservice-b0093482-f946-4af0-b632-7a31b9c8011a       127.0.0.1/127.0.0.1         test-microservice
```

#### Kafkajs logs
```
{"level":"DEBUG","timestamp":"2019-06-20T11:16:41.976Z","logger":"kafkajs","message":"[Connection] Connecting","broker":"192.222.134.113:9092","clientId":"test-microservice","ssl":true,"sasl":true}
[...]

PR #511, opened 2019-09-30
Skip control records without auto-resolve

tulios:
This PR solves issue #403.

It does two things:

1) Automatically resolve the control record offset when resolving the last non-filtered offset of the batch
2) When the batch is an empty control batch, automatically resolve the  offset

I wrote more on the code documentation; I will paste here for documentation purposes.

> The transactional producer generates a control record after committing the transaction.
> The control record is the last record on the RecordBatch, and it is filtered before it
> reaches the `eachBatch` callback. When disabling auto-resolve, the user-land code won't
> be able to resolve the control record offset, since it never reaches the callback,
> causing stuck consumers as the consumer will never move the offset marker.
> 
> When the last offset of the batch is resolved, we should automatically resolve
> the control record offset as this entry doesn't have any meaning to the user-land code,
> and won't interfere with the stream processing.

I am pinging the people in the original issue since it's a bit old and you might miss this PR.
@kkurten @drojas @JaapRood @jasine @brandonl

### DIFF
--- a/docs/Transactions.md
+++ b/docs/Transactions.md
@@ -30,7 +30,7 @@ const producer = client.producer({ maxInFlightRequests: 1, idempotent: true })
 Within a transaction, you can produce one or more messages. If `transaction.abort` is called, all messages will be rolled back.
 
 ```javascript
-const  transaction = await producer.transaction()
+const transaction = await producer.transaction()
 
 try {
   await transaction.send({ topic, messages })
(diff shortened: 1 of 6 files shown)


JaapRood:
Seems like a totally reasonable way to address this issue 👌👍

tulios:
Pre-release `1.12.0-beta.0` published with the fix.

PR #521, opened 2019-10-10
Adding Consumer GROUP_JOIN information in the docs

capucho:
The purpose of this is to make people aware that only `connect()` is not enough to actually being able to consume messages. After running into some issues with integration tests, this knowledge would have been really helpful, so it might be interesting for other people as well :)

### DIFF
--- a/docs/Consuming.md
+++ b/docs/Consuming.md
@@ -35,6 +35,8 @@ The consumer will not match topics created after the subscription. If your broke
 
 KafkaJS offers you two ways to process your data: `eachMessage` and `eachBatch`
 
+The consumer will be ready to process incoming messages to the topic after joining the consumer group, when it sends the `GROUP_JOIN` event.
+
 ## <a name="each-message"></a> eachMessage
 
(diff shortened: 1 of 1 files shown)


ankon:
When reading this I would now ask myself: "And how do I await that 'event'?"

I suspect one could link the text to https://kafka.js.org/docs/instrumentation-events, but maybe a bit of example code might be more helpful.

JaapRood:
When I first got started with KafkaJS I ran into this as well, as we wanted to hook into the point where new assignments were received. However, I agree with @ankon that this now reads as if listening for this event is necessary for consuming messages, which it's not.

capucho:
Thinking about it, maybe a section on how to actually do integration tests it would be better? It's something that you probably won't run into when running your service, but could often run into when doing integration tests. @ankon I've added a small piece of code on the issue.

Nevon:
I'm going to reject this PR, as I share Ankon's concerns about the current phrasing. I would like to have some documentation available on how to do integration tests though. I started working on it some weeks ago, but it's a fair bit of work.

Issue #329, opened 2019-04-02
Add timestamp parameter to fetchTopicOffsets


dstelljes:
Would it be possible to support fetching topic offsets by timestamp? If this sounds reasonable, I can open a PR:

1. Look for `timestamp` in `cluster.fetchTopicsOffset` as well as `fromBeginning`
2. Add optional `timestamp` parameter to `admin.fetchTopicOffset`

This is a continuation of #126—apologies for never following up.

PR #525, opened 2019-10-17
Add timestamp parameter to fetchTopicOffsets

rvighne:
Add support for fetching offsets that correspond to a particular timestamp (implements #329).


rvighne:
Not sure what was wrong with the tests before, but they passed after I re-ran them by merging the latest master.

ShanonJackson:
Can we get this merged haha? Need this functionality really bad

Nevon:
Thanks for the contribution, but we're going to try to merge #604 instead, which adds a new public method rather than extending `fetchTopicOffsets` with another parameter.

PR #530, opened 2019-10-24
Fix right-side bar hrefs in docs

pedroscaff:
The hrefs with `&` still don't work properly, but it is an issue with https://github.com/facebook/docusaurus, I will report it there and post it here.

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -18,7 +18,7 @@ The option `retry` can be used to customize the configuration for the admin.
 
 Take a look at [Retry](Configuration.md#default-retry) for more information.
 
-## <a name="create-topics"></a> Create topics
+## Create topics
 
 `createTopics` will resolve to `true` if the topic was created successfully or `false` if it already exists. The method will throw exceptions in case of errors.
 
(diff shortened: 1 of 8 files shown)


pedroscaff:
https://github.com/facebook/docusaurus/issues/1877

Nevon:
Thanks for reporting the issue to docusaurus. I won't merge this PR, however, as we intentionally create our own anchor links. The reason is that we ship references to the documentation as part of KafkaJS' error messages, which means that we need the links to never change, even if we end up changing the title or Docusaurus changes the way they generate anchors.

pedroscaff:
Do you think that is the best way to do it? I think breaking the documentation navigation is not a good solution for the issue. Docusaurus generates the anchors as we would expect (bugs aside hehe), so we could stick to what they generate.

Nevon:
> I think breaking the documentation navigation is not a good solution for the issue.

No, of course, we don't want the sidebar nav to break either. I think facebook/docusaurus#561 is the correct solution.

Maybe a workaround could be to generate the anchor via Docusaurus and _also_ have our own stable anchor. That way the links in error messages keep working, while the sidebar nav also works, even if they link to different anchors.

PR #533, opened 2019-10-24
Update Configuration.md

jweinsteincbt:
Recommend setting `ssl: true`  when using SASL, otherwise SASL will not work properly.


Nevon:
Thanks for the contribution!

Technically there's nothing stopping you from using SASL without SSL (other than the fact that your credentials will be sent over an unencrypted connection). It's called `SASL_PLAINTEXT` on the broker side. I'm not sure if we even support that in KafkaJS though, and I'm quite sure the examples you modified do require it.

PR #535, opened 2019-10-25
Clarified impact on manually committing without auto resolving offsets

brandonl:
I might be miss understanding something but AFAIK the docs for manually committing batches leads one to believe failing to commit offsets (under all circumstances) wont effect progress of batch fetching only restarts. Happy to be educated here if I'm mistaken or to reword.

### DIFF
--- a/docs/Consuming.md
+++ b/docs/Consuming.md
@@ -160,7 +160,7 @@ When disabling [`autoCommit`](#auto-commit) you can still manually commit messag
 - By [sending message offsets in a transaction](Transactions.md#offsets).
 - By using the `commitOffsets` method of the consumer (see below).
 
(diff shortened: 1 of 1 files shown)


JaapRood:
I think you've hit on a bit of a deeper problem within the documentation, which is the difference between _resolving_ and _committing_ offsets. The latter is a Kafka primitive users see in various kafka implementations and behave like advertised: it only controls from where processing starts, it doesn't advance processing. The former, _resolving_, is a more complicated term, as it controls both the _consumed_ offset pointer, but also influences the _committed_ offset.

@tulios I recall us discussing this a while back, wondering whether `checkpoint` wouldn't make for a less ambiguous alternative to `resolve`.

tulios:
Yes, `checkpoint` is definitely a better term. We can add a new section describing more in depth what `resolve` is and how it works.

PR #536, opened 2019-10-26
brokers now can be a string with comma separated brokers

silversoul93:
What do you think about this little feature?
I did it because I use your library and I needed to split a comma separated list of brokers

### DIFF
--- a/src/cluster/connectionBuilder.js
+++ b/src/cluster/connectionBuilder.js
@@ -7,6 +7,13 @@ const validateBrokers = brokers => {
   }
 }
 
+const splitBrokersIfNeeded = brokers => {
+  if (Array.isArray(brokers)) {
+    return brokers
+  }
+  return brokers.split(',')
+}
+
 module.exports = ({
   socketFactory,
   brokers,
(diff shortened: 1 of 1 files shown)


JaapRood:
It works, but I'm having a hard time coming up with a reason for why this logic has to live in the client. The logic is the canonical example of using `Array.prototype.split`, with no further parsing required. Is it unreasonable to expect users of the client to call split on a comma separated broker list themselves?

silversoul93:
You are right it's not a necessary feature; I did it because it's common in lots of kafka libraries to accept brokers as comma separated string and, moreover, I had the need to take the brokers list by environment variable as a comma separated values string and I had to split it; I know, is one line of code xD but I thought that could be useful so I did it

tulios:
Hi @silversoul93, I am not a big supporter of variadic types for method signatures, like accepting an array or a string. I will +1 @JaapRood and say that we shouldn't have this in the client. I chose to accept an array because I think we shouldn't mix concerns, the way you receive the hosts in your app is your business and the client shouldn't force you into a particular format, so I chose a regular javascript array. I've seen some libraries creating some protocol for the hosts, something like `kafka://host1,host2,host3`, I think this just adds complexity and confusion, as beginners will think that the prefix `kafka:` it's a thing.

Thanks for the PR, but we won't merge at this time. I hope we will see other PRs from you; I really appreciate the initiative. Thanks again.

PR #542, opened 2019-11-05
Add yarn to the devDependencies

ankon:
This is needed for running tests and other development tasks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,8 @@
     "prettier": "^1.18.2",
     "semver": "^6.2.0",
     "typescript": "^3.5.3",
-    "uuid": "^3.3.2"
+    "uuid": "^3.3.2",
+    "yarn": "^1.19.1"
   },
   "dependencies": {
     "long": "^4.0.0"


Nevon:
Yarn should not be a devDependency, as it is the package manager used to install said dependencies. Theoretically we could adapt the npm scripts to work with either npm or yarn, but since the lockfile is in yarn's format, we expect yarn to be installed globally and be used to install dependencies regardless.

ankon:
Hmm. 

Fair enough, but I think in that case I think it should be mentioned explicitly in the contribution guide (https://kafka.js.org/docs/development-environment / https://kafka.js.org/docs/contribution-guide). Note that apart from running the tests I didn't have any other issues, so I guess with a caveat of "you know what you're doing" either a global or local install will work -- but maybe for the "canonical" environment then the guide could add this to the TL;DR (`npm install -g yarn`) and a short note on the development-environment that this project is "... developed with the yarn package manager". WDYT?

Nevon:
Yeah that sounds reasonable. I think it could go in the same spot where we currently say that `docker` and `docker-compose` need to be installed.

> When developing KafkaJS, we run a Kafka cluster in a similar way to what is described in Running Kafka in Development, using docker and docker-compose. Before you proceed, make sure that you have both docker and docker-compose available.

PR #544, opened 2019-11-05
Clarify the behavior in the partition assigner protocol

ankon:
The existing typing was off, and made it look like the user data would be provided _to_ the `protocol` and `assign` functions as argument fields.

Reading through https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-GroupMembershipAPI and the KafkaJS code is not true at all. What actually happens is that
1. each assigners' `protocol` function _can_ provide user data (by encoding it into the result), then
2. Kafka will select a leader, and provide that one in the JoinGroupResponse with an array of all members _and the metadata they provided_, which
3. the group leader will then process through the `assign` function. At this point it can take the user data into account, and it can produce new user data* 

The first commit in this PR updates the TypeScript types to match this behavior and adds a integration test to demonstrate it, the second commit then cleans up the roundRobinAssigner's code to remove the "obviously unused" argument fields.

---
* From what I can see and read in the documentation the user data in the response is unused for the 'consumer' protocol, but likely is used for more fancier protocols, for example in the 'connect' protocol (https://www.confluent.io/blog/incremental-cooperative-rebalancing-in-kafka is fancy!)

### DIFF
--- a/src/consumer/assigners/roundRobinAssigner/index.js
+++ b/src/consumer/assigners/roundRobinAssigner/index.js
@@ -10,11 +10,15 @@ module.exports = ({ cluster }) => ({
   version: 1,
 
   /**
+   * Assign the topics to the provided members.
+   *
+   * The members array contains information about each member, `memberMetadata` is the result of the
+   * `protocol` operation.
+   *
    * This process can result in imbalanced assignments
    * @param {array} members array of members, e.g:
-                              [{ memberId: 'test-5f93f5a3' }]
+                              [{ memberId: 'test-5f93f5a3', memberMetadata: Buffer }]
    * @param {array} topics
-   * @param {Buffer} userData
    * @returns {array} object partitions per topic per member, e.g:
    *                   [
    *                     {
@@ -33,7 +37,7 @@ module.exports = ({ cluster }) => ({
    *                     }
    *                   ]
    */
-  async assign({ members, topics, userData }) {
+  async assign({ members, topics }) {
     const membersCount = members.length
     const sortedMembers = members.map(({ memberId }) => memberId).sort()
     const assignment = {}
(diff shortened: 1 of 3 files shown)


ankon:
#231 may need to be reviewed: It added the userData parameter in some places where this PR removes it again from.

ankon:
This looks "big", but is actually just typing fixes + a unit test that this actually "works".

src/consumer/assigners/roundRobinAssigner/index.js
```diff
@@ -59,18 +63,16 @@ module.exports = ({ cluster }) => ({
       memberAssignment: MemberAssignment.encode({
```
Nevon: Shouldn't it provide the `userData` for that member in the `MemberAssignment`? The only difference is that the `assign` method gets the `userData` from each member (as `memberMetadata`). Maybe I'm missing something.

Nevon:
Sorry that it's taking forever to get this reviewed. I'm on vacation at the moment, and I suspect Tulio has his plate full.

src/consumer/assigners/roundRobinAssigner/index.js
ankon: The user data is part of the `members`, encoded into the `memberMetadata` by the `protocol` function.

The protocol function decides about the user data (in the RR assigner: nothing), and sends that to the broker as part of the member metadata. The broker then collects the members (and their metadata), and triggers the `assign` call (for the selected leader). The `assign` function implementation then call pull out the user data from the provided content.

The flow is indeed a bit odd, and I was arguing with myself here whether to leave some references lying around, but ultimately the RR assigner doesn't use any user data, and having it here is just confusing. Instead I added the unit test to show how to pass user data. Maybe a "how to write your own assigner" article could also help.

PR #551, opened 2019-11-07
Handle an empty assignment for a group member correctly

### LINKED ISSUE #550: RangeError in consumer group syncGroup response processing for member without assignment
~~~~
 Crash: KafkaJSNumberOfRetriesExceeded: Attempt to access memory outside buffer bounds ({"groupId":"XXXX","retryCount":0,"stack":"KafkaJSNumberOfRetriesExceeded
  Caused by: RangeError [ERR_BUFFER_OUT_OF_BOUNDS]: Attempt to access memory outside buffer bounds
    at boundsError (internal/buffer.js:70:11)
    at Buffer.readInt16BE (internal/buffer.js:492:5)
    at Decoder.readInt16 (/app/node_modules/kafkajs/src/protocol/decoder.js:36:31)
    at Object.decode (/app/node_modules/kafkajs/src/consumer/assignerProtocol.js:71:24)
    at ConsumerGroup.sync (/app/node_modules/kafkajs/src/consumer/consumerGroup.js:160:48)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at /app/node_modules/kafkajs/src/consumer/runner.js:52:9
    at /app/node_modules/kafkajs/src/consumer/runner.js:354:11"})
~~~~

From my logs it seems that at that point this member didn't receive an assignment, so this piece of code fails:
https://github.com/tulios/kafkajs/blob/34824cbcb2d9721441518ad6875e48f604fd3864/src/consumer/assignerProtocol.js#L71

I think there needs to be a check in any of these places, I'm just not sure which one:
1. Check in ConsumerGroup.sync whether the memberAssignment is empty
2. Check in MemberAssignment.decode whether the given buffer is empty

ankon:
Fixes #550

The argument for checking this here, rather than in the decode() function: Here we know that we don't care about the `version` property, but only about the actual assignment. Checking here is easy therefore, while in the decode() function I would have to come up with a reasonable return value, which would lead to possibly more changes to handle these.

### DIFF
--- a/src/consumer/consumerGroup.js
+++ b/src/consumer/consumerGroup.js
@@ -157,7 +157,8 @@ module.exports = class ConsumerGroup {
       groupAssignment: assignment,
     })
 
-    const decodedAssignment = MemberAssignment.decode(memberAssignment).assignment
+    const decodedAssignment =
+      memberAssignment.length === 0 ? {} : MemberAssignment.decode(memberAssignment).assignment
     this.logger.debug('Received assignment', {
       groupId,
       generationId,


ankon:
I think one could add a test for this, similar to #544.

Nevon:
Sorry that it took so long to take a look at this. I ended up opening up #567, because while this solution works for this case, I want to make sure that the MemberAssignment decoder doesn't crash if it receives an empty buffer. Otherwise if we have any other place that uses the MemberAssignment decoder, we have to make sure that that also does this check.

PR #555, opened 2019-11-12
Azure: New EventHubs (Kafka Surface) don't have an initial offset for nextOffset()

TheGauntt:
Hi,

I ran into a strange issue today. First, a little background. We use "Kafka" through [Azure's EventHubs](https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-for-kafka-ecosystem-overview).

So, it's not 100% Kafka, and I don't know that they implement the protocol exactly. That said, it works great. We ran into an issue today where a Kafka topic we hadn't used in a few days was crashing the kafkajs client. It was really strange, because both our Production and Staging Kafka topics were working perfectly. Stranger still that even after deleting the topic, the cluster, etc... same issue.

In a nutshell, the issue was this line over and over again:
```
12T21:00:25.969Z","logger":"kafkajs","message":"[Consumer] Crash: KafkaJSNumberOfRetriesExceeded: Cannot read property 'low' of undefined"
```

Which comes from the nextOffset method in OffsetManager:
```
let offset = this.resolvedOffsets[topic][partition];

if (Long.fromValue(offset).equals(-1)) {
```

I was able to fix it by changing `nextOffset()` like this:
```
let offset = this.resolvedOffsets[topic][partition] || '0'
```

Stranger still, once I had successfully done at least one read (from Consumer), it worked fine (even without my change). I think this is because it then had a commit log to use here:
```
this.resolvedOffsets[topic][partition] = this.committedOffsets()[topic][partition]
```

Anyway, this worked for me. It may be a reasonable default. I'm not quite sure how my Topic got into this state to begin with or why this happens for all new Azure Event Hubs. It may just be an issue on Azure's side

Let me know if I can explain this further or help test if there's a better solution


TheGauntt:
Noticed that the tests are failing. Is this the right way to run them locally?
```
yarn test
```

When I do that, I see:
```
Summary of all failing tests
 FAIL  src/utils/shuffle.spec.js
  ● Utils > shuffle › shuffles

    > 15 |     expect(shuffle(array)).toEqual([1, 3, 2])
         |                            ^
      16 |     expect(shuffle(array)).toEqual([3, 2, 1])
      17 |     expect(Math.random).toHaveBeenCalledTimes(6)
      18 |   })

```

```
Test Suites: 1 failed, 199 passed, 200 of 238 total
Tests:       1 failed, 750 passed, 751 total
Snapshots:   3 passed, 3 total
Time:        43.429s
```

ankon:
> We ran into an issue today where a Kafka topic we hadn't used in a few days was crashing the kafkajs client [...]

A bit on a tangent: This might be caused by a too small retention time for the offsets, in "normal" Kafka you'd configure that on the broker: `offsets.retention.minutes`. See https://cwiki.apache.org/confluence/display/KAFKA/KIP-186%3A+Increase+offsets+retention+default+to+7+days for some more details around this topic.

ankon:
> Crash: KafkaJSNumberOfRetriesExceeded: Cannot read property 'low' of undefined

We have seen this error a couple of times as well, never had time to debug this. Will try to see whether I still have logs around ...

TheGauntt:
@ankon Thanks for getting back to me. We had retention at one day, I think. I bumped it to seven

Update on repro steps:
- Using a new consumer groupId causes the same problem

Also, here's how I'm constructing my Consumers:

```
LOG_PREFIX = "myapp..."

config = {
	"host": "OurThing.servicebus.windows.net:9093",
	"connectionString": "Endpoint=sb://OurConnectionString...",
	"clientId": "some-debug-value",
	"topics": ["topic1", "topic2"],
	"groupId": "pancakes"
}

...

this.client = new Kafka({
	clientId: config.clientId || 'some-debug-name',	// debug name for viewing logs
	brokers: [config.host],
	ssl: {
		rejectUnauthorized: false
	},
	sasl: {
		mechanism: 'plain',
		username: '$ConnectionString',		// azure method of auth
		password: config.connectionString
	},
	logLevel: debug ? logLevel.DEBUG : logLevel.INFO
})

this.consumer = this.client.consumer({
	groupId: groupId
});

await this.consumer.connect();
console.info(LOG_PREFIX, `Connected. Subscribing to topics:`, topics.join(', '));

for (let topic of topics) {
	await this.consumer.subscribe({
		topic: topic,
		fromBeginning: true
	})

	console.info(LOG_PREFIX, `Subscribed to ${topic}`);
}
```

mmacai:
Can confirm, we are experiencing the same issue with similar setup.

songpr:
We have the similar issue with Azure Event Hub after topic is inactive for days.
The retention have been set to 7 days so I don't think it is the cause.

Below is log from kafkajs
`{"level":"ERROR","timestamp":"2019-11-26T03:13:05.419Z","logger":"kafkajs","message":"[Consumer] Crash: KafkaJSNumberOfRetriesExceeded: Cannot read property 'low' of undefined","groupId":"rich_menu","retryCount":0,"stack":"KafkaJSNumberOfRetriesExceeded\n  Caused by: TypeError: Cannot read property 'low' of undefined\n    at Function.fromValue (/usr/src/app/node_modules/long/src/long.js:286:25)\n    at OffsetManager.nextOffset (/usr/src/app/node_modules/kafkajs/src/consumer/offsetManager/index.js:63:14)\n    at partitionsPerLeader.(anonymous function).map.partition (/usr/src/app/node_modules/kafkajs/src/consumer/consumerGroup.js:373:45)\n    at Array.map (<anonymous>)\n    at ConsumerGroup.fetch (/usr/src/app/node_modules/kafkajs/src/consumer/consumerGroup.js:371:58)\n    at process._tickCallback (internal/process/next_tick.js:68:7)"}`

beneramos:
Unsubscribe

Em ter, 26 de nov de 2019 01:13, songpr <notifications@github.com> escreveu:

mmacai:
Last week (pretty much 7 days ago) we used the trick that @TheGauntt mentioned and it worked for a while, now it hit the same problem. Assume retention period might cause it (7 days in our case).

tulios:
I am taking another stab at this issue today

mmacai:
@tulios what I did to reproduce it:

- Saturday - created new eventhub (topic) in Azure with 1 day retention period
- Saturday - joined with consumer group A, produced message, all good
- Monday - joined with consumer group A, all good
- Monday - joined with consumer group B, got mentioned error

tulios:
We discovered that this was caused by a race condition when mutating the offsets in the offset manager, can you try the code in #578?

TheGauntt:
@tulios 
I'm seeing the same issue as before with #578 

```
nextOffset(topic, partition) {
    if (!this.resolvedOffsets[topic][partition]) {
      this.resolvedOffsets[topic][partition] = this.committedOffsets()[topic][partition]
    }

    let offset = this.resolvedOffsets[topic][partition]    // offset is undefined
    if (Long.fromValue(offset).equals(-1)) {             // crashes here
```

tulios:
@TheGauntt can you try this PR: https://github.com/tulios/kafkajs/pull/582

TheGauntt:
@tulios That worked! 🎈 🍰 🎉

mmacai:
With the fix in #582 we are actually experiencing different problem now. If we create new consumer group for some topic it successfully connects consumer, but no messages are consumed. Seems like the changes in #582 just detect partition without any offset and throw it away. I would say it should try to get offset from broker, or?

songpr:
@tulios the #582 do not fix this case. I have test with Azure Eventhub, if the topic is no sent message for days, with this fix it no error message.
However, when send message to the topic, messages are not consumed.
Just as @mmacai  have mentioned.

songpr:
> 
> 
> @tulios That worked! 🎈 🍰 🎉

@TheGauntt is it really work in Azure Eventhub? The fix is clear the error message but new messages after that are not consumed.

tulios:
@songpr I can take another look at this. Can you help us with the azure event hubs data? It's very to reproduce since we don't use Azure Event hubs, I will try to reach out to Microsoft and see if we can get some free hours to debug this problem.

TheGauntt:
@tulios I should be able to set you up with an Event Hub and how to connect if that helps. Send me a DM

@songpr  I haven't debugged too deeply - our messages seem to work but I haven't tried
a) A new consumer group
b) A new topic
c) Verifying 100% of messages are going through
d) Testing messages where I saw the problem in the first place (our production cluster always seemed fine)

I'll update when I've done A-D

tulios:
@TheGauntt that would be amazing, I think we can debug this much quicker if we can easily simulate the problem. How do you want me to DM you? You could join our slack channel, and we take it from there.

justinmchase:
@TheGauntt Are you doing anything with certs to connect? I'm unable to connect to evenhubs with that same connection configuration you have there. I'm wondering if anything has changed or if you have any information about configuring eventhubs to make this work?

PR #567, opened 2019-11-28
Handle receiving empty member assignment

### LINKED ISSUE #550: RangeError in consumer group syncGroup response processing for member without assignment

Nevon:
Fixes #550.

This is a slightly different approach than #551 because I wanted to make sure that the MemberAssignment decoder could handle receiving a buffer with insufficient data, instead of expecting the caller to do the checking for it.


src/consumer/assignerProtocol.js
```diff
@@ -68,6 +68,10 @@ const MemberAssignment = {
     const indexAssignment = (obj, { topic, partitions }) =>
       Object.assign(obj, { [topic]: partitions })
 
+    if (!decoder.canReadInt16()) {
+      return null
+    }
+
```
ankon: As I mentioned in #551: Doing this does mean that callers need to be changed to be aware of the potential `null` return value. I think it would be good to modify the JSDoc hint then to `@returns {Object|null}` as well.

Nevon: Definitely. Missed that. Fixed in 48eee27

types/index.d.ts
```diff
@@ -338,7 +338,7 @@ export type MemberAssignment = {
 
 export const AssignerProtocol: {
   MemberMetadata: ISerializer<MemberMetadata>
-  MemberAssignment: ISerializer<MemberAssignment>
+  MemberAssignment: ISerializer<MemberAssignment | null>
```
ankon: That means MemberAssignment#encode() would need to handle `null` as well now, right?

src/consumer/assignerProtocol.js
JaapRood: Whenever I see something returning `null`, I get a little bit weary, especially because of these kinds of consequences. My question would be, how could `MemberAssignment` itself contain the fact that there's no topic-partitions to it's assignment? Technically, the consumer _was_ considered in the rebalance by the leader, and it's assignment just happens to be 0 topic-partitions. If, for example, this return an array of topic-partition pairs, it would simply return an empty array, rather than `null`.

However, with the current structure, I'm not how that was to be represented, and `null` might be the only pragmatic option. Worth discussing at least, I hope.

Nevon: Yeah, that was briefly mentioned in #551, that it would be ideal if `MemberAssignment` could return a sensible value to represent the absence of an assignment. In a different language, I would have had it return something like `Optional<MemberAssignment>`, but we don't have such a construct and the codebase is not structured around encoding non-results as anything but null.

types/index.d.ts
Nevon: Right! Took me a bit to see what you meant, but yes indeed, it would mean that `encode` would also need to handle receiving null. 

I changed it now so that only the `decode` has a nullable return type.

tulios:
@Nevon can you add an integration test?

src/consumer/assignerProtocol.js
ankon: One potential idea could be to introduce a magic constant for "NO_ASSIGNMENT", so at least it becomes obvious what this `null` means ... and of course at that point you don't have to make it null, but rather use an actual value like `{}`.

```typescript
export interface RealMemberAssignment = {
  version: number
  assignment: Assignment
  userData: Buffer
}
export const EMPTY_MEMBER_ASSIGNMENT = {}
export type MemberAssignment = RealMemberAssignment | typeof EMPTY_MEMBER_ASSIGNMENT
export const AssignerProtocol: {
  MemberMetadata: ISerializer<MemberMetadata>
  MemberAssignment: ISerializer<MemberAssignment>
}
```

(Only typed this in Github now, would have to play with that idea a bit and see how the rest of the code feels with that qua readability)

JaapRood: Fair enough!

Nevon:
> @Nevon can you add an integration test?

Done! 9c97aac

Issue #370, opened 2019-05-22
5 sec lag with multi broker


vykimo:
Hi everyone, I noticed a very special behavior of kafkajs.
When I use only one broker I consume immediately  messages produced.
But when I have a multi broker node, I get systematically a 5 sec lag between production and consumption.

I tried with kafka console in order to know if it is a bad configuration but no, produced messages are immediately written into consummer console.
Maybe you have already heard of similar issue.
Let me know.

tulios:
Without a lot of investigation, I can say that the default `maxWaitTime` is 5s, the fetch request will block for `maxWaitTime` or `minBytes`, so I am guessing that when the topic doesn't have new messages and you publish it will still wait for 5s (since the brokers have to sync due to `acks=-1` by default). On the next request, the `minBytes` rule will be satisfied, and the data will return right away.

vykimo:
So this behavior doesn't happens with one broker because sync isn't required, is it?

Then, to fix it I have to change `acks` value and/or lower `maxWaitTime` and `minBytes`?

vykimo:
Hmm so, `minBytes` is set by default to 1 (my messages are always greater than 1byte) and `acks` is mandatory set to -1 to guarantee idempotency.
So the only parameter I am able to change is `maxWaitTime` right? 
If I set it to 100ms, is it too low or it is ok?

vykimo:
So it's working with 100ms for maxWaitTime.
But I have 3 questions :
1. Why when have only one broker, messages are immediately read and when I have multi-brokers they wait for maxWaitTime (I confirmed this behavior again. With same messages and 1, then 2 brokers. bug?)
2. What are the consequences to drop `maxWaitTime` from 5000 to 100ms ?
3. Where can I set `maxWaitTime` in a constructor or something (I force it in kafkajs code currently) ?

tulios:
Hey @vykimo, sorry for the delay.

> Where can I set maxWaitTime in a constructor or something (I force it in kafkajs code currently) ?

You can configure the `maxWaitTime` when creating the consumer https://kafka.js.org/docs/consuming#a-name-options-a-options

The option is called `maxWaitTimeInMs`, e.g:
```javascript
kafka.consumer({ groupId: 'test', maxWaitTimeInMs: 100 })
```

> What are the consequences to drop maxWaitTime from 5000 to 100ms ?

[...]

PR #570, opened 2019-11-29
Replace fetch promise all with async generator

tulios:
This PR fixes issue #370, and it can improve the overall performance of KafkaJS (pending benchmark).

## Explaining the problem:

The fetch operation has its responsibility split into two modules: `runner` and `consumerGroup`. The runner is responsible for managing the consumer life-cycle and feeding the data to the userland code. The consumer group module abstracts the fetch operation by hiding the complexity of leadership, multiple brokers, and partitions.

Issue #370 highlighted the following problem, given a topic with 3 partitions where partition 0 hasn't received any messages in a specific t0, a call to `consumerGroup#fetch` would block the processing for `maxWaitTimeInMs`, usually 5s, even if the other partitions had data. This behavior is happening because `consumerGroup#fetch` uses `Promise.all` to wait for the requests to all partition leaders, and `runner#fetch` waits for the list of batches before it feeds any data to the user-land code.

## Proposed solution:

`consumerGroup#fetch` returns an async generator that can yield new results over time, so using the previous example, the user-land code would start processing messages from the fastest partition first, without paying any upfront cost for the partitions without data.

## One drawback:

We won't be able to provide `numberOfBatches` to the `FETCH` instrumentation event since the batches will be yielded overtime. I think it is a fine compromise, WDYT?

@vykimo @bobzsj87 @IvanovOleg I know this is an old issue and you might have moved away, but any input would be appreciated.


src/consumer/runner.js
```diff
@@ -297,24 +299,63 @@ module.exports = class Runner {
...
-    )
 
-    await this.autoCommitOffsets()
-    await this.consumerGroup.heartbeat({ interval: this.heartbeatInterval })
+    if (!this.running) {
+      return
+    }
+
+    const enqueuedTasks = []
+    let expectednumberOfExecutions = 0
```
ankon: Nit-pick: `expected_N_umberOfExecutions` reads better.

src/utils/bufferedAsyncIterator.js
```diff
@@ -0,0 +1,94 @@
...
+    })
+    .catch(async e => {
+      try {
+        await handleError(e)
+        emitter.emit('error', e)
+      } catch (newError) {
+        emitter.emit('error', newError)
+      }
+    })
+}
```
ankon: This would read (slightly) better with async/await:
```js
  try {
    const result = await promise
    if (emitter.listenerCount('data') > 0) {
      emitter.emit('data', { result, id: i })
    } else {
      results.push(result)
    }
  } catch (e) {
    try {
      await handleError(e)
      emitter.emit('error', e)
    } catch (newError) {
      emitter.emit('error', newError)
    }
  }
```

src/consumer/consumerGroup.js
```diff
@@ -437,45 +439,48 @@ module.exports = class ConsumerGroup {
...
+      await this.recoverFromOffsetOutOfRange(e)
+    }
 
-      throw e
+    if (e.name === 'KafkaJSBrokerNotFound') {
+      this.logger.debug(`${e.message}, refreshing metadata and retrying...`)
+      await this.cluster.refreshMetadata()
     }
+
+    throw e
```
JaapRood: For all these cases where we handle the error, shouldn't we return from the function afterwards? Or is there some reason why we still want the error to be re-thrown?

tulios: We want the error to bubble up and trigger the retry mechanism, the error handling here apply what should happen after the error, the regular flow is responsible for handling the new state.

src/consumer/runner.js
tulios: 02b874d

src/utils/bufferedAsyncIterator.js
tulios: 02b874d

tulios:
We will need to bump to node 10. Node 8 LTS reaches end of life in the end of the year, so we can release this after that.

src/consumer/runner.js
```diff
@@ -297,24 +299,63 @@ module.exports = class Runner {
...
+              if (!this.running) {
+                return
+              }
+
+              if (batch.isEmpty()) {
+                return
+              }
+
+              await onBatch(batch)
+              await this.autoCommitOffsets()
```
Nevon: This changes the behavior compared to before. Unless I'm mistaken, before we would only autoCommitOffsets and heartbeat after all the batches, not in between.

tulios: you are right, but I think we should have the hearbeats here, since they won't trigger if the interval is not right.

src/consumer/consumerGroup.js
JaapRood: The retry mechanism, of course! Missed that at the late Friday afternoon 😅.

src/consumer/runner.js
tulios: 9c24622

tulios:
The tests are green; I will wait for the next release and the end of node 8 LTS support before I merge.

andrewreineke:
Just wanted to echo support for this change since we're currently using KafkaJS in production and have a large number of topic-partitions in anticipation of future load - and as such are paying the E2E latency tax.

Node 8 LTS is EOL as of the new year, so thanks for having this on your radar!  (and also for all the work maintaining the rest of KafkaJS)

Cheers

src/consumer/runner.js
```diff
@@ -297,22 +307,62 @@ module.exports = class Runner {
...
+              if (!this.running) {
+                return
+              }
+
+              if (batch.isEmpty()) {
+                return
+              }
+
+              await onBatch(batch)
+              await this.consumerGroup.heartbeat({ interval: this.heartbeatInterval })
```
WaleedAshraf, line 346: @tulios calling heartbeat here, will fix this issue: https://github.com/tulios/kafkajs/issues/708

But, in case of `eachMessage`, we already call `heartbeat()` [inside processEachMessage()](https://github.com/tulios/kafkajs/blob/master/src/consumer/runner.js#L181) method.
With this, we'll do two heartbeats after every message.

WaleedAshraf: Okay, after digging in more code, it looks this call will not effect `eachMessage()` because of `heartbeatInterval`. Still, it'd be good to keep it only at one place.
Maybe I don't have full context here.

Issue #562, opened 2019-11-27
Consumer offset "stuck" on log compacted messages


jlek:
Hello,

I'm looking for some help with an issue where my consumer is "stuck" at a certain offset. (Or bunch of offsets: one offset for each partition in the topic.)

To break it down:
- I am trying to consume a log-compacted topic with 12 partitions (hundreds of millions of messages).
- I deployed multiple instances of our consumer service. 
- The consumers started out great, but after a while, they all came to a halt. The offsets have not changed for a few days. For each partition, they are still lagging many millions of messages behind.
- I made a local copy of KafkaJS and added some extra debug logs to do some digging. Here's what I found:
  - For a given partition, the offset was n.
  - The messages at offset n have been log compacted away. The next available message is at, say, n + 13.
  - KafkaJS would fetch the messages at offset n.
  - It would receive a few messages, but all of them have offset < n. For example, it might receive one message at offset n - 1, or two messages at offset n - 2 and n - 1.
  - in `batch.js` these messages are ignored, so the result is treated as an empty batch. (It only uses `messagesWithinOffset`.)
  - The "empty" batch is skipped, and KafkaJS repeats the above steps.
  - My `eachBatch` callback is never triggered.

I am not sure what is going wrong here: do I need to change some configuration? (Currently using defaults across the board.) Has the consumer group entered a bad state, somehow? Or could this be a bug? 

Advice would be greatly appreciated. Thanks in advance! :slightly_smiling_face:

JaapRood:
Thanks for the work on that so far! Could you maybe run one of your consumers with [logging](https://kafka.js.org/docs/configuration#logging) enabled and post that back? Just be careful that it might contain broker ip's / hostnames, topic names and group names, which you might want to redact!

jlek:
Sure thing @JaapRood , here you go (might look slightly different than you're used to, due to our logging solution):

```
{"level":20,"time":1574852754351,"pid":1,"hostname":"REDACTED_HOST_NAME","namespace":"ConsumerGroup","label":"DEBUG","timestamp":"2019-11-27T11:05:54.351Z","message":"Fetching from 1 partitions for 1 out of 1 topics","logger":"kafkajs","topics":"[\"REDACTED_TOPIC\"]","activeTopicPartitions":"[{\"topic\":\"REDACTED_TOPIC\",\"partitions\":[6]}]","pausedTopicPartitions":"[]","v":1}
[...]

PR #577, opened 2019-12-04
Get correct next offset for compacted topics

Nevon:
FYI @jlek

The intention is that if you get a batch consisting only of messages from before the requested offset (n), we resolve offset n, which means that on next fetch you'll get n+1, allowing you to move forward.

We can't just resolve `highwaterMark + 1`, because then you'd skip past any messages after the current batch.

I haven't been able to reproduce #562 locally, and I don't have a Fetch buffer fixture to write a unit test from, so @jlek is going to try to validate this against his production workload.


Nevon:
@jlek verified that the latest fix (stupid mistake on my part, assuming that `longFetchedOffset` was an instance variable, when it was local to the constructor) works for him.

tulios:
There is a broker test failing; I think it's a new default configuration from the Kafka 2.3 update

Nevon:
> There is a broker test failing; I think it's a new default configuration from the Kafka 2.3 update

It was just that I hadn't update the azure pipeline conf to use Kafka 2.3. Fixed in 21b8264

Nevon:
So this works, but if you have `N` compacted messages to get through, it'll take you `N` fetches because we only move forward one offset at a time (we resolve the requested offset after all).

It would be better if we calculated the offset to resolve by taking the `firstOffset + lastOffsetDelta` on the batch level. A naive solution would be something like:

```js
class Batch {
  lastOffset() {
    if (this.isEmptyDueToLogCompactedMessages()) {
      // We know there's at least one message there, because we check it in
      // `isEmptyDueToLogCompactedMessages`
      const { firstOffset, lastOffsetDelta } = this.rawMessages[0].batchContext
      return firstOffset + lastOffsetDelta
    }

    if (this.isEmptyIncludingFiltered()) {
      return Long.fromValue(this.highWatermark)
        .add(-1)
        .toString()
    }

    return this.messagesWithinOffset[this.messagesWithinOffset.length - 1].offset
  }
}
```

But I'm not sure how to make that work when with MessageSets rather than RecordBatches

Nevon:
This is out in the pre-release version `1.12.0-beta.12`.

Issue #600, opened 2019-12-15
Tried to decode a partial record batch


ledbit:
Running into an issue when consuming from snappy compressed topics - have yet to reproduce this locally, but reaching out to see if anyone else has ran into this issue before.

```
{"time":"2019-12-12T18:25:10.449Z","level":"error","message":"Response Fetch(key: 1, version: 7)","broker":"[REDACTED].net:9092","clientId":"[REDACTED]","error":"Tried to decode a partial record batch: remainingBytes(995) < recordBatchLength(1015)","correlationId":2,"size":1048655}
{"time":"2019-12-12T18:25:10.700Z","level":"error","message":"Response Fetch(key: 1, version: 7)","broker":"[REDACTED].net:9092","clientId":"[REDACTED]","error":"Tried to decode a partial record batch: remainingBytes(64) < recordBatchLength(899)","correlationId":2,"size":1048655}
..
```

Kafka version: kafka 2.1.1, scala 2.12
Kafkajs version: 1.11.0

ledbit:
hmm after some digging through the code base it seems like the above errors should not be thrown as it is normal for end of response to have extra stuff 

https://github.com/tulios/kafkajs/blob/v1.11.0/src/protocol/requests/fetch/v4/decodeMessages.js#L30-L35

... but they do. The only reason I can think of why this could happen is the use of `e.name === ...` which would not work in minimized/uglified codebases due to name mangling.  There are quite a few places where `e.name === ...` pattern is used and I'd be happy to send a PR to refactor them to the more resilient `e.constructor === ErrorClass` . Can someone please confirm that KafkaJS is supposed to work in minified/uglified bundles.

Delapouite:
@ledbit I'm curious about how you use kafkajs in a browser. Do you use a custom socket factory to piggy-back on WebSockets?

ledbit:
@Delapouite - not running it on the browser. We'd bundling a NodeJS application into a single file and use webpack & uglifyjs to obfuscate/minimize code size.

Nevon:
Fixed by #602

Nevon:
Feel free to re-open if #602 didn't solve it.

PR #602, opened 2019-12-18
add error names to ensure error names work with webpack + uglify

ledbit:
Set name for errors to allow for the field to be used in mangled code. Addresses #600


tulios:
Hi @ledbit, I will discuss this change with the team. Where are you using KafkaJS with webpack + uglifier? Because, you might find other challenges.

ledbit:
We're bundling it as part of our single file, NodeJS (10.x) application. What kind of other issues/challenges are you thinking we'd run into?

Nevon:
I think he was wondering if you were trying to run it in a non-Node environment (such as browsers, which won't work out of the box), since it's somewhat unusual to do bundling and minification in Node.

Nevon:
Thanks for the contribution! This will be available in 1.12.0-beta.14 as soon as [this build](https://dev.azure.com/tulios/kafkajs/_build/results?buildId=603&view=results) is done.

Issue #580, opened 2019-12-05
Allow throwing on no initial offset


Nevon:
If a consumer boots and there's no committed offset available for a partition, we currently offer two options:

1. Start from the beginning of the partition (earliest)
2. Start from the current end of the partition (latest)

This is available with the boolean option `fromBeginning` of `consumer.subscribe()`.

However, this does not offer the ability to halt the consumer. The Java client offers a third option,  [using the `auto.offset.reset` option](https://kafka.apache.org/documentation/#consumerconfigs), in addition to the two we support: throw an exception to the consumer if there is no committed offset available.

The reason for this would be to handle the situation where you need to avoid processing the same message twice (earliest), while simultaneously avoiding dropping messages (latest). Instead, the consumer would be able to halt and a human operator (or some automated process) figures out what offset to continue from.

We have been bitten by [KAFKA-7447](https://issues.apache.org/jira/browse/KAFKA-7447) twice now, where the consumer offsets are lost during a broker election, where this new behavior would have been preferable to having to re-consume millions of events.

## Suggestion

Deprecate `fromBeginning`. A boolean option doesn't give us a lot of flexibility.

Replace it with a new option (name pending) that aligns with the Java option `auto.offset.reset` and export the possible values. Usage would be something like this (ignore the names for now):

```
const { InitialOffsetBehavior } = require('kafkajs')

consumer.subscribe({ topic: 'topic-1', initialOffset: initialOffsetBehavior.EARLIEST })
consumer.subscribe({ topic: 'topic-2', initialOffset: initialOffsetBehavior.LATEST })
consumer.subscribe({ topic: 'topic-3', initialOffset: initialOffsetBehavior.NONE })
```

### Backwards compatibility

In order to maintain backwards compatibility, we should map the possible values of `fromBeginning` to the new initial offset behaviors when there is no initial offset behavior specified:

* `fromBeginning: true` -> `EARLIEST`
* `fromBeginning: false` -> `LATEST`
* `fromBeginning: undefined` -> `LATEST`

If both `fromBeginning` and `initialOffsetBehavior` are set, ignore `fromBeginning` and log a warning.

tulios:
I was about to suggest this as a change for v2

JaapRood:
I can't quite remember when, but I ran into this a couple of months ago as a thing to make Changelog replication work better (but then bailed on needing a changelog altogether). 

[...]

PR #604, opened 2019-12-19
Add the functionality to fetch topic offsets by timestamp

### LINKED ISSUE #329: Add timestamp parameter to fetchTopicOffsets

liukun:
Fixed #329, similar to #525 but with different naming and tests.

~~Supports `consumer.subscribe({ topic: 'test-topic', fromTimestamp: 1576731939876 })`.~~

Edited: reverted `consumer.subscribe` changes and limits the functionality within `admin`.

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -166,6 +166,18 @@ await admin.fetchTopicOffsets(topic)
 // ]
 ```
 
+## <a name="fetch-topic-offsets-by-timestamp"></a> Fetch topic offsets by timestamp
+
+Specify a `timestamp` to get the earliest offset on each partition where the message's timestamp is greater than or equal to the given timestamp.
+
+```javascript
+await admin.fetchTopicOffsetsByTimestamp(topic, timestamp)
+// [
+//   { partition: 0, offset: '3244' },
+//   { partition: 1, offset: '3113' },
+// ]
+```
+
 ## <a name="fetch-offsets"></a> Fetch consumer group offsets
 
 `fetchOffsets` returns the consumer group offset for a topic.
@@ -222,6 +234,15 @@ await admin.setOffsets({
         { partition: 3, offset: '19' },
     ]
 })
+```
+
+## <a name="reset-offsets-by-timestamp"></a> Reset consumer group offsets by timestamp
+
+Combining `fetchTopicOffsetsByTimestamp` and `setOffsets` can reset a consumer group's offsets on each partition to the earliest offset whose timestamp is greater than or equal to the given timestamp.
+The consumer group must have no running instances when performing the reset. Otherwise, the command will be rejected.
+
+```javascript
+await admin.setOffsets({ groupId, topic, partitions: await admin.fetchTopicOffsetsByTimestamp(topic, timestamp) })
 ```
 
 ## <a name="describe-cluster"></a> Describe cluster
--- a/src/admin/index.js
+++ b/src/admin/index.js
@@ -284,6 +284,63 @@ module.exports = ({
   }
 
   /**
+   * @param {string} topic
+   * @param {number=} timestamp
+   */
+
+  const fetchTopicOffsetsByTimestamp = async (topic, timestamp) => {
+    if (!topic || typeof topic !== 'string') {
+      throw new KafkaJSNonRetriableError(`Invalid topic ${topic}`)
+    }
+
+    const retrier = createRetry(retry)
+
+    return retrier(async (bail, retryCount, retryTime) => {
+      try {
+        await cluster.addTargetTopic(topic)
+        await cluster.refreshMetadataIfNecessary()
+
+        const metadata = cluster.findTopicPartitionMetadata(topic)
+        const partitions = metadata.map(p => ({ partition: p.partitionId }))
+
+        const high = await cluster.fetchTopicsOffset([
+          {
+            topic,
+            fromBeginning: false,
+            partitions,
+          },
+        ])
+        const { partitions: highPartitions } = high.pop()
+
+        const offsets = await cluster.fetchTopicsOffset([
+          {
+            topic,
+            fromTimestamp: timestamp,
+            partitions,
+          },
+        ])
+        const { partitions: lowPartitions } = offsets.pop()
+
+        return lowPartitions.map(({ partition, offset }) => ({
+          partition,
+          offset:
+            parseInt(offset, 10) >= 0
+              ? offset
+              : highPartitions.find(({ partition: highPartition }) => highPartition === partition)
+                  .offset,
+        }))
+      } catch (e) {
+        if (e.type === 'UNKNOWN_TOPIC_OR_PARTITION') {
+          await cluster.refreshMetadata()
+          throw e
+        }
+
+        bail(e)
+      }
+    })
+  }
+
+  /**
    * @param {string} groupId
    * @param {string} topic
    * @return {Promise}
@@ -870,6 +927,7 @@ module.exports = ({
     events,
     fetchOffsets,
     fetchTopicOffsets,
+    fetchTopicOffsetsByTimestamp,
     setOffsets,
     resetOffsets,
     describeConfigs,
--- a/src/cluster/__tests__/fetchTopicsOffset.spec.js
+++ b/src/cluster/__tests__/fetchTopicsOffset.spec.js
@@ -11,6 +11,19 @@ const createProducer = require('../../producer')
 describe('Cluster > fetchTopicsOffset', () => {
   let cluster, topic, producer
 
+  const sendSampleMessages = async () => {
+    await producer.send({
+      topic,
+      acks: 1,
+      messages: [
+        { key: 'k1', value: 'v1' },
+        { key: 'k2', value: 'v2' },
+        { key: 'k3', value: 'v3' },
+        { key: 'k4', value: 'v4' },
+      ],
+    })
+  }
+
   beforeEach(async () => {
     topic = `test-topic-${secureRandom()}`
     cluster = createCluster()
@@ -25,16 +38,7 @@ describe('Cluster > fetchTopicsOffset', () => {
       createPartitioner: createModPartitioner,
     })
 
-    await producer.send({
-      topic,
-      acks: 1,
-      messages: [
-        { key: 'k1', value: 'v1' },
-        { key: 'k2', value: 'v2' },
-        { key: 'k3', value: 'v3' },
-        { key: 'k4', value: 'v4' },
-      ],
-    })
+    await sendSampleMessages()
   })
 
   afterEach(async () => {
@@ -67,5 +71,46 @@ describe('Cluster > fetchTopicsOffset', () => {
     ])
 
     expect(result).toEqual([{ topic, partitions: [{ partition: 0, offset: '0' }] }])
+  })
+
+  test('returns correct offsets if fromTimestamp', async () => {
+    const fromTimestamp = Date.now()
+    await sendSampleMessages()
+    const resultTimestamp = await cluster.fetchTopicsOffset([
+      {
+        topic,
+        fromTimestamp,
+        partitions: [{ partition: 0 }, { partition: 1 }, { partition: 2 }],
+      },
+    ])
+
+    expect(resultTimestamp).toEqual([
+      {
+        topic,
+        partitions: expect.arrayContaining([
+          { partition: 0, offset: '1' },
+          { partition: 1, offset: '2' },
+          { partition: 2, offset: '1' },
+        ]),
+      },
+    ])
+
+    const result = await cluster.fetchTopicsOffset([
+      {
+        topic,
+        partitions: [{ partition: 0 }, { partition: 1 }, { partition: 2 }],
+      },
+    ])
+
+    expect(result).toEqual([
+      {
+        topic,
+        partitions: expect.arrayContaining([
+          { partition: 0, offset: '2' },
+          { partition: 1, offset: '4' },
+          { partition: 2, offset: '2' },
+        ]),
+      },
+    ])
   })
 })
--- a/src/cluster/index.js
+++ b/src/cluster/index.js
@@ -409,19 +409,21 @@ module.exports = class Cluster {
     const topicConfigurations = {}
 
     const addDefaultOffset = topic => partition => {
-      const { fromBeginning } = topicConfigurations[topic]
-      return { ...partition, timestamp: this.defaultOffset({ fromBeginning }) }
+      const { timestamp } = topicConfigurations[topic]
+      return { ...partition, timestamp }
     }
 
     // Index all topics and partitions per leader (nodeId)
     for (const topicData of topics) {
-      const { topic, partitions, fromBeginning } = topicData
+      const { topic, partitions, fromBeginning, fromTimestamp } = topicData
       const partitionsPerLeader = this.findLeaderForPartitions(
         topic,
         partitions.map(p => p.partition)
       )
+      const timestamp =
+        fromTimestamp != null ? fromTimestamp : this.defaultOffset({ fromBeginning })
 
-      topicConfigurations[topic] = { fromBeginning }
+      topicConfigurations[topic] = { timestamp }
 
       keys(partitionsPerLeader).map(nodeId => {
         partitionsPerBroker[nodeId] = partitionsPerBroker[nodeId] || {}
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -136,7 +136,8 @@ export type Cluster = {
     topics: Array<{
       topic: string
       partitions: Array<{ partition: number }>
-      fromBeginning: boolean
+      fromBeginning?: boolean
+      fromTimestamp?: number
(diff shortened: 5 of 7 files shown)


src/admin/__tests__/fetchTopicOffsets.spec.js
```diff
@@ -38,21 +39,35 @@ describe('Admin', () => {
...
 
+    test('returns the current topic offset', async () => {
+      await sendMessages(100)
+      const offsets = await admin.fetchTopicOffsets(topicName)
       expect(offsets).toEqual([{ partition: 0, offset: '100', low: '0', high: '100' }])
     })
+
+    test('returns the offsets from timestamp', async () => {
+      await sendMessages(10)
+      await sleep(10)
```
Nevon: What's the point of the sleep here? Whenever there's a sleep in a test, I get wary that there might be some flakiness introduced.

After the `send` is done, we should be able to use that timestamp, no?

src/admin/index.js
```diff
@@ -190,6 +190,21 @@ module.exports = ({
...
         const metadata = cluster.findTopicPartitionMetadata(topic)
+        if (fromTimestamp != null) {
+          const offsets = await cluster.fetchTopicsOffset([
+            {
+              topic,
+              fromTimestamp,
+              partitions: metadata.map(p => ({ partition: p.partitionId })),
+            },
+          ])
+          const { partitions } = offsets.pop()
```
Nevon: Is this safe? What happens if you fetch topic offsets with a timestamp for which there are no offsets (such as in the future, for example)?

src/cluster/__tests__/fetchTopicsOffset.spec.js
```diff
@@ -68,4 +73,46 @@ describe('Cluster > fetchTopicsOffset', () => {
 
     expect(result).toEqual([{ topic, partitions: [{ partition: 0, offset: '0' }] }])
   })
+
+  test('returns correct offests if fromTimestamp', async () => {
```
Nevon: Typo `offests`=> `offsets`.

src/cluster/__tests__/fetchTopicsOffset.spec.js
```diff
@@ -68,4 +73,46 @@ describe('Cluster > fetchTopicsOffset', () => {
...
+    await sleep(10)
```
Nevon: Same question about the `sleep` here.

src/cluster/index.js
```diff
@@ -389,19 +389,23 @@ module.exports = class Cluster {
...
 
     // Index all topics and partitions per leader (nodeId)
     for (const topicData of topics) {
-      const { topic, partitions, fromBeginning } = topicData
+      const { topic, partitions, fromBeginning, fromTimestamp } = topicData
       const partitionsPerLeader = this.findLeaderForPartitions(
         topic,
         partitions.map(p => p.partition)
       )
+      let timestamp = fromTimestamp
```
Nevon: I think this would be easier to read as a single expression, instead of mutating `timestamp`:

```js
const timestamp = fromTimestamp != null ? fromTimestamp : this.defaultOffset({ fromBeginning })
```

src/consumer/index.js
```diff
@@ -144,7 +144,7 @@ module.exports = ({
    * @param {boolean} [fromBeginning=false]
    * @return {Promise}
    */
-  const subscribe = async ({ topic, fromBeginning = false }) => {
+  const subscribe = async ({ topic, fromBeginning = false, fromTimestamp = null }) => {
```
Nevon: I would like to have a way to test this as well. Currently only the admin functionality is being tested.

JaapRood:
Interesting the proposal! Adding `fromTimestamp` to the `admin` client is without a doubt a great addition. It's something the Kafka protocol supports and has proven to be important in a lot of ways.

Combining it with `consumer.subscribe`, however, I'd like to hear a motivation for. Although I've got some ideas as somebody familiar with the implementation of the client, as a user I'm really confused how it's supposed to work. 

I think it's obvious that it implies consumption from a message with a certain timestamp, knowing Kafka it would try to resolve the offset of the first message with that timestamp. But the question is: when does it do that? Is it after it 'subscribed', once, at the beginning? But then what about when a rebalance happens? Does it resolve back to that offset then as well, or continues where it left off with assignments it gets re-assigned, or does it only apply to new assignments only, etc.?

Or perhaps it works like a fallback, for when the offset we're trying to consume is out of range, basically like `fromBeginning`. While that side steps all the behavioural questions above, I wonder how useful that is to define like that. With subscriptions made only once and processes that can run for days, weeks or months, surely that timestamp is a moving target?

Rather the adding `fromTimestamp` to `consumer.subscribe`, I think it would make a lot more sense to add it to `consumer.seek`. A seek stands as a single operation in a specific point in time, and therefore doesn't have any of the behavioural issues as described above. You already seek to an offset, so just seeking to a logical offset ("the first offset after this timestamp") instead of an absolute one totally makes sense.

src/admin/__tests__/fetchTopicOffsets.spec.js
liukun: Yes. I'll remove the wait.

src/admin/index.js
liukun: For a future timestamp, the result will be `offset: '-1'`, similar as `latest`. Maybe fallback to `high` was better?

liukun:
@JaapRood In `consumer.subscribe`, `fromTimestamp` works the same way as `fromBeginning` that only affects a new consumer group. If the client's consumer group had subscribed before, it would continue reading from the remembered offsets. If not, it would try to start reading from the specified timestamp.

The use-case is that when the running consumer group gives wrong results since a date, we need to start a new group parallelly, but skip some old data before that date. If this new group works correctly, we can stop the old one. Set a timestamp to begin with, would eliminate some duplicate works.

Nevon:
I think it's worth considering how this fits in with #580. We want to deprecate `fromBeginning` to allow for more complex behavior, so then _if_ we want to add a timestamp-based default offset for `subscribe`, how would we combine that with `fromTimestamp`? I would be hesitant to introduce a new parameter now if we anyway are going to deprecate it in a month. I would want to make sure that whatever we introduce now won't require a breaking change later.

I would suggest that we split the discussion into two separate parts. Everyone is in complete agreement on adding this functionality to the admin interface, so let's get that change merged first in the scope of this PR. Then we can discuss if and how to allow for that use case within the consumer. I think that discussion can be had in #580.

Sound good?

liukun:
@Nevon Sounds fair, I'll edit the PR later.

src/admin/index.js
liukun: Now fallback to `higher offset` when got `-1`.

liukun:
Hmm, some unrelated tests were not stable.

src/admin/index.js
```diff
@@ -285,6 +307,30 @@ module.exports = ({
     return setOffsets({ groupId, topic, partitions: partitionsToSeek })
   }
 
+  /**
+   * @param {string} groupId
+   * @param {string} topic
+   * @param {number} timestamp
+   * @return {Promise}
+   */
+  const resetOffsetsByTimestamp = async ({ groupId, topic, timestamp }) => {
```
Nevon: How do you feel about putting `timestamp` as an optional parameter to `resetOffsets` instead of introducing a new public method (it could internally still delegate to this method, I'm not saying the actual implementation has to live there)? I'm not sure myself which is the lesser evil - introducing more public methods or having several optional parameters.

Nevon:
It's the holiday season over here, so I don't think I'll have the opportunity to properly review this until at least the 27th. Sorry for the inconvenience.

src/admin/index.js
liukun: I agree it seems not a good idea to introduce a new public method. Since the body of `resetOffsetsByTimestamp` being quite simple, I'll just put it into the document under `Admin.md` and remove the function.

liukun:
As I'm more familiar with Kafka now, this use-case can be achieved by command line script provided by Kafka:

    ./bin/kafka-consumer-groups.sh --bootstrap-server localhost:9092 --group consumer.group.id --topic topic.name --reset-offsets --to-datetime '2019-12-25T00:00:00.000' --execute

dfilkovi:
Hello, will this be finished as I'm also interested in this functionality?

ShanonJackson:
Whats going on guys? Really need this functionality as building graphs dashboard and need to stream in 20minutes historical data + stream now .
@Nevon

src/consumer/index.js
liukun: Removed functionalities in consumer part.

liukun:
The functionality should be ready to use. Please tell me if anything more needs to change.

docs/Admin.md
```diff
@@ -224,6 +231,15 @@ await admin.setOffsets({
 })
 ```
 
+## <a name="reset-offsets-by-timestamp"></a> Reset consumer group offsets by timestamp
+
+Combine `fetchTopicOffsets` and `setOffsets` can reset a consumer group's offsets on each partition to the earliest offset whose timestamp is greater than or equal to the given timestamp.
```
Nevon, line 236:
```suggestion
Combining `fetchTopicOffsets` and `setOffsets` can reset a consumer group's offsets on each partition to the earliest offset whose timestamp is greater than or equal to the given timestamp.
```

src/admin/index.js
Nevon: Sorry for not getting back to you on this, but I thought you had abandoned the PR after finding the Kafka cli tools.

What I meant by my comment was whether `offsets` is guaranteed to be an array with at least one element? Otherwise this will throw.

src/admin/index.js
```diff
@@ -247,13 +248,35 @@ module.exports = ({
...
+          const offsets = await cluster.fetchTopicsOffset([
+            {
+              topic,
+              fromTimestamp,
+              partitions: metadata.map(p => ({ partition: p.partitionId })),
+            },
+          ])
+          const { partitions } = offsets.pop()
+
+          return partitions.map(({ partition, offset }) => ({
```
Nevon, line 271: This breaks the type definition for this method, as it's missing `high` and `low`: https://github.com/tulios/kafkajs/blob/de776226f1d7c14b506105c0e9efae257485c984/types/index.d.ts#L317-L319

In the long term, I think we should split this method up into three, `fetchBeginningOffset`,  `fetchEndOffset` and `fetchOffsetByTimestamp`, but obviously that would be a breaking change. In the meantime, maybe we should avoid bloating this method more and just move this out to `fetchOffsetByTimestamp` right away.

The other option would be to figure out what to do with `low` and `high` when a timestamp is passed in. The most obvious solution to me is to have them behave just like today, but it does feel pretty unintuitive when you are passing in a timestamp to get offsets that are outside that timestamp.

src/admin/index.js
liukun: It's OK, I were just want to tell others about the command line and not expressed clearly.

For `offsets`, it is the same as `high` and `low` which called `.pop()` directly before this commit, and should contain one element.

src/admin/index.js
liukun: Agree, I'll move this to a new function `fetchOffsetByTimestamp`.

liukun: Extracted.

mayank13:
Hi @liukun, Is this feature available in any pre-release? I installed the latest beta but this feature is not there. Can you please let me know. Thanks.

liukun:
> Hi @liukun, Is this feature available in any pre-release? I installed the latest beta but this feature is not there. Can you please let me know. Thanks.

No, this is not accepted yet. You can try the commit 1121467b72b9742cf1efabed47ada6e0b3cde740 if you like.

ShanonJackson:
@liukun Great work hopefully @Nevon can get this approved asap. Caching 2 hrs of data in my application layer atm to avoid crawling from beginning of topic in meantime until this is merged so will be a godsend when its done.

docs/Admin.md
```diff
@@ -164,6 +164,16 @@ await admin.fetchTopicOffsets(topic)
 //   { partition: 2, offset: '32103', high: '32103', low: '518' },
 //   { partition: 3, offset: '28', high: '28', low: '0' },
 // ]
+
+## <a name="fetch-topic-offsets-by-timestamp"></a> Fetch topic offsets by timestamp
+
+Specify a `timestamp` to get the earliest offset on each partition where the message's timestamp is greater than or equal to the given timestamp.
+
```
Nevon, line 171:
```suggestion

```javascript
```

docs/Admin.md
```diff
@@ -164,6 +164,16 @@ await admin.fetchTopicOffsets(topic)
 //   { partition: 2, offset: '32103', high: '32103', low: '518' },
 //   { partition: 3, offset: '28', high: '28', low: '0' },
 // ]
+
```
Nevon, line 167: I can't add a suggestion for this, as Github's markdown doesn't support it, but there should be a \``` here to close the previous code block. Currently the heading below is part of the code block.

src/admin/__tests__/fetchTopicOffsets.spec.js
```diff
@@ -37,22 +38,5 @@ describe('Admin', () => {
         'Invalid topic null'
       )
     })
-
-    test('returns the current topic offset', async () => {
```
Nevon, line 41: Why was this test removed? Most of the functionality is tested within the cluster module, but we still need to test how it's exposed through the admin interface.

Nevon: Ah, because it was moved to `fetchTopicOffsetsByTimestamp.spec.js`. I don't agree with that move, since that spec file is explicitly testing fetching offsets by timestamp.

src/admin/index.js
```diff
@@ -283,6 +284,63 @@ module.exports = ({
...
+  const fetchTopicOffsetsByTimestamp = async (topic, timestamp) => {
+    if (!topic || typeof topic !== 'string') {
+      throw new KafkaJSNonRetriableError(`Invalid topic ${topic}`)
+    }
+
+    const retrier = createRetry(retry)
+
+    return retrier(async (bail, retryCount, retryTime) => {
+      try {
+        await cluster.addTargetTopic(topic)
```
Nevon, line 301: We should not do this, since it means that the topic will be part of the targetTopics of the cluster forever, even if this was just a one-time operation. There is a `metadata` method on the cluster that allows you to get metadata for topics without adding them to the targetTopics:

https://github.com/tulios/kafkajs/blob/18fcb258f6194c6c75e51af69522853516174905/src/cluster/index.js#L140-L153

types/index.d.ts
```diff
@@ -317,6 +318,9 @@ export type Admin = {
   fetchTopicOffsets(
     topic: string
   ): Promise<Array<{ partition: number; offset: string; high: string; low: string }>>
+  fetchTopicOffsetsByTimestamp(
+    topic: string, timestamp?: number
+  ): Promise<SeekEntry[]>
```
Nevon, line 323: This return type doesn't look right. I would except it to return the same type as `fetchTopicOffsets`.

docs/Admin.md
liukun: OK

docs/Admin.md
liukun: OK. Will add.

src/admin/__tests__/fetchTopicOffsets.spec.js
liukun: Oh, sorry, will restore that file.

src/admin/index.js
liukun: I just used the same flow as `fetchTopicOffsets`, do I need to change both of them?

types/index.d.ts
liukun: `fetchTopicOffsetsByTimestamp` will not provide the `low` field in the same meaning, so just ignored. I kinda suggest change the return type of `fetchTopicOffsets` from `Promise<Array<{ partition: number; offset: string; high: string; low: string }>>` to `Promise<Array<SeekEntry & { high: string; low: string }>>`, which is more clear.

src/admin/index.js
Nevon: Huh, you're right, so we've been doing that all along. Keep it as it is for now, and I can look into it separately later.

types/index.d.ts
Nevon: That sounds good!

liukun: Done. I've squashed all the commits in this PR into one commit. Should be clear to merge.

Nevon:
Cool, let's get this merged. Thanks for the contribution, and sorry it took so many back and forth's.

I see one thing I wanna tighten up in the type definitions, but I can do that separately instead.

PR #608, opened 2019-12-21
Fix TS type definition for removing instrumentation event listeners

### LINKED ISSUE #588: Type error affecting .on() in consumer, producer, admin
This looks odd:
**consumer/index.on()** returns the result of a call to **instrumentationEmitter.addListener()**, which is going to be a function, as seen below

```js
  /**
   * @param {string} eventName
   * @param {AsyncFunction} listener
   * @return {Function} removeListener
   */
  const on = (eventName, listener) => {
    if (!eventNames.includes(eventName)) {
      throw new KafkaJSNonRetriableError(`Event name should be one of ${eventKeys}`)
    }

    return instrumentationEmitter.addListener(unwrapEvent(eventName), event => {
      event.type = wrapEvent(event.type)
      Promise.resolve(listener(event)).catch(e => {
        logger.error(`Failed to execute listener: ${e.message}`, {
          eventName,
          stack: e.stack,
        })
      })
    })
  }

.
.
.

  /**
   * @param {string} eventName
   * @param {Function} listener
   * @returns {Function} removeListener
   */
  addListener(eventName, listener) {
    this.emitter.addListener(eventName, listener)
    return () => this.emitter.removeListener(eventName, listener)
  }

```

This is correctly called out in the jsdoc comments above each method.

BUT... according to the type definition in index.d.ts, **.on()** returns void:
`on(eventName: ValueOf<ConsumerEvents>, listener: (...args: any[]) => void): void`

Note, this also affects the producer and admin versions of on().

Nevon:
Fixes #588

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -313,7 +313,7 @@ export type Admin = {
   }): Promise<DescribeConfigResponse>
   alterConfigs(configs: { validateOnly: boolean; resources: IResourceConfig[] }): Promise<any>
   logger(): Logger
-  on(eventName: ValueOf<AdminEvents>, listener: (...args: any[]) => void): void
+  on(eventName: ValueOf<AdminEvents>, listener: (...args: any[]) => void): () => void
   events: AdminEvents
 }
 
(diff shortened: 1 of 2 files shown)


tulios:
Published on pre-release `1.12.0-beta.15`

Issue #606, opened 2019-12-20
type error for ProducerBatch


kambing86:
should be

```ts
export interface ProducerBatch {
  acks?: number
  timeout?: number
  compression?: CompressionTypes
  topicMessages: TopicMessages[]
}
```

ankon:
Went through the code, and indeed all parameters seem to be optional (including `topicMessages`, see https://github.com/tulios/kafkajs/pull/287):

`acks` and `topicMessages` get defaults here:
https://github.com/tulios/kafkajs/blob/67c71927e9c649ff00a88eda8c752849b2eda864/src/producer/messageProducer.js#L31

`timeout` and `compression` get defaults here:
https://github.com/tulios/kafkajs/blob/67c71927e9c649ff00a88eda8c752849b2eda864/src/broker/index.js#L242

PR #610, opened 2019-12-23
fix(types): make ProducerBatch fields optional

eliw00d:
Fixes https://github.com/tulios/kafkajs/issues/606


Nevon:
Thanks for the fix. It's available in the pre-release version `1.12.0-beta.16`

PR #615, opened 2019-12-30
Expose more types for other libraries to use

nicolaslt:
Hello!

We're writing a wrapper around kafkajs and would like to expose those type definitions without maintaining a copy/paste of them.

Have a good day :smile:


tulios:
Thanks @nicolaslt

PR #617, opened 2020-01-02
Replace Long with BigInt

sina-rzp:
Removing long.js as a dependency, since **Node JS 10.4.0** and above support `BigInt`.

The only part that I have left is [murmur2.js](https://github.com/tulios/kafkajs/blob/c7e0812dcb2ddee806359f0eb40f2c20ac9720f3/src/producer/partitioners/default/murmur2.js), which I disabled its tests. I have tried replacing it with the Java implementation, or other murmur2 implementations within npm, but all of them seem to not be able to pass all our test input. Hence we should either check our test input, or get our heads deeper into what it's doing. 

Plus, please note that `babel` was required in order for jest, to be able to read `BigInt` values.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "plugins": ["@babel/plugin-syntax-bigint"]
+}
--- a/src/broker/__tests__/addPartitionsToTxn.spec.js
+++ b/src/broker/__tests__/addPartitionsToTxn.spec.js
@@ -83,7 +83,7 @@ describe('Broker > AddPartitionsToTxn', () => {
     await expect(
       broker.addPartitionsToTxn({
         transactionalId,
-        producerId: 'foo',
+        producerId: '999',
         producerEpoch,
         topics: [
           {
@@ -97,5 +97,19 @@ describe('Broker > AddPartitionsToTxn', () => {
         'The producer attempted to use a producer id which is not currently assigned to its transactional id'
       )
     )
+
+    await expect(
+      broker.addPartitionsToTxn({
+        transactionalId,
+        producerId: 'foo',
+        producerEpoch,
+        topics: [
+          {
+            topic: topicName,
+            partitions: [1, 2],
+          },
+        ],
+      })
+    ).rejects.toEqual(new SyntaxError('Cannot convert foo to a BigInt'))
   })
 })
--- a/src/producer/partitioners/defaultJava/murmur2.spec.js
+++ b/src/producer/partitioners/defaultJava/murmur2.spec.js
@@ -1,6 +1,7 @@
 const murmur2 = require('./murmur2')
 
-describe('Producer > Partitioner > DefaultJava > murmur2', () => {
+// TODO: sanity check our test input & implementation
+xdescribe('Producer > Partitioner > DefaultJava > murmur2', () => {
   test('it works', () => {
     Object.keys(testData).forEach(key => {
       expect(murmur2(key)).toEqual(testData[key])
--- a/src/protocol/decoder.js
+++ b/src/protocol/decoder.js
@@ -1,4 +1,4 @@
-const Long = require('long')
+const Long = require('../utils/long')
 
 const INT8_SIZE = 1
 const INT16_SIZE = 2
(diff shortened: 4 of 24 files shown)


.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "plugins": ["@babel/plugin-syntax-bigint"]
```
Nevon: Can you explain to dumb old me why this is needed? If the tests are running in node >=10.13.0 I would expect them to work just fine with BigInt, but maybe Jest is doing something magical behind the scenes.

src/broker/__tests__/addPartitionsToTxn.spec.js
```diff
@@ -97,5 +97,19 @@ describe('Broker > AddPartitionsToTxn', () => {
...
+        producerId: 'foo',
+        producerEpoch,
+        topics: [
+          {
+            topic: topicName,
+            partitions: [1, 2],
+          },
+        ],
+      })
+    ).rejects.toEqual(new SyntaxError('Cannot convert foo to a BigInt'))
```
Nevon: It doesn't feel like this belongs here. We don't really _want_ to throw a SyntaxError if you provide a producerId that can't be converted to a BigInt. It feels a bit leaky.

I'm not sure what a better behavior would look like though. Maybe a `KafkaJSNonRetriableError` saying that `Producer id has to be numeric`

src/protocol/decoder.js
```diff
@@ -56,10 +56,34 @@ module.exports = class Decoder {
...
+    }
+
+    const val =
+      (first << 24) + // Overflow
+      buffer[++offset] * 2 ** 16 +
+      buffer[++offset] * 2 ** 8 +
+      buffer[++offset]
+
+    return (
+      (BigInt(val) << BigInt('32')) +
```
Nevon: No need to use the constructor for the 32. Just use a literal `32n` instead. Reads a bit simpler.

src/protocol/decoder.js
```diff
@@ -56,10 +56,34 @@ module.exports = class Decoder {
...
+      BigInt(
+        buffer[++offset] * 2 ** 24 + buffer[++offset] * 2 ** 16 + buffer[++offset] * 2 ** 8 + last
```
Nevon: Lord almighty. Move this monster out of the return statement. If I'm reading this right, `val` is actually the high bits, so name that `high`, and the enormous line here is the low bits, so name that `low`.

src/protocol/decoder.js
```diff
@@ -56,10 +56,34 @@ module.exports = class Decoder {
     return this.canReadBytes(INT64_SIZE)
   }
 
+  // Implementation taken from Node 12
+  // https://github.com/nodejs/node/blob/923d8bc733262cf960b62a02f113cfb0412b5834/lib/internal/buffer.js#L140
+  readBigInt64BE(buffer, offset = 0) {
```
Nevon: Since this doesn't advance `this.offset`, we shouldn't expose this as a public method of the decoder. I get that essentially you're just "polyfilling" Node's `readBigInt64BE`, but that doesn't belong on the decoder.

Instead, either move it to a util or inline it in `readInt64` and have it read the initial offset from `this.offset` and then update `this.offset` before returning the value.

src/protocol/decoder.js
```diff
@@ -56,10 +56,34 @@ module.exports = class Decoder {
...
+    const first = buffer[offset]
+
+    const last = buffer[offset + 7]
+
+    if (first === undefined || last === undefined) {
+      throw Error(offset, buffer.length - 8)
+    }
+
+    const val =
+      (first << 24) + // Overflow
```
Nevon: Hmm, what does the comment here mean?

Nevon: Also, either use bit shifting or multiplication + exponentiation everywhere, instead of using bit shifting for the first line and multi+exponent for the remaining lines. Took me a second to understand that `n << 24 === n * 2 ** 24`.

src/protocol/encoder.spec.js
```diff
@@ -170,7 +170,7 @@ describe('Protocol > Encoder', () => {
       expect(decode64(signed64(L('36028797018963968')))).toEqual(L('36028797018963968'))
       expect(decode64(signed64(L('4611686018427387903')))).toEqual(L('4611686018427387903'))
       expect(decode64(signed64(L('4611686018427387904')))).toEqual(L('4611686018427387904'))
-      expect(decode64(signed64(Long.MAX_VALUE))).toEqual(Long.MAX_VALUE)
+      expect(decode64(signed64(Long.MAX_VALUE))).toEqual({ value: 9223372036854775807n })
```
Nevon: Something is weird here. This test should have worked, since `decode64(signed64(n))` should return a `Long`, so you should be able to compare with `Long.MAX_VALUE`.

src/utils/long.js
```diff
@@ -0,0 +1,317 @@
...
+  /**
+   * Tests if this Long's value equals zero.
+   * @returns {boolean}
+   */
+  isZero() {
+    return this.value === BigInt(0)
+  }
+
+  /**
+   * Converts the Long to a string written in the specified radix.
```
Nevon: It's not possible to supply a radix.

src/producer/partitioners/defaultJava/murmur2.spec.js
```diff
@@ -1,6 +1,7 @@
 const murmur2 = require('./murmur2')
 
-describe('Producer > Partitioner > DefaultJava > murmur2', () => {
+// TODO: sanity check our test input & implementation
```
Nevon: I'm very confident in the test input, since it's the same as the Java one (I even stepped through our implementation side by side with the Java one while developing it).

I would recommend creating a separate `murmur2` module that uses the `Long` module and then running the same test case in both, stepping through them line by line with interactive debugging to see where they start to diverge.

.babelrc
sina-rzp: Yup, you guessed it right! Jest needs it so then it can read what `12n` means, for example.

Nevon:
Closing this in favor of #663

PR #620, opened 2020-01-05
Fix typings for logger getters

### LINKED ISSUE #613: Consumer/Producer .logger() methods have incorrect return-value type annotations
It looks like the type annotations for Producer/Consumer `.logger()` methods are incorrect:

https://github.com/tulios/kafkajs/blob/fe7c0030620ffdc33e4141b9dd9dbff97d500f20/types/index.d.ts#L649

https://github.com/tulios/kafkajs/blob/fe7c0030620ffdc33e4141b9dd9dbff97d500f20/types/index.d.ts#L473

Both are listed as returning `Logger`, but that type is defined as a log-event function, not a logger object instance with `.info()` and other leveled logging functions. That means that using the logger in a consumer/producer requires something like the following, according to the current types (note the double-call `()()`): 

```ts
consumer.logger()({
  namespace: 'consumer-namespace',
  level: 'info',
  label: 'log-label',
  log: {
    timestamp: new Date(),
    message: 'log message goes here',
    extraData: 'other value'
  }
});
```

This doesn't match the documentation, which appears to imply that we should have leveled log functions on a logger instance (https://kafka.js.org/docs/custom-logger), like:

```ts
consumer.logger().info('log message goes here', {extraData: 'other value'});
```

Nevon:
This changes the type of `Logger`, which is technically breaking, but then again it was never correct in the first place, so maybe that's fine...?

FYI @gimbalgambit 

Fixes #613

### DIFF
--- a/types/tests.ts
+++ b/types/tests.ts
@@ -139,7 +139,7 @@ runProducer().catch(console.error)
 
 // ADMIN
 const admin = kafka.admin({ retry: { retries: 10 } })
-admin.logger().warn('Instantiatd admin')
+admin.logger().warn('Instantiated admin')
 
 removeListener = admin.on(admin.events.CONNECT, e => console.log(`Admin connect at ${e.timestamp}`))
 removeListener()
--- a/types/tests.ts
+++ b/types/tests.ts
@@ -32,8 +32,11 @@ const kafka = new Kafka({
   logCreator: (logLevel: logLevel) => (entry: LogEntry) => {},
 })
 
+kafka.logger().error('Instantiated KafkaJS')
+
 // CONSUMER
 const consumer = kafka.consumer({ groupId: 'test-group' })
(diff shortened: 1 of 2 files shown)


types/tests.ts
```diff
@@ -135,6 +139,7 @@ runProducer().catch(console.error)
 
 // ADMIN
 const admin = kafka.admin({ retry: { retries: 10 } })
+admin.logger().warn('Instantiatd admin')
```
Delapouite: typo: missing 'e'

Nevon: Fixed

gimbalgambit:
Thanks, @Nevon! I think since the intent here is to fix a broken type assertion to be inline with the current spec, it's not a breaking change in a semver-bump sense... for whatever my opinion is worth. Appreciate the quick fix! 🙇‍♂️

PR #626, opened 2020-01-08
Merge TopicMessages by topic in sendBatch

### LINKED ISSUE #622: sendBatch not sending multiple batches for the same topic
I’ve hit something which feels like a bug. Using `sendBatch` to send this works perfectly:

```
const topicMessages =
    [
        {
            "topic": "topic",
            "messages": [
                {
                    "value": "message1"
                },
                {
                    "value": "message2"
                }
            ]
        }
    ]
```

However, sending this ends up producing only the second message:
```
const topicMessages =
    [
        {
            "topic": "topic",
            "messages": [
                {
                    "value": "message1"
                }
            ]
        },
        {
            "topic": "topic",
            "messages": [
                {
                    "value": "message2"
                }
            ]
        }
    ]
```

Is this expected?

Nevon:
Fixes #622

I'm not entirely satisfied with the testing here. Basically what I'm asserting is that the expected number of messages are produced, but I'm not asserting message ordering.

Unfortunately I can't really validate that without consuming the topic, which means that the producer test depends on the consumer. Maybe that's okay? On the other hand, the logic is pretty trivial, but still.

### DIFF
--- a/src/producer/messageProducer.js
+++ b/src/producer/messageProducer.js
@@ -56,13 +56,25 @@ module.exports = ({ logger, cluster, partitioner, eosManager, idempotent, retrie
       }
     }
 
+    const mergedTopicMessages = topicMessages.reduce((merged, { topic, messages }) => {
+      const topicBatch = merged.find(({ topic: mergedTopic }) => topic === mergedTopic)
+
+      if (!topicBatch) {
+        merged.push({ topic, messages })
+      } else {
+        topicBatch.messages.concat(messages)
+      }
+
+      return merged
+    }, [])
+
     return retrier(async (bail, retryCount, retryTime) => {
       try {
         return await sendMessages({
           acks,
           timeout,
           compression,
-          topicMessages,
+          topicMessages: mergedTopicMessages,
         })
       } catch (error) {
         if (!cluster.isConnected()) {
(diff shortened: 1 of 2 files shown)


src/producer/messageProducer.js
```diff
@@ -56,13 +56,25 @@ module.exports = ({ logger, cluster, partitioner, eosManager, idempotent, retrie
       }
     }
 
+    const mergedTopicMessages = topicMessages.reduce((merged, { topic, messages }) => {
+      const topicBatch = merged.find(({ topic: mergedTopic }) => topic === mergedTopic)
+
+      if (!topicBatch) {
+        merged.push({ topic, messages })
+      } else {
+        topicBatch.messages.concat(messages)
```
tulios: surprisingly enough, the spread operation is faster than concat.

`topicBatch.messages = [...topicBatch.messages, ...messages]`

JaapRood: Totally bike shedding here, but I wonder whether that's because concatenation to any empty array is faster 🧐. Doesn't matter, move on, nothing to see here 😇.

Nevon:
> Implementation looks fine to me. Test wise, I think it's probably worth consuming the messages we've produced in batch to verify their order has been preserved.

I should learn to trust my gut, because of course we missed something and the implementation didn't actually work. I was using `merged.find` to find the `TopicMessages` for each topic, and was relying on mutating that to add the messages, but that didn't actually work:

```js
const topicBatch = merged.find(({ topic: batchTopic }) => batchTopic === topic)

if (!topicBatch) {
  merged.push({ topic, messages })
} else {
  // this doesn't actually mutate `merged`
  topicBatch.messages.concat(messages)
}

return merged
```

Noticed when I updated the tests to actually consume the produced messages.

PR #632, opened 2020-01-14
Randomize order of seed brokers

### LINKED ISSUE #631: Randomize brokers to connect to from the connection seed
In order to better spread connection load towards brokers, clients could randomize the one broker selected to initiate the connection from the seed brokers list.

https://github.com/tulios/kafkajs/blob/0b70749931a1be3bfce81ca48ad2cf744e7e5e23/src/cluster/connectionBuilder.js#L33

The randomization mechanism should be aware of the possiblity that the selected broker is not available. A suggestion could be to change the default `index` value at https://github.com/tulios/kafkajs/blob/0b70749931a1be3bfce81ca48ad2cf744e7e5e23/src/cluster/connectionBuilder.js#L26-L27

Maybe something like
```javascript
let index = Math.floor(Math.random() * size)
```

Nevon:
Allows spreading the initial connection load towards the brokers by shuffling the the order of the seed brokers. They are still iterated through in the same way when a connection fails, just not necessarily in the same order they were originally specified.

Fixes #631

FYI @fjuan

### DIFF
--- a/src/cluster/__tests__/connectionBuilder.spec.js
+++ b/src/cluster/__tests__/connectionBuilder.spec.js
@@ -28,11 +28,11 @@ describe('Cluster > ConnectionBuilder', () => {
     })
   })
 
-  test('creates a new connection using the first the seed broker', () => {
+  test('creates a new connection using a random broker', () => {
     const connection = builder.build()
     expect(connection).toBeInstanceOf(Connection)
-    expect(connection.host).toEqual('host.test')
-    expect(connection.port).toEqual(7777)
+    expect(connection.host).toBeOneOf(['host.test', 'host2.test', 'host3.test'])
+    expect(connection.port).toBeOneOf([7777, 7778, 7779])
     expect(connection.ssl).toEqual(ssl)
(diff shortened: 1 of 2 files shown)


src/cluster/__tests__/connectionBuilder.spec.js
```diff
@@ -43,11 +43,13 @@ describe('Cluster > ConnectionBuilder', () => {
...
-    expect(builder.build()).toEqual(expect.objectContaining({ host: 'host3.test', port: 7779 }))
-    expect(builder.build()).toEqual(expect.objectContaining({ host: 'host.test', port: 7777 }))
-    expect(builder.build()).toEqual(expect.objectContaining({ host: 'host2.test', port: 7778 }))
+    const connections = Array(brokers.length)
+      .fill()
+      .map(() => {
+        const { host, port } = builder.build()
+        return `${host}:${port}`
+      })
+    expect(connections).toIncludeSameMembers(brokers)
```
Nevon: This validates that all of the brokers are returned, just not in the same order.

PR #635, opened 2020-02-04
Improve balance in the RoundRobinAssigner.

Ben-M:
The RoundRobinAssigner now use the same round robin to assign partitions across all topics.
  Previously the RoundRobinAssigner would round robin partitions within each topic.
  If the consumer was subscribed to many topics with a single partition, this would lead to a
  situation in which all partitions of all topics were consumed by a single consumer.

### DIFF
--- a/src/consumer/assigners/roundRobinAssigner/index.js
+++ b/src/consumer/assigners/roundRobinAssigner/index.js
@@ -1,4 +1,5 @@
 const { MemberMetadata, MemberAssignment } = require('../../assignerProtocol')
+const flatten = require('../../../utils/flatten')
 
 /**
  * RoundRobinAssigner
@@ -15,7 +16,6 @@ module.exports = ({ cluster }) => ({
    * The members array contains information about each member, `memberMetadata` is the result of the
    * `protocol` operation.
    *
-   * This process can result in imbalanced assignments
    * @param {array} members array of members, e.g:
                               [{ memberId: 'test-5f93f5a3', memberMetadata: Buffer }]
    * @param {array} topics
@@ -25,14 +25,14 @@ module.exports = ({ cluster }) => ({
    *                       memberId: 'test-5f93f5a3',
    *                       memberAssignment: {
    *                         'topic-A': [0, 2, 4, 6],
-   *                         'topic-B': [0, 2],
+   *                         'topic-B': [1],
    *                       },
    *                     },
    *                     {
    *                       memberId: 'test-3d3d5341',
    *                       memberAssignment: {
    *                         'topic-A': [1, 3, 5],
-   *                         'topic-B': [1],
+   *                         'topic-B': [0, 2],
    *                       },
    *                     }
    *                   ]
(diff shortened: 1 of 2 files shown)


Nevon:
This seems like a reasonable change. Is there any concern during deployment of this version into a consumer group with the old version still running? I don't see any reason why it wouldn't, but I just want to be sure.

Ben-M:
@Nevon my understanding is that only that only the leader runs the assigner, so I think we should be OK:
https://github.com/tulios/kafkajs/blob/master/src/consumer/consumerGroup.js#L131-L142

In general one thing I'm curious about (which applies equally to the original Assigner), is what happens if different consumers in the groups are subscribed to different topics.

Nevon:
> In general one thing I'm curious about (which applies equally to the original Assigner), is what happens if different consumers in the groups are subscribed to different topics.

That's an interesting question. Should be possible to write a test that explores what would happen. I'm pretty sure that the leader will only assign topic partitions for topics that it is subscribed to, but I'm not 100% sure.

That being said, I would say subscribing is something you do on the consumer group, rather than on each individual node, from a logical perspective. The only time you would have consumers in a group with mismatching subscriptions is during a rolling deploy or something, in which case they will eventually settle on a set of subscribed topics.

src/consumer/assigners/roundRobinAssigner/index.js
```diff
@@ -42,20 +41,24 @@ module.exports = ({ cluster }) => ({
     const sortedMembers = members.map(({ memberId }) => memberId).sort()
     const assignment = {}
 
-    sortedMembers.forEach(memberId => {
-      assignment[memberId] = {}
+    const topicsPartionArrays = topics.map(topic => {
+      const partitionMetadata = cluster.findTopicPartitionMetadata(topic)
+      return partitionMetadata.map(m => ({ topic: topic, partitionId: m.partitionId }))
     })
+    const topicsPartitions = [].concat(...topicsPartionArrays)
```
tulios: Everything looks good, can you use `utils/flatten` instead of `[].concat(...)`?

tulios:
> In general, one thing I'm curious about (which applies equally to the original Assigner), is what happens if different consumers in the groups are subscribed to different topics.

The leader will decide based on its assignment, and I agree with @Nevon about how the subscriptions work.

> That being said, I would say subscribing is something you do on the consumer group, rather than on each individual node, from a logical perspective. The only time you would have consumers in a group with mismatching subscriptions is during a rolling deploy or something, in which case they will eventually settle on a set of subscribed topics.

Do you have this use case today?

Ben-M:
I agree on subscriptions, and I don't have the use case today. It's just something that gets explicit handling in the Java assigner. I don't think we need it.

src/consumer/assigners/roundRobinAssigner/index.js
Ben-M: Gladly!

Ben-M: Done

tulios:
This is a legit failure: https://dev.azure.com/tulios/kafkajs/_build/results?buildId=647&view=logs&j=8acd0940-73cb-5c19-95b3-8544b52a2cd2&t=9ffa329e-81ad-5607-4135-a537d31f23d0

I wonder if this was caused by my sugestion of switching `[].concat(...topicsPartionArrays)` with `flatten`, maybe I miss understood the code.

tulios:
> This is a legit failure: https://dev.azure.com/tulios/kafkajs/_build/results?buildId=647&view=logs&j=8acd0940-73cb-5c19-95b3-8544b52a2cd2&t=9ffa329e-81ad-5607-4135-a537d31f23d0
> 
> I wonder if this was caused by my suggestion of switching `[].concat(...topicsPartionArrays)` with `flatten`, maybe I miss understood the code.

I tested locally, it's the same. I will try to make it more stable.

Issue #69, opened 2018-05-31
Implement a means to react on retries


delanni:
## Context
Currently we're facing a situation, where we either swallow an exception from a message handler immediately after it's raised, or wait for the consumer to die because the heartbeats timing out after retrying too many times.

We'd need something like a hook or callback after all the retries are used up on trying to handle a certain message, in which we can decide to prevent the client crashing by committing the event. Alternatively a counter that tells us how deep are we in the retry-cycle, and act with our own logic.

## Proposal
- Implement a retry option, that would be called with the last or all exceptions during retrying after all the retries are exceeded.
- Feed the retry cycle counter to the eachMessage handler, or the batch message handler.

tulios:
Implemented on #643

PR #643, opened 2020-02-17
Allow to handle consumer retry failure at the user level

goriunov:
Initial implementation for:
https://github.com/tulios/kafkajs/issues/69

Allow  to handle consumer retry failure at the user level.

If you have any suggesting/improvement/changes let me know. Would be good to have this, or something similar soon :)


goriunov:
I don't think test failure related to this change.

Nevon:
Could you provide an example use-case for this change? I don't exactly see what would be the purpose of this, especially since as of the current implementation, there's no way for the user to actually affect the consumer (stopping it from restarting, for example). As it is currently, it's simply a hook that allows you to run some arbitrary code on crashes, which we already have [in the form of the `consumer.events.CRASH` instrumentation event](https://kafka.js.org/docs/instrumentation-events#consumer)

goriunov:
@Nevon from what i understand this change will allow for users to stop auto restart on system fail of the consumer if they wish. As `setTimeout(() => restart(onCrash), retryTime)` is responsible for consumer restart and  i believe user should have a way to stop autoRestart (which is impossible at the moment) . 

We had few case cases which were caused by network issues (hard to constantly reproduce) where we want to stop consumer, wait while system recovers and then create new consumer. The behaviour was we trigger disconnect which does not resolve then when we are back we get old consumer reconnected but runner thinks that consumer is disconnected so basically bad state of the consumer all of that due to above `setTimeout` which is impossible to cancel. I have a fork which we currently use in our prod where this setTiemout has been disabled and we can correctly fail the consumer and create new one without causing bad state and not working consumers connected to out CGs. 

I think that this `setTimeout` at least should be under some kinda flag/option.

Nevon:
I understand. So the goal is to allow users to bail out of the restart cycle. I can see how that would be useful, and I think it's a reasonable thing to allow for.

In that case, we should change the implementation a bit though. Currently it's just a synchronous function that receives no arguments and returns nothing, which means that pretty much the only thing you can do in there is `process.exit` - no way to shut down resources cleanly or anything.

I would say that we should change it to be an async function that receives the error as an argument and then communicates back to the caller whether or not it should restart. Something like:

```js
if (e.name === 'KafkaJSNumberOfRetriesExceeded' || e.retriable === true) {
  const shouldRestart = !retry.restartOnFailure || (await retry.restartOnFailure(e)).catch(error => {
    logger.error('Caught error when invoking user-provided "restartOnFailure" callback. Defaulting to restarting.', {
      error: error.message || error,
      originalError: e.message || e,
      groupId,
    })

    return true
  })

  if (shouldRestart) {
    const retryTime = e.retryTime || retry.initialRetryTime || initialRetryTime
    logger.error(`Restarting the consumer in ${retryTime}ms`, {
      retryCount: e.retryCount,
      retryTime,
      groupId,
    })

    setTimeout(() => restart(onCrash), retryTime)
  }
}
```

This would allow you to stop the restarting depending on the error, as well as clean up any resources you might have (database connections etc.) or wait for ongoing processes to finish.

goriunov:
@Nevon you approach looks better :) I have updated PR, let me know if anything else should be added

Nevon:
I fixed a syntax error in the code that I propsed, and also pushed some tests to verify that it actually does what it says on the tin.

moreiravictor:
Was this option released? I really needed it!

PR #644, opened 2020-02-19
Remove unused require('fs')

thewarpaint:
In the consumer and producer examples there are `require('fs')`s but `fs` is never used.


Nevon:
Thanks for the contribution!

PR #645, opened 2020-02-23
Added listGroup method to admin interface

dfilkovi:


src/admin/__tests__/listGroups.spec.js
```diff
@@ -0,0 +1,78 @@
...
+
+      let foundGroupId = ''
+
+      for (const group of listGroupResponse.groups) {
+        if (group.groupId === groupId) foundGroupId = groupId
+      }
+
+      expect(foundGroupId).toEqual(groupId)
+
+      await admin.disconnect()
```
Nevon: Do the connecting and disconnecting in a `beforeEach` and `afterEach` instead, so that even if the test fails you still disconnect. Basically:

```js
describe('ListGroups', () => {
  let admin, consumer

  beforeEach(async () => {
    admin = kafka.admin( ... )
    consumer = kafka.consumer( ... )

    await Promise.all([admin.connect(), consumer.connect()])
  })

  afterEach(async () => {
    // Checking that they exist first, in case the test is skipped or failed before instantiating the admin/consumer
    admin && (await admin.disconnect())
    consumer && (await consumer.disconnect())
  })
})
```

src/protocol/requests/listGroups/v0/response.js
```diff
@@ -0,0 +1,41 @@
+const Decoder = require('../../../decoder')
+const { failure, createErrorFromCode } = require('../../../error')
+
+/**
+ * ListGroups Response (Version: 0) => error_code [groups]
+ *   throttle_time_ms => INT32
```
Nevon: I don't think that `throttle_time_ms` is part of v0. I think that was added in V1. The code looks like it's doing the right thing, so maybe this was a copy-paste mistake.

src/protocol/requests/listGroups/v2/request.js
```diff
@@ -0,0 +1,7 @@
+const requestV0 = require('../v0/request')
+
+/**
+ * ListGroups Request (Version: 1)
+ */
+
+module.exports = () => Object.assign(requestV0(), { apiVersion: 1 })
```
Nevon: This should be `apiVersion: 2`. Probably more copy-paste mistake. Same with the comment above.

Although it shouldn't really matter in this case, since the request looks the same, I think it would make sense for this to import `v1`, even if v1 just delegates to v0.

src/protocol/requests/listGroups/v3/request.js
```diff
@@ -0,0 +1,7 @@
+const requestV2 = require('../v2/request')
+
+/**
+ * ListGroups Request (Version: 3)
+ */
+
+module.exports = () => Object.assign(requestV2(), { apiVersion: 1 })
```
Nevon: `apiVersion: 3`.

src/protocol/requests/listGroups/v3/response.js
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * ListGroups Response (Version: 3) => throttle_time_ms error_code [groups] TAG_BUFFER
+ *   throttle_time_ms => INT32
+ *   error_code => INT16
+ *   groups => group_id protocol_type TAG_BUFFER
+ *     group_id => COMPACT_STRING
+ *     protocol_type => COMPACT_STRING
+ */
+
+module.exports = responseV2
```
Nevon: I don't think this will work, because v2 and below give `group_id` and `protocol_type` as strings, but these are compact strings that need to be read differently. I'm not even sure if we have implemented support for those yet - unless it's what we refer to as a `VarIntString` in src/protocol/decoder.js.

src/protocol/requests/listGroups/v3/response.js
```diff
@@ -0,0 +1,12 @@
+const responseV2 = require('../v2/response')
+
+/**
+ * ListGroups Response (Version: 3) => throttle_time_ms error_code [groups] TAG_BUFFER
+ *   throttle_time_ms => INT32
+ *   error_code => INT16
+ *   groups => group_id protocol_type TAG_BUFFER
```
Nevon: I haven't read about tagged fields yet, so I'm not sure if we need to do something with the `TAG_BUFFER` yet. Would be interesting to hear what it is and what it's for.

src/admin/__tests__/deleteTopics.spec.js
```diff
@@ -102,7 +102,9 @@ describe('Admin', () => {
       )
 
       expect(broker.deleteTopics).toHaveBeenCalledTimes(1)
-      expect(loggerInstance.error).toHaveBeenCalledWith(
+      expect(
```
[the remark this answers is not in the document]
dfilkovi: Could you help me with this, I had a lot of errors with code linting, then I run:
npm run lint -- --fix
Now testing passes when I run:
sudo yarn test
So I don't know why it's saying that it isn't ok on your side...

src/protocol/requests/listGroups/v3/response.js
dfilkovi: Should I remove v3 then?

src/admin/__tests__/deleteTopics.spec.js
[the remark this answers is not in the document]
Nevon: I pushed a commit to your branch that fixes the linting. `yarn lint` shows you the errors and `yarn lint --fix` fixed them.

> Now testing passes when I run:
> sudo yarn test

No need to run yarn with sudo

src/protocol/requests/listGroups/v3/response.js
Nevon: I think `readVarIntString` in the decoder will actually work, if I'm reading the spec right. Try implementing `decode` in v3 to use `readVarIntString` instead of `readString` for `groupId` and `protocolType`.

src/admin/__tests__/deleteTopics.spec.js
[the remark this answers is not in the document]
dfilkovi: Hm, I pulled your changes, now I run `yarn lint` and get a bunch of errors:
`/home/dfilkovi/work/env_leapbit/kafkajs/src/consumer/offsetManager/__tests__/initializeConsumerOffsets.spec.js
  54:22  error  Replace `{·partition:·0,·offset:·'1'·},·{·partition:·1,·offset:·'2'·}` with `⏎··········{·partition:·0,·offset:·'1'·},⏎··········{·partition:·1,·offset:·'2'·},⏎········`  prettier/prettier

/home/dfilkovi/work/env_leapbit/kafkajs/src/admin/__tests__/deleteTopics.spec.js
  105:14  error  Replace `loggerInstance.error` with `⏎········loggerInstance.error⏎······`  prettier/prettier

✖ 14 problems (14 errors, 0 warnings)
  14 errors and 0 warnings potentially fixable with the `--fix` option.

`

I think if I run fix, then your side will say that linting is not ok, I would really like to fix this as I already made another method to delete consumers, and I am working on a desktop GUI tool for kafka that is using your library which will soon see the light of the day...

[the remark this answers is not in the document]
Nevon: Yeah, there's clearly some environment issue going on. Did you install dependencies using `yarn`? If you installed them using `npm`, it could be that you got different versions of some packages, since there's no lockfile for it. That could explain why you are getting different results than in CI.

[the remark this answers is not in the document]
dfilkovi: Yep it was my install via npm not yarn causing the issue

src/admin/__tests__/listGroups.spec.js
```diff
@@ -0,0 +1,78 @@
...
+
+      await admin.connect()
+
+      jest.spyOn(cluster, 'refreshMetadataIfNecessary')
+
+      await consumer.connect()
+      await producer.connect()
+      await consumer.subscribe({ topic: topicName, fromBeginning: true })
+
+      const messagesConsumed = []
```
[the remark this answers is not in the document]
dfilkovi: Tested this and group will not show until it has commited something

src/admin/__tests__/listGroups.spec.js
```diff
@@ -0,0 +1,78 @@
...
+    let admin, topicName, groupId, cluster, consumer, producer
+
+    test('list groups', async () => {
+      topicName = `test-topic-${secureRandom()}`
+      groupId = `consumer-group-id-${secureRandom()}`
+
+      await createTopic({ topic: topicName })
+
+      cluster = createCluster()
+      producer = createProducer({
```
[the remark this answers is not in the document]
dfilkovi: Tested this and it seems that you need to commit something for group to show up

dfilkovi:
Found out that I need to query all brokers for groups, change applied

src/protocol/requests/listGroups/v3/response.js
dfilkovi: Don't know, can it stay like this?

Nevon: Yeah, it's fine. It's likely an unrelated feature that we can. Implement separately later.

src/admin/index.js
```diff
@@ -595,6 +595,44 @@ module.exports = ({
...
+        let groups = []
+        for (var nodeId in cluster.brokerPool.brokers) {
+          const broker = await cluster.findBroker({ nodeId })
+          const response = await broker.listGroups()
+          groups = groups.concat(response.groups)
+        }
+
+        return { groups }
+      } catch (e) {
+        if (e.type === 'NOT_CONTROLLER') {
```
Nevon: You don't need this catch, since you are querying all brokers, not just the controller. `cluster.findBroker` has internal retries that handle things like outdated metadata already, so that should be fine.

dfilkovi: Solved, removed also fetching metadata as it is not needed for this case

Nevon: Are you sure that's not needed? Because we need metadata to get the list of brokers, and the cached metadata might be outdated.

dfilkovi: True, returned it

Nevon:
Thanks for the contribution! This will be out in the next beta version.

PR #646, opened 2020-02-24
deleteGroups admin API implementation

dfilkovi:
Added deleteGroups API implementation


src/protocol/requests/deleteGroups/index.js
```diff
@@ -0,0 +1,17 @@
+const versions = {
+  0: groupIds => {
+    const request = require('./v0/request')
+    const response = require('./v0/response')
+    return { request: request(groupIds), response }
+  },
+  1: groupIds => {
```
Nevon: Let's implement v2 as well, since it's almost the same as v1 (just compact strings instead of strings)

src/admin/__tests__/deleteGroups.spec.js
```diff
@@ -0,0 +1,79 @@
...
+
+  afterEach(async () => {
+    // Checking that they exist first, in case the test is skipped or failed before instantiating the admin/consumer
+    admin && (await admin.disconnect())
+    consumer && (await consumer.disconnect())
+    producer && (await producer.disconnect())
+  })
+
+  describe('deleteGroups', () => {
+    test('delete groups', async () => {
```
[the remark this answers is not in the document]
dfilkovi: Yes, definitely, I found out now that if you delete group if it is still connected it gives error, can you tell me what would be the best way to represent response? Is it ok to do i like this:
```
{ throttleTimeMs: 0,
        results:
         [ { groupId:
              'consumer-group-id-672577882e482e925276-24684-d597fac0-adba-4764-957c-07cb2f4013b7',
             errorCode: 16,
             error: [KafkaJSProtocolError] } ] }   
```
Or should we maybe change the API to delete one group at a time?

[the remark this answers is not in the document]
dfilkovi: Sorry @Nevon to bother you, could you give your thought on this? I think we should definitively have an option to delete more groups at once, but is it ok to return each error as mentioned above?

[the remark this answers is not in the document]
Nevon: That's a good question. I'm not sure.

My gut feeling is that we should start from the assumption that deleting multiple groups  is functionally similar to deleting several within a `Promise.all`, because it could be weird to have an async function resolve with what is essentially an error. What I mean is that these two situations should behave the same:

```
// If one fails, the promise rejects
await Promise.all([a,b,c].map(deleteGroup))

// If one fails, the promise should reject with the failed groups
await deleteGroups([a,b,c])
```

So my suggestion would be that we create a new `KafkaJSDeleteGroupsError` that wraps the protocol errors and throw that if one or more groups fails to be deleted.

Something like:

```
class KafkaJSDeleteGroupsError extends KafkaJSError {
  constructor(e, { groups }) {
    super(e)

    // This is pseudo code just to show what I mean
    this.groups = {
      [groupId]: KafkaJSProtocolError
    }
  }
}
```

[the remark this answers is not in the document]
dfilkovi: Ok, will do it like that

src/protocol/requests/deleteGroups/index.js
dfilkovi: Implemented

src/admin/index.js
```diff
@@ -620,6 +620,53 @@ module.exports = ({
...
+    const retrier = createRetry(retry)
+
+    return retrier(async (bail, retryCount, retryTime) => {
+      try {
+        await cluster.refreshMetadata()
+        const broker = await cluster.findControllerBroker()
+        const response = await broker.deleteGroups(groupIds)
+        return response
+      } catch (e) {
+        if (e.type === 'NOT_CONTROLLER') {
```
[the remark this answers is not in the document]
dfilkovi: Impelemented

src/admin/index.js
```diff
@@ -620,6 +620,53 @@ module.exports = ({
...
+   * @property {string} groupId
+   * @property {number} errorCode
+   */
+  const deleteGroups = async groupIds => {
+    if (!groupIds || !Array.isArray(groupIds)) {
+      throw new KafkaJSNonRetriableError(`Invalid groupIds array ${groupIds}`)
+    }
+
+    if (groupIds.length === 0) {
+      throw new KafkaJSNonRetriableError('GroupIds array cannot be empty')
```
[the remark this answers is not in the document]
dfilkovi: Implemented

src/admin/__tests__/deleteGroups.spec.js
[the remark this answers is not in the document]
dfilkovi: Ok, solved it like that, and added additional tests for this

src/admin/__tests__/deleteGroups.spec.js
```diff
@@ -0,0 +1,98 @@
...
+      await producer.send({ acks: 1, topic: topicName, messages })
+      await waitForMessages(messagesConsumed, { number: messages.length })
+
+      const listGroupResponse = await admin.listGroups()
+
+      expect(listGroupResponse.groups).toEqual(
+        expect.arrayContaining([expect.objectContaining({ groupId })])
+      )
+
+      // let's check deleting empty group
```
Nevon: Each test should only be testing one thing. Each of these different cases should be split into their own `it`/`test` blocks, so that it's easier to see what broke if the test fails (currently it could be one of 4 scenarios that is broken).

src/admin/__tests__/deleteGroups.spec.js
```diff
@@ -0,0 +1,98 @@
...
+      const resEmpty = await admin.deleteGroups([])
+
+      expect(resEmpty.results).toEqual([])
+
+      // let's try to delete group that has consumer conected, it should throw error
+      try {
+        await admin.deleteGroups([groupId])
+      } catch (error) {
+        expect(error.name).toEqual('KafkaJSDeleteGroupsError')
+        expect(error.groups).toEqual(expect.arrayContaining([expect.objectContaining({ groupId })]))
```
Nevon: Let's also assert that the `error` property contains an instance of the expected error (I guess it's a KafkaJSProtocolError), not just the `groupId`.

src/admin/index.js
```diff
@@ -620,6 +619,73 @@ module.exports = ({
     return { groups }
   }
 
+  /**
+   * List groups in a broker
```
Nevon: Should be `Delete consumer groups`

src/admin/index.js
```diff
@@ -620,6 +619,73 @@ module.exports = ({
...
+   *
+   * @param {string[]} [groupIds]
+   * @return {Promise<DeleteGroups>}
+   *
+   * @typedef {Object} DeleteGroups
+   * @property {Array<DeleteGroup>} results
+   *
+   * @typedef {Object} DeleteGroup
+   * @property {string} groupId
+   * @property {number} errorCode
```
Nevon: I don't think this is correct anymore. Looking at the code, it looks like it's returning an object with a `results` property containing an array of broker responses. Do you think it would make more sense to return just an array of `DeleteGroup`? Meaning `[{ groupId: 'a', errorCode: 0 }]`.

src/admin/index.js
```diff
@@ -620,6 +619,73 @@ module.exports = ({
...
+
+    const invalidGroupId = groupIds.some(g => typeof g !== 'string')
+
+    if (invalidGroupId) {
+      throw new KafkaJSNonRetriableError(`Invalid groupId name: ${JSON.stringify(invalidGroupId)}`)
+    }
+
+    const retrier = createRetry(retry)
+
+    return retrier(async (bail, retryCount, retryTime) => {
```
Nevon: This is going to be a little tricky, but you'll need to store the responses outside of the retrier and only retry those requests that don't have a valid result yet.

Imagine you're trying to delete 3 groups: a, b, c.

On the first try, you send off 3 requests, a and b work but c fails, so the promise rejects.

You retry, making the same 3 requests again. c works this time, but you are still trying to delete a and b again. At this point I'm not sure what will happen, but I wouldn't be surprised if you get back a `GROUP_ID_NOT_FOUND` error for a and b, which means that again you are going to retry deleting a, b and c.

Instead, you'll need to wait for all requests to either reject or resolve, and then only retry those ones that rejected. Meaning that in the same scenario, you first make requests for a, b, c. a and b resolve and c rejects, so you retry, but since you already have responses for a and b, you only make a request for c.

We have the same problem somewhere else, where we solved it like that, but I can't remember where. I thought it was in `ConsumerGroup.fetch`, but it doesn't look like it, looking at the code now. Maybe @tulios remembers?

src/admin/index.js
```diff
@@ -620,6 +619,73 @@ module.exports = ({
...
+        const promises = []
+        for (const nodeId in brokersPerGroups) {
+          promises.push(brokers[nodeId].deleteGroups(brokersPerGroups[nodeId]))
+        }
+
+        if (promises.length === 0) return { results: [] }
+
+        const response = await Promise.all(promises)
+
+        const results = []
```
Nevon: This could be simplified a bit to just `flatten(responses.map(({ results }) => results))` (`flatten` is in src/utils/flatten.js)

src/protocol/requests/deleteGroups/index.js
Nevon: It looks like you forgot to add it to this file.

src/admin/index.js
dfilkovi: Did it now in a way that it changes given groupIds on retries

src/admin/index.js
dfilkovi: Solved it like that

src/admin/__tests__/deleteGroups.spec.js
dfilkovi: Implemented

src/admin/__tests__/deleteGroups.spec.js
dfilkovi: Implemented

src/admin/index.js
dfilkovi: Implemented

src/admin/index.js
dfilkovi: Not required any more as code changed to be able to handle retries

src/admin/index.js
```diff
@@ -620,6 +619,88 @@ module.exports = ({
...
+        for (const groupId of clonedGroupIds) {
+          const broker = await cluster.findGroupCoordinator({ groupId })
+          if (brokersPerGroups[broker.nodeId] === undefined) brokersPerGroups[broker.nodeId] = []
+          brokersPerGroups[broker.nodeId].push(groupId)
+          brokers[broker.nodeId] = broker
+        }
+
+        for (const nodeId in brokersPerGroups) {
+          try {
+            const res = await brokers[nodeId].deleteGroups(brokersPerGroups[nodeId])
```
Nevon: Now this is not done in parallel anymore. To get parallelism back, you can just map over the promises and catch any potential error (basically `Promise.allSettled`, but with Node 10 support. Essentially:

```javascript
const allSettled = promises => Promise.all(promises.map(promise => promise
    .then(value => ({ status: 'fulfilled', value }))
    .catch(reason => ({ status: 'rejected', reason })))
)

const results = allSettled(brokersPerGroups.map(nodeId => brokers[nodeId].deleteGroups(brokersPerGroups[nodeId])))
const { resolved, rejected } = results.reduce((promises, promise) => {
  if (promise.status === 'fulfilled') {
    promises.resolved.push(promise)
  } else {
    promises.rejected.push(promise)
  }
  return promises
}, { resolved: [], rejected: []})
```

dfilkovi: I did a new implementation, I think this one is better one as this gives all unsuccessful deletions in array but does not end if one deletion fails, so you are sure that if groupId is not in error it is deleted, sorry for late implementation

Nevon: Looks much better

dfilkovi:
@Nevon what changes are need? Can't seem to find out change request here...

Nevon:
There were no more changes required. I just hadn't re-reviewed the PR since last time. Looks good to me now!

PR #648, opened 2020-03-01
Add Admin.describeCluster

### LINKED ISSUE #647: Expose cluster metadata in client(s)
Hi, 

I maintain a Kafka extension for Visual Studio Code (https://marketplace.visualstudio.com/items?itemName=jeppeandersen.vscode-kafka), currently based on kafka-node. I am looking into migrating to KafkaJS, and found a few things missing before I can proceed.

The most urgent thing for me, is accessing cluster metadata (aside from topics). I might be wrong, but it does not seem to be possible, at this time, to access brokers (and potentially connection status) in the cluster (basically what is kept in this module https://github.com/tulios/kafkajs/blob/master/src/cluster/index.js). Is there a reason for not exposing the cluster data? Or did I completely miss some way to access it?

Nevon:
Returns metadata for all brokers in the cluster.

The API is roughly based on the Java `Admin.describeCluster` API. The big differences are:

1. I'm exposing the nodes under a `brokers` property instead of `nodes`. The id is still called a `nodeId`, but I don't think we really use the `Node` terminology anywhere, so I went with `brokers` instead.
2. The Java API has something called `AuthorizedOperations` that can be optionally included. I didn't add it as the current use-case doesn't need it, and I haven't actually looked into what it is or why you'd want it.

Usage is simply: `await admin.describeCluster()` and returns:

```javascript
{
  brokers: [{ nodeId: 0, host: 'localhost', port: 9092 }],
  controller: 0,
  clusterId: 'foobar'
}
```

Fixes #647


docs/Admin.md
```diff
@@ -184,6 +184,21 @@ await admin.setOffsets({
...
+await admin.describeCluster()
+// {
+//   brokers: [
+//     { nodeId: 0, host: 'localhost', port: 9092, connected: true }
+//   ],
+//   controller: 0,
+//   clusterId: 'f8QmWTB8SQSLE6C99G4qzA'
+// }
+```
+
```
JaapRood: Straightforward example, but maybe missing a little bit of context of situations _why_ you might want to do this. I always find example use cases very useful when I'm new to a library when trying to decide what I _have_ to do or _should_ be doing, what best practices are, etc.

src/admin/index.js
```diff
@@ -595,6 +597,55 @@ module.exports = ({
...
+
+    const { brokers: nodes, clusterId, controllerId } = await cluster.metadata({ topics: [] })
+    const brokers = await Promise.all(
+      nodes.map(async ({ nodeId, host, port }) => ({
+        nodeId,
+        host,
+        port,
+        connected: await isConnected(nodeId),
+      }))
+    )
```
JaapRood: I wonder what the consequences of querying the connection state asynchronously are. Technically, you could get stale `connected` parameters for some brokers, as some might resolve faster than others. That _could_ be totally fine, but I'm not sure what the use case for describing the cluster is and using these flags is.

More generally speaking, is the connection state query asynchronous because we're trying to wait for state transitions to resolve (like `connecting -> connected`)? Is that appropriate to the use case?

JaapRood:
@Nevon overall, seems like a straightforward PR worth adding. The Java API having exposed it is generally a good indication there's a good reason for it, #647 gives useful context as well. The two design decisions seems totally on point to me.

src/admin/index.js
Nevon: > is the connection state query asynchronous because we're trying to wait for state transitions to resolve (like connecting -> connected)?

No, the connection state query could technically be synchronous, because we are just looking at brokers that are already in memory, and then checking one of their properties to see if it is already connected. The only reason it is asynchronous is because it's using `Brokerpool.findBroker`, which normally also connects the broker if it is not already connected. For this use-case, we're passing an option that make it not connect the broker, so it's not really awaiting anything.

I could migrate the current `findBroker` to just... You know, _find_ the broker, and then create a new `findAndConnectBroker` or something that delegates to `findBroker` and then connects. That way `findBroker` can be completely synchronous.

docs/Admin.md
Nevon: I agree that a use-case would be nice. Can you think of any beyond the current one? The current use-case we have is to show a list of brokers and their current connection status in a vscode plugin, which is a bit specific. 😄 I tried finding a KIP for when `DescribeCluster` was introduced, but I couldn't find one, so maybe it was added before they were doing KIPs.

src/admin/index.js
Nevon: One problem for this particular use-case is that there's no way to know when the connection state changes, other than to poll `describeCluster`. That's really unrelated to this PR, but it could be solved with instrumentation events that are emitted on broker connect and disconnect.

docs/Admin.md
JaapRood: I think the current one is probably the valid use of it. Even if that's a bit on the nose, it would be a good to indication that this is mostly useful for monitoring / ops and _not_ essential to control flow or writing consumers. To a consumer, the shape of a cluster shouldn't really matter, instead just be focused on topics, partitions, offsets, groups, etc.

src/admin/index.js
JaapRood: Ahh, for some reason I thought `broker.isConnected()` returned a `Promise` as well. That doesn't mean `findBroker` also creating a connection isn't an issue from a monitoring standpoint.

Taking a step back, I'm wondering what one would learn from a 'connected' flag when describing an cluster through the Admin? Surely, it's describing whether that Admin is connected through the broker at that instant, but is that relevant information? You'd probably want to know whether your consumers / producers are in a monitoring use case, but if we create and connect an Admin to describe the cluster, yeah, we're going to be connected to the brokers. I guess, it can show whether the brokers in the cluster are reachable?

Nevon: Well, you're connecting an admin client, so at least one broker is going to be connected, but you won't necessarily be connected to the other brokers in the cluster. We just get all the brokers from metadata, but we don't connect to them from the admin client.

To be honest, I don't really see what problem is solved by that either, but it is a requirement for #647, and it is simple enough to provide.

JaapRood: While we _can_ add it easily enough, if the semantics don't really make sense for any use case of it, we're better of cutting it, as it's just waiting to be misinterpreted by someone.

> Taking a step back, I'm wondering what one would learn from a 'connected' flag when describing an cluster through the Admin? Surely, it's describing whether that Admin is connected through the broker at that instant, but is that relevant information? You'd probably want to know whether your consumers / producers are in a monitoring use case, but if we create and connect an Admin to describe the cluster, yeah, we're going to be connected to the brokers. I guess, it can show whether the brokers in the cluster are reachable?

@jlandersen you might hold the key to this and be able to enlighten us all 😊. Could you shine some light on why the `connected` flag would be useful for the use case of #647?

Nevon: I agree. Although I don't mind the inclusion of it, I would rather not introduce it if there isn't a good reason for it.

jlandersen: I don't have a good reason to include a connected flag. I would leave it out as well.
My extension does show in the broker list whether connected to a specific broker, but besides that, there is no real reason for it. Will probably just remove that.

Nevon: Removed in 1ca7f25

docs/Admin.md
Nevon: Fixed in daae6ed

docs/Admin.md
```diff
@@ -184,6 +184,22 @@ await admin.setOffsets({
 })
 ```
 
+## <a name="describe-cluster"></a> Describe cluster
+
+Allows you to get information about the broker cluster. This is mostly useful
+for monitoring or operations, and is usually not relevant for typical event processing.
```
JaapRood: Short and sweet way to prevent a whole bunch of confusion, nice!

PR #650, opened 2020-03-03
Add single requestTimeout runner instead of setTimeout per request

goriunov:
WIP, open for discussions

Propose change to `requestTimeout` logic to have separate runner which will check every inflight request every `x` ms. This is more performant than previous implementation and allow to enable requestTimeout for every single request without sacrificing performance much


goriunov:
I will fix tests properly if you are happy to process with this solution :)

src/network/requestQueue/index.js
```diff
@@ -42,6 +40,16 @@ module.exports = class RequestQueue {
           queueSize: this.pending.length,
         })
     }
+
+    // FIXME: How do we correctly clearInterval ?
```
Nevon: The `RequestQueue` is only instantiated in one place, so in that place we'd need to call a `requestQueue.destroy()` that calls `clearInterval` if there already is an instance of the request queue.

src/admin/index.spec.js
```diff
@@ -66,7 +66,6 @@ describe('Admin', () => {
     const emitter = new InstrumentationEventEmitter()
     const cluster = createCluster({
       requestTimeout: 1,
-      enforceRequestTimeout: true,
```
Nevon: I don't think we should remove the `enforceRequestTimeout` option, since that would be a breaking change. Let's start by making it work, enabling it ourselves, and then at some point make it active by default.

src/network/connection.js
```diff
@@ -199,6 +199,7 @@ module.exports = class Connection {
     }
 
     this.logDebug('disconnecting...')
+    this.requestQueue.destroy();
```
goriunov: @Nevon  Just to make sure that it works correctly. Do we expect that new Connection will be always created on reconnect ? Otherwise we can get situation when we cleared interval and we still reuse the same connection.

src/admin/index.spec.js
```diff
@@ -67,6 +67,7 @@ describe('Admin', () => {
     const cluster = createCluster({
       requestTimeout: 1,
       enforceRequestTimeout: true,
+      requestTimeoutCheckInterval: 1,
```
Nevon: Rather than introduce a top level option that requires a lot of knowledge to understand, can't we calculate an appropriate number based on the request timeout value? I'd rather not bloat the API if possible.

src/network/requestQueue/index.js
```diff
@@ -42,6 +43,14 @@ module.exports = class RequestQueue {
           queueSize: this.pending.length,
         })
     }
+
+    this.requestTimeoutIntervalId = setInterval(() => {
```
Nevon: We should only do this is `enforceRequestTimeout` is enabled, otherwise there's no need to create the setInterval.

src/network/connection.js
Nevon: It should be the case. Maybe in the setInterval we could check `this.connected` of the connection, and clear the interval if we are no longer connected?

goriunov: Good idea i think this will make sure to reset interval properly, will also timeout all inflight requests if we are not connected.

goriunov: I have tried adding isConnected handler, however not sure if it is worth of effort as before system is connected interval is running which will cancel immediately, we may need to move RequestQueue instantiation in to on connect handler (or have another function to trigger the interval in the RequestQueue). 

I can implement above logic if you would like to :) 

I have skimmed the code and looks like overall we correctly trigger disconnect which should clean up this interval.

src/network/requestQueue/socketRequest.spec.js
```diff
@@ -162,32 +156,5 @@ describe('Network > SocketRequest', () => {
         },
       })
     })
-
-    it('emits NETWORK_REQUEST_TIMEOUT', async () => {
```
goriunov: This test has been moved in to `requestQueue/index.spec.js` due to the timeout interval now been set in `index.js.

src/network/connection.js
Nevon: What makes me a little uncomfortable is that the `RequestQueue` is instantiated in the constructor of the connection, not in the `connect`, so as soon as the object is created the setInterval is started. I don't think that we call `disconnect` in every single of those cases (unit tests, for example). And even if we do, if we aren't connected, `disconnect` returns early, so the `setInterval` would not be cleared.

I think just passing in a function to the `RequestQueue` from the connection to check if the interval should be cleared is simple enough:

```js
this.requestQueue = new RequestQueue({
  getConnection: () => this
})
```

```js
class RequestQueue {
  constructor({ getConnection }) {
    if (this.enforceRequestTimeout) {
        const requestTimeoutCheckInterval = Math.min(this.requestTimeout, 100)
        this.requestTimeoutIntervalId = setInterval(() => {
            this.inflight.forEach(request => {
                if (Date.now() - request.sentAt > request.requestTimeout) {
                    request.timeoutRequest()
                }
            })
            const connection = getConnection()
            if (!connection || !connection.isConnected()) {
                clearInterval(this.requestTimeoutIntervalId)
            }
        }, requestTimeoutCheckInterval)
    }
  }
}
```

goriunov: @Nevon in this case we will clear the interval even before the system is actually connected as by default `this.connected` is `false` and if we are unable to connect (call `onConnect` in `Connection` class) within the first interval call it will clear the interval. May be it is better to move RequestQueue creation in to `onConnect` ? Or move `setInterval` logic in to separate function and call it from `onConnect` ? 

like:

```js
  const onConnect = () => {
        clearTimeout(timeoutId)
        this.connected = true
        this.requestQueue.scheduleRequestTimeoutCheck()
        resolve(true)
      }
```

Nevon: Right, of course 🤦‍♂️

Try moving it into connect. I'm wondering if there's a case for adding things to the queue before connecting, but I'm not sure.

goriunov: @Nevon had a look at `send` method looks like we have `failIfNotConnected` so there should not be any way to push in the queue before connection. I would try to move queue creation in to onConnect

goriunov: @Nevon i have done it slightly different way it will also work properly and in my opinion suits better. Would like you to check it out and let me know if anything should be changed.

Nevon:
This looks really good now, in my opinion.

Nevon:
Thanks for the contribution! There will be a pre-release version out once the master build is done (https://dev.azure.com/tulios/kafkajs/_build/results?buildId=747&view=results). It should be version `1.13.0-beta.6`. If you have a need for this feature, it would be great if you would give the pre-release version a spin and let us know how it's working.

goriunov:
:+1:  Definitely will try to run it within next few week. Thanks :)

Issue #654, opened 2020-03-09
Reduce test flakiness


Nevon:
The test pipeline has been reaching new levels of flakiness over the last month or so. On master, the current pass rate of the pipeline is 40%. Looking at the stats from Azure pipelines it's a small number of tests that contribute to the flakiness (this is including PRs, so some of these could be legit failures, but the recurring ones are probably due to flakiness):

<img width="998" alt="Screenshot 2020-03-09 at 18 42 31" src="https://user-images.githubusercontent.com/83586/76241806-d97f2a80-6235-11ea-8fdb-1c1c372f756a.png">

Putting some effort into the worst offenders will probably improve the pipeline stability immensely.

ankon:
Somewhat related: When trying to run the tests they are configured to "bail" on the first error: https://github.com/tulios/kafkajs/blob/897715fcc812820284c5ef3d76bd07bf4fac1478/jest.config.js#L12

Given the flakiness and the considerable costs for actually starting the tests compared to running them I would propose to actually switch bail to `false`, at least in the default configuration -- if you find it useful for Azure maybe it could be configured in the CI settings through a command-line option instead.
Right now I try running the tests locally but given the flakiness combined with the bailing I give them very little attention: They likely don't even run tests for code I just touched anyways. Keeping a local change to disable bailing is ugly as well, so right now "don't bother with tests" is the way of least resistance :)

Nevon:
Yeah I think we should change that setting. It was added as PR #7, so it's quite old 😄 

I guess it was set because it was better to get early feedback then to have to wait for tons of tests to run, but while developing I anyway always run targeted tests in watch mode, so it's not a problem to run the remaining tests even if one fails. Even in CI I would rather see all the failing tests than stop as soon as a suite fails.

Maybe it's some help to you, but the way I usually do is:

```sh
$ ./scripts/dockerComposeUp.sh
$ ./scripts/createScramCredentials.sh
$ yarn test:local:watch [path-to-spec-file]
```

That way I don't have to restart containers if I restart the tests, and jests interactive mode means I can switch between which tests to run as I'm developing.

ankon:
Thanks, these commands will come in handy.

I filed https://github.com/tulios/kafkajs/pull/672, let's see how Azure likes that :)

Nevon:
Gonna close this, as the pipeline is much better these days. Whenever I find a test that seems to be flaky, I add the `[flaky]` prefix to the name so that I can run only flaky tests to try and fix them.

PR #655, opened 2020-03-09
Retry failing tests in CI

Nevon:
The new jest runner, jest-circus, supports re-running individual failing tests. I have changed our tests to use that runner and to set the retry to 2 in CI (0 when running locally). One downside is that the junit test report doesn't include the number of retries per test case. The runner exposes it as `invocations`, but there's no place in the junit test report spec for that information. But if it makes it possible for our pipeline to succeed, I'll take that small downside.

This doesn't solve #654, but it's a start.


src/admin/index.spec.js
```diff
@@ -79,15 +81,20 @@ describe('Admin', () => {
...
+    await retrier(async () => {
+      try {
+        await admin.createTopics({
           waitForLeaders: false,
           topics: [{ topic: `test-topic-${secureRandom()}` }],
         })
-      )
-      .catch(e => e)
+      } catch (e) {
+        if (e.name !== 'KafkaJSRequestTimeoutError') {
```
Nevon: The theory here is that in some circumstances, `createTopics` failed with some other error than a request timeout, which would make the test fail. Now we retry on any other retriable failure, which should hopefully fix it.

src/consumer/__tests__/errorRecovery.spec.js
```diff
@@ -204,7 +204,7 @@ describe('Consumer', () => {
     await waitFor(() => restartOnFailure.mock.calls.length > 0)
     expect(restartOnFailure).toHaveBeenCalledWith(error)
 
-    await expect(waitFor(() => eachMessage.mock.calls.length)).resolves.toBe(1)
+    await expect(waitFor(() => eachMessage.mock.calls.length)).resolves.toBeGreaterThanOrEqual(1)
```
Nevon: Sometimes the consumer restarts before this `waitFor` is checked, which causes `eachMessage.mock.calls` to have 2 calls instead of one.

src/consumer/__tests__/runner.spec.js
```diff
@@ -247,22 +247,21 @@ describe('Consumer > Runner', () => {
...
       consumerGroup.commitOffsetsIfNecessary.mockClear()
       consumerGroup.commitOffsets.mockClear()
+    })
 
+    it('should commit offsets while running', async () => {
       await runner.commitOffsets(offsets)
+
       expect(consumerGroup.commitOffsetsIfNecessary).toHaveBeenCalledTimes(0)
-      expect(consumerGroup.commitOffsets).toHaveBeenCalledTimes(1)
+      expect(consumerGroup.commitOffsets.mock.calls.length).toBeGreaterThanOrEqual(1)
```
Nevon: Sometimes the `eachMessage` has time to run, so because of `autoCommit` `commitOffsets` has been called twice by the time we assert.

PR #657, opened 2020-03-09
Always return an iterator from ConsumerGroup.fetch

### LINKED ISSUE #656: Unhandled promise rejection in consumerGroup.fetch
Since introducing the `BufferedAsyncIterator` to `ConsumerGroup.fetch`, the test below is extremely flaky. It is currently the biggest contributor to #654. It has two failure modes, and I'm not sure if they are related or not. This is the less common one, but the more worrying one:

```
(node:32993) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 201)
 FAIL  src/consumer/__tests__/assignmentForUnsubscribedTopic.spec.js
  ● Consumer › starts consuming from new topics after already having assignments

    KafkaJSNonRetriableError
      Caused by: TypeError: iterator.next is not a function

      319 |
      320 |     while (true) {
    > 321 |       const result = iterator.next()
          |                               ^
      322 |       if (result.done) {
      323 |         break
      324 |       }

      at Runner.fetch (src/consumer/runner.js:321:31)
      at src/consumer/runner.js:382:9
```

It looks like `iterator` is sometimes not an iterator, which makes sense as there are two places where it returns early by returning an empty array:

https://github.com/tulios/kafkajs/blob/728a32556a4a9dc50578e7d47c1421030f9599b2/src/consumer/consumerGroup.js#L353

https://github.com/tulios/kafkajs/blob/728a32556a4a9dc50578e7d47c1421030f9599b2/src/consumer/consumerGroup.js#L463

My current assumption is that these are the cause of the issue. I naively tried to wrap the empty arrays into a `BufferedAsyncIterator`, but then all hell broke loose. I then replaced the `[]` with `(function*() {})()`, which seems to work, but  I don't feel I understand the issue enough to feel confident in the change. I would like some input from @tulios. 

The following test passes just fine:

```javascript
it('yields an empty iterator if passed an empty array of promises', async () => {
  const iterator = BufferedAsyncIterator([])

  expect(result.done).toBe(true)
})
```

Nevon:
I have no idea why I was having issues with this before, but now it works just fine.

Fixes #656


src/consumer/consumerGroup.js
```diff
@@ -350,7 +352,7 @@ module.exports = class ConsumerGroup {
         })
 
         await sleep(this.maxWaitTime)
-        return []
+        return EMPTY_FETCH_RESULT
```
tulios: It's better to return a new instance here, iterators have an internal state.

Nevon: Given that we want a "done" iterator for this case, isn't that a benefit?

Issue #658, opened 2020-03-10
createPartitions missing method in admin client


adalbert-homa:
This method is mentioned in the documentation for the admin client, however is not implemented or documented further.

Nevon:
That is very true. Looking at the protocol for `createPartitions`, it looks quite simple to implement, if you want to take a crack at it: https://kafka.apache.org/protocol#The_Messages_CreatePartitions

PR #661, opened 2020-03-13
Add createPartitions to admin client

adalbert-homa:
Fixes: #658


examples/admin.js
```diff
@@ -0,0 +1,162 @@
...
+  return md.topics.filter(filterTopics).sort(sortTopics)
+}
+
+const wait = nrOfSec => {
+  return new Promise((resolve, _) => {
+    setTimeout(resolve, nrOfSec * 1000)
+  })
+}
+
+const run = async () => {
```
Nevon: I think this file should not be part of master. The examples in here should be as minimal as possible, and show a simple version of a common task (produce a few messages, consume from a topic). This is way too detailed and will be hard for people to understand.

src/broker/index.js
```diff
@@ -504,6 +504,26 @@ module.exports = class Broker {
     return await this.connection.send(createTopics({ topics, validateOnly, timeout }))
   }
 
+  /**
+   * @public
+   * @param {Array} topicPartitions e.g:
```
Nevon: This is missing the `count` argument., and in the documentation `newAssignments` is referred to as `assignments`. I personally prefer `assignments`.

adalbert-homa: Will fix it. I used assignments in the requests, so it should match here

src/admin/index.js
```diff
@@ -124,6 +124,52 @@ module.exports = ({
...
+
+    const retrier = createRetry(retry)
+
+    return retrier(async (bail, retryCount, retryTime) => {
+      try {
+        await cluster.refreshMetadata()
+        const broker = await cluster.findControllerBroker()
+        await broker.createPartitions({ topicPartitions, validateOnly, timeout })
+
+        return true
```
Nevon: Instead of resolving with `true`, let's just resolve with nothing. It's already clear that resolving means the operation worked and rejecting means it didn't. That leaves us room to add a proper response if future versions of the protocol starts returning some meaningful information that we need to expose.

adalbert-homa: Done

Nevon:
Awesome! The pre-release version `1.13.0-beta.7` will be out with this feature.

PR #663, opened 2020-03-14
Remove long.js in favor of BigInt

Nevon:
Closes #617.

I took another stab at driving #617 to conclusion. The main two things were:

* Fix the linter, which didn't understand the BigInt literal (4000e84)
* Fix an issue where `Long.toInt` would return a floating point double rather than a 32 bit integer, which caused the murmur2 hashing to fail. (8e7ceb0)


.eslintrc
```diff
@@ -1,26 +1,42 @@
 {
   "extends": [
     "standard",
-    "prettier"
+    "prettier",
+    "plugin:node/recommended"
```
Nevon: This was necessary to get BigInt literals to be supported by eslint.

src/broker/__tests__/addPartitionsToTxn.spec.js
```diff
@@ -86,7 +86,7 @@ describe('Broker > AddPartitionsToTxn', () => {
     await expect(
       broker.addPartitionsToTxn({
         transactionalId,
-        producerId: 'foo',
+        producerId: '123456789',
```
Nevon: I looked at whether we should check for this early and validate that the producer id is something that can be converted to a bigint, but since it's not a public api, I think it's fine to leave as it is. The change here is just to make the operation fail for the right reason.

src/protocol/decoder.js
```diff
@@ -57,11 +57,22 @@ module.exports = class Decoder {
...
 
   readInt64() {
-    const lowBits = this.buffer.readInt32BE(this.offset + 4)
-    const highBits = this.buffer.readInt32BE(this.offset)
-    const value = new Long(lowBits, highBits)
+    const first = this.buffer[this.offset]
+    const last = this.buffer[this.offset + 7]
+
+    const low =
+      (first << 24) + // Overflow
```
Nevon: I tried to change this to use left shifting everywhere, instead of the exponentiation, since in my mind `x << y` is the same as `x * 2 ** y`, but it seems that is not the case, because when I tried I always got weird results.

Nevon: One thing I still don't really understand is that in every other implementation I've seen, reading a BigInt64 from a buffer first checks if there are at least 8 bytes available (the length of a 64 bit integer), but then it only advances the offset by 6 bytes (where I'm doing `offset + n` the other implementations do `++offset`). That would lead to incorrect results, so I don't understand why they do that or why it works for them.

src/protocol/encoder.js
```diff
@@ -8,7 +8,7 @@ const INT64_SIZE = 8
 const MOST_SIGNIFICANT_BIT = 0x80 // 128
 const OTHER_BITS = 0x7f // 127
 const UNSIGNED_INT32_MAX_NUMBER = 0xffffff80
-const UNSIGNED_INT64_MAX_NUMBER = Long.fromBytes([-1, -1, -1, -1, -1, -1, -1, -128])
+const UNSIGNED_INT64_MAX_NUMBER = Long.fromNumber(-128)
```
Nevon: I'm not sure why this variable is called `UNSIGNED_INT64_MAX_NUMBER`, because that's definitely not `-128`. `-128` is the lowest number that possible to represent with an 8 bit signed integer, but I have long forgotten how exactly this piece of code works...

Nevon: Changed it to `0xffffffffffffff80n`, which makes a hell of a lot more sense than -128. No clue why it worked with -128 to begin with...

tulios: I will 10h to explain this, jokes aside, it's hard to remember the details of number based things. I called this `*_MAX_NUMBER`, but I don't remember why anymore...

tulios:
⭐️ @Nevon , I will try this branch out for a bit. This is quite a big change!

src/protocol/encoder.js
Nevon: Yeah, it's something about calculating the size of a varlong. I tried to find the documentation that we read when doing the initial implementation, to see if I could find a clue, but I couldn't find them.

Nevon:
> ⭐️ @Nevon , I will try this branch out for a bit. This is quite a big change!

Did you ever give this a whirl?

tulios (approved):

kessiler:
Hi, @tulios and @Nevon I think this PR drops the support for NodeJS 8 and downwards. Is that desired?
It should probably be mentioned in your README or at least a major bump since you guys follow semver practices.

Nevon:
Node 8 support was already dropped in a previous PR which was included in 1.13.0, which is mentioned in the release notes of that release.

You're right that normally that would be a breaking change that would warrant a major version. In truth, we accidentally released it in a minor because we simply forgot that we had included that change (it was a huge release). When we later realized, we decided to update the release notes to make a note of it rather than deprecate 1.13.0 and re-release it as 2.0.0. The reason being that Node 8 has not been supported for over 9 months now by Node itself, so no one should be using Node 8 anymore since you aren't even receiving security updates. Therefore we figured this wouldn't actually be an issue for 99.999% of people.

kessiler:
Totally agree with you @Nevon, I didn't notice the drop support note in the changelog, that's actually the reason I was trying to bring this up here. Thanks for your attention 👍

Issue #651, opened 2020-03-05
Deleted topics causing cascading metadata refreshes


kburdett:
I have encountered an issue with extensive metadata refreshes when a topic is deleted from the Kafka cluster. The refresh counts are far in excessive of the retry policy.

KakfkaJS 1.12.0
Node 11.15.0

### Symptoms
A producer performs many metadata refreshes, spanning close to a minute, for a single send request. Normally, the retry policy limits this to 5 attempts and the error is returned in a few seconds. However, in the case of a deleted topic, the refresh is attempted many times and introduces a significant delay in response.

### Steps to reproduce
1. Create a producer
2. Send a message to topic X
3. Delete X from the Kafka cluster
4. Send another message to topic X

### What I have uncovered so far
This seems very similar to #252, but the fix there does not correct this particular problem. I think there are really two parts to this problem:

First, it seems that when metadata is fetched for a topic, this topic name is added to the `targetTopics` Set on the cluster object. Future metadata refreshes for this producer include all members of the `targetTopics` Set. So, once _any_ member of the `targetTopics` set is deleted, _all_ metadata refreshes on this producer fail with an "This server does not host this topic-partition" error. 

Second, there does seem to be nested instances of the retrier at play here. That would explain why the 5 retry policy is being violated. In my case, there were 42 errors with message "Response Metadata" and 36 with "Response Produce" taking just over a minute while using the default retry policy. Possibly 5 retries on the send with 5 retries on metadata refresh per send attempt? The numbers don't quite add up, but that could be a timing thing? I'm uncertain.

### Possible solutions
I am happy to make an attempt a PR but I'd like some feedback first, as I don't know the codebase very well.

I thought to simply remove the topic from the `targetTopics` Set when this occurs. This is the solution implemented in #252. I haven't dug into the returned message, but I'm assuming that Kafka doesn't identify which topic is at fault, so this might not be possible when `targetTopics` contains more than one topic.

On a deeper level, I'm not really sure what `targetTopics` is here for. It seems like the producer is "over-refreshing" and fetching metadata for _all topics_ that it knows about, rather than just the topic _currently being acted upon_. If we could avoid this behavior, then a solution would perhaps be easier here. We could possibly clear the `targetTopics` Set when refresh errors are encountered.  I'm not sure what the implications could be here. Possibly `targetTopics` could be eliminated in cases where only a single topic refresh is necessary?

As for the over-retry issue, I've not made much progress here. Any guidance would be appreciated.

[...]

PR #666, opened 2020-03-16
Remove invalid topics from targetTopics on INVALID_TOPIC_EXCEPTION

### LINKED ISSUE #665: KafkaJSProtocolError: The request attempted to perform an operation on an invalid topic
Hi, we just found a weird problem about " The request attempted to perform an operation on an invalid topic".
we use dynamic topic like this:
```
function produce(type, content, response) {

  const topic = TOPIC_PREFIX + content.project + '-' + content.platform + '-' + type;
  let message = JSON.stringify(content);
  
  kafka.produce(topic, message).then(() => {
    output(response, 200, 'OK');
  }).catch(err => {
    output(response, 252, 'server error.');
  });
}
```
and when the value of _type_  contains special character, such as _type =  abc)(*&^%_ , it will throw an error like " KafkaJSProtocolError: The request attempted to perform an operation on an invalid topic".

that is all OK.

but after that , all produce operations , include the normal topic , can not work. the all throw a invalid topic error.

if we restart the server, the problem will fixed. but  if the wrong topic appear agian, and the problem will come back.

plz help us , thanks!

Nevon:
Fixes #665

### DIFF
--- a/src/cluster/index.js
+++ b/src/cluster/index.js
@@ -1,4 +1,5 @@
 const BrokerPool = require('./brokerPool')
+const Lock = require('../utils/lock')
 const createRetry = require('../retry')
 const connectionBuilder = require('./connectionBuilder')
 const flatten = require('../utils/flatten')
@@ -78,6 +79,10 @@ module.exports = class Cluster {
     })
 
     this.targetTopics = new Set()
+    this.mutatingTargetTopics = new Lock({
+      description: `updating target topics`,
+      timeout: requestTimeout,
+    })
     this.isolationLevel = isolationLevel
     this.brokerPool = new BrokerPool({
       connectionBuilder: this.connectionBuilder,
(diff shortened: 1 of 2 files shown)


src/cluster/index.js
```diff
@@ -163,14 +163,23 @@ module.exports = class Cluster {
...
 
     const hasChanged = previousSize !== this.targetTopics.size || !this.brokerPool.metadata
 
     if (hasChanged) {
-      await this.refreshMetadata()
+      try {
+        await this.refreshMetadata()
+      } catch (e) {
+        if (e.type === 'INVALID_TOPIC_EXCEPTION') {
+          this.targetTopics = previousTopics
```
JaapRood: By storing the previous topics per method call, is there a possibility of us introducing a race condition here? I'm too unfamiliar with this part of the codebase whether that's something reasonable to expect or not.

If we _are_ creating a race condition here, the most simple way to mitigate in my mind is to store the previous topics on the cluster instance and use that (or a Promise of it) as a lock to only allow a single inflight call. Alternatively, we could be more careful with adding new topics to the targets list, only revert the ones for which we got the exception (which should be idempotent, so deletion of them of a Set should be without consequence).

Nevon: Ah, you are a learned and erudite gentleman. I didn't think about the possibility of a race condition here, but you are indeed correct that that could happen.

Long-term I think we should actually remove the list of target topics, if at all possible, as mentioned in #651, but in the meantime I think a simple lock here would be fine.

Nevon: Fixed in `091f5c`.

JaapRood: You reckon it's worth adding a test case that confirm the lock works as expected? It could be useful, but triggering race conditions reliably is hard and an ingredient for flaky tests.

Nevon: That was exactly my thinking. I would like to test it, but it also sounds nightmareish to get such a test to work consistently.

ankon:
Mostly FYI: It seems I'm able to hit a lock timeout for this new lock in our production environment. What is worrying here is the number of "waiting locks" in the log messages goes continuously up, starting from 2 to 1789 after several hours when I finally killed the process.

PR #667, opened 2020-03-17
Eliminate target topics list on the cluster

kburdett:
Following the conversation in #651, this PR removes the `targetTopics` collection in an effort to prevent refreshing more topics than necessary.


kburdett:
I tried to be as surgical as possible while making edits. I'm unfamiliar with the codebase, so I've tried to change as little as makes sense. This is a bit out of character for me, so I'll gladly expand scope where you guys feel necessary. Any and all feedback is appreciated

There are a couple tests still failing for me, I'm not sure if it is an issue with my local environment, or a problem with the changes. I'm still trying to track that down. Based on comments from @Nevon , I've created this PR anyways.

kburdett:
Some general notes and discussion points...

It seemed to me that the `targetTopics` list was most important to ensure that valid metadata wasn't overwritten by subsequent refreshes. I was able to work around this by moving to a simplistic cache map for the topic metadata. To limit the scope of the changes, I left the old metadata object in place, as is. I think perhaps a more holistic approach to a metadata cache could be warranted, but it's a pretty big jump in effort, so I'd like to hear your thoughts first.

For breaking changes, two "public" members were eliminated from Cluster, specifically`addTargetTopic` and `addMultipleTargetTopics`. I put public in quotes because I believe this entire object may well be internal. `getTopicMetadata` changes a bit as well. This function was reliant on the other two for establishing it's behavior. In this PR, it now behaves very similar to `fetchTopicMetadata`.

The metadata cache, in general, felt like it was bleeding across a few places. I considered moving it, wholesale, to the Cluster and out of the BrokerPool but I thought it better to keep the scope small until someone more knowledgeable could offer an opinion.

kburdett:
Also, I didn't see any commit guidelines, but I can easily squash or edit these commits as you like. Please let me know :)

src/admin/index.js
```diff
@@ -187,15 +185,11 @@ module.exports = ({
...
+        await cluster.refreshMetadata(topics)
         const broker = await cluster.findControllerBroker()
         await broker.deleteTopics({ topics, timeout })
 
-        // Remove deleted topics
-        for (const topic of topics) {
-          cluster.targetTopics.delete(topic)
-        }
-
+        // full metadata refresh to remove recently deleted topics
```
Nevon: Hmm, I don't really like using `refreshMetadata` with an empty array as a way to clear stale metadata. I would rather have an explicit `removeMetadata` or something.

src/admin/index.js
```diff
@@ -687,9 +667,8 @@ module.exports = ({
    */
   const listGroups = async () => {
     let groups = []
+    await cluster.refreshMetadata()
```
Nevon: I suspect this was done intentionally once for each node, to avoid having stale metadata. This could just be turned into a `Promise.all` instead of looping through each broker one by one though.

Nevon:
> It seemed to me that the targetTopics list was most important to ensure that valid metadata wasn't overwritten by subsequent refreshes. I was able to work around this by moving to a simplistic cache map for the topic metadata. To limit the scope of the changes, I left the old metadata object in place, as is. I think perhaps a more holistic approach to a metadata cache could be warranted, but it's a pretty big jump in effort, so I'd like to hear your thoughts first.

I think your approach is valid. A simple topic -> metadata map with an expiry key should be enough, I think. We might have to think of an expiration mechanism though, to make sure that we eventually get rid of old metadata for topics that are no longer required. 

> The metadata cache, in general, felt like it was bleeding across a few places. I considered moving it, wholesale, to the Cluster and out of the BrokerPool but I thought it better to keep the scope small until someone more knowledgeable could offer an opinion.

I've had the same thought. I've seen the cluster reach into the brokerpool to read the metadata several times, which probably means it shouldn't live there (or maybe the broker pool should expose it through a public method). Not sure if it makes more sense to keep it in brokerpool or more it to the cluster. I think that can be done as a separate thing though.

src/admin/index.js
kburdett: Yeah, it isn't so much removing the metadata, as refreshing everything. However, your point is taken and this refresh is unnecessary. I'll add something.

src/admin/index.js
kburdett: I thought so too, but I have a couple issues with it...

1. I dug through `refreshMetadata` and apart from the `shuffle` present in `findConnectedBroker`, I don't see any way to change or control which connected broker is used for the connection. Repeatedly refreshing in this method seems an unreliable, at best, method of grabbing metadata from each broker.
2. Pulling metadata from each broker seems unnecessary. It's the same metadata, right? If it's not the same metadata, then the cluster is in a split brain state and we have bigger problems. This is especially true since the topic metadata isn't used. I believe only the broker list is required to fulfill this particular request.

Let me know your thoughts. I'll move the refresh back into the loop and change it to a `Promise.all` if you still feel it's better to keep it where it is.

src/cluster/brokerPool.js
```diff
@@ -122,6 +125,11 @@ module.exports = class BrokerPool {
     const broker = await this.findConnectedBroker()
     const { host: seedHost, port: seedPort } = this.seedBroker.connection
 
+    // clear expired entries to prevent cache bloat
```
kburdett: @Nevon Regarding:
> I think your approach is valid. A simple topic -> metadata map with an expiry key should be enough, I think. We might have to think of an expiration mechanism though, to make sure that we eventually get rid of old metadata for topics that are no longer required.

I added this code as part of `refreshMetadata` for this reason. `refreshMetadata` already takes place on semi-regular intervals due to the `metadataMaxAge` setting. It's not a particularly tight mechanism to manage cache expiry, but it will prevent uncontrolled growth.

src/cluster/brokerPool.js
```diff
@@ -122,6 +125,11 @@ module.exports = class BrokerPool {
...
+    Array.from(this.topicMetadataCache.entries())
+      .filter(([_, cached]) => Date.now() > cached.expireAt)
+      .forEach(([topic]) => this.topicMetadataCache.delete(topic))
```
kburdett: @Nevon Regarding:
> I think your approach is valid. A simple topic -> metadata map with an expiry key should be enough, I think. We might have to think of an expiration mechanism though, to make sure that we eventually get rid of old metadata for topics that are no longer required.

I added this code as part of `refreshMetadata` for this reason. `refreshMetadata` already takes place on semi-regular intervals due to the `metadataMaxAge` setting. It's not a particularly tight mechanism to manage cache expiry, but it will prevent uncontrolled growth.

kburdett:
I am finally getting back to this PR. I rebased and corrected conflicts with the master branch. I was able to confirm that this fixes half of the issue identified in #651. The removal of target topics does indeed protect unaffected topics from being trapped in the refresh loop. However, it is still over-refreshing on topics when they are deleted. I'm digging into that currently.

While I continue investigating, any additional feedback is welcome.

kburdett:
This PR is ready to go. The other issue identified in #651 will be handled separately.

Please let me know if any further changes are needed on this PR.

tulios:
@kburdett sorry for the long delay, I will review this one tomorrow. I think @Nevon already looked at this PR.

suvorovis:
Any updates here? I have faced the issue described in https://github.com/tulios/kafkajs/pull/1185 in production with a producing rate of 6k/sec.

Nevon:
Here we are, 2 years later...

This PR needed a bit of attention to adapt to changes that have been made since then, especially since we restructured a large chunk of the consumer fairly recently. I went ahead and made those changes, and things look good. My plan is to merge this tomorrow, after having gone through the codebase once more (this time I was mostly just looking to get master merged in). Specifically, I want to make sure that there's no mistake somewhere that's causing us to make way more metadata requests than before, as I had this happen while I was working on this.

PR #668, opened 2020-03-18
Simplify and speed up the SeekOffsets code

ankon:
* ~has: Use the similar function from the parent Map class, hoping that that
  is potentially faster than the brute-force approach through keys/some~
* pop: Replace the complex and inefficient O(n)-ish logic with a simpler
  O(1) approach.
  We don't need to make an array of all entries, reverse that, and then pop
  one element -- we only need to know the first key/value pair.


ankon:
Hold off on this one please.

`has` doesn't work this way unfortunately, because of how the Map equality works. The change in `pop` should still be fine.

ankon:
Removed the `has` change; this should be good to go now.

I see the checks failing, and I have some issues locally on getting a stable test run -- looks a lot like flakes to me though (apart from the Azure database misconfiguration?)

Nevon:
Could you merge latest master into this, and we can get it merged?

ankon:
@Nevon sure :)

Issue #674, opened 2020-03-23
Support for OAUTHBEARER SASL mechanism


uwburn:
I have stumbled upon https://github.com/strimzi/strimzi-kafka-oauth that provides an implementation (both broker and Java client sides) for OAUTHBEARER mechanism.

As it looks like an interesting way to define programmatically authentication and authorization in Kafka, i'm guessing if there's any plan or interest in integrating some kind of support for OAUTHBEARER, so that it can be leveraged to send OAuth tokens for authentication.

PS
I might be willing to help with a PR, even if at the moment i didn't find a matching between the actual implementation of the existing mechanisms and the protocol spec at https://kafka.apache.org/protocol. Did you had to reverse engineer the Java client for it?

tulios:
The protocol just tells you how to format the bytes; it's sadly hard to understand how everything is supposed to work. This feature is definitely wanted, and I think it is based on the token mechanism. Kafka introduced a change some time ago where the mechanisms can re-auth, so you can refresh the tokens.

I looked into this some time ago, I can try to find the KIP that describes the mechanism and it would be awesome to have someone driving the change.

uwburn:
It's https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876 the KIP you're referring to?

By looking at the Java library i linked it looked like a token renewal was needed, but looking at the proposal it seems the token is evaluated at connection time, and it stays valid as long as the connection is alive. Eventually, it needs to be refreshed in case of reconnections.

As the KIP notes, it might be a problem in case the token is used also for authorization, but the proposal seems to consider this an edge case and unlikely to be problematic.

I will try expiring the token with the Java impl. and see what happens, just to get the idea.

tulios:
Yes @uwburn , but I remember another one handling re-auth. Nice, we can try to enrich this issue with information.

uwburn:
Found it: https://cwiki.apache.org/confluence/display/KAFKA/KIP-368%3A+Allow+SASL+Connections+to+Periodically+Re-Authenticate

I have also found the following, giving some explanations at least about the initial authentication:

- Issue explainin part of the OAUTH SASL payload: https://issues.apache.org/jira/browse/KAFKA-7182
- RFC 5801 - GS header: https://tools.ietf.org/html/rfc5801
- A pull request about OAUTH SASL payload: https://github.com/apache/kafka/pull/5391
[...]

PR #680, opened 2020-03-28
Oauthbearer support

uwburn:
#674 

Implemented the SASL OAUTHBEARER request, the reauthentication mechanism seems to be handled automatically by already present `isConnected()` check in `src/broker/index.js`, as it evalueates the sessionLifetime that is replied back from Kafka when proper options (connections.max.reauth.ms) are configured in the broker.

A base spec ha been introduced to check the most obvious errors, more elaborate testing seems difficult to set up, as requires plenty of configuration on the broker.

Some tests are failing, but they do seem unrelated.

Updates to the documentation are missing at the moment.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -66,6 +66,41 @@ new Kafka({
 })
 ```
 
+### OAUTHBEARER Example
+
+```javascript
+new Kafka({
+  clientId: 'my-app',
+  brokers: ['kafka1:9092', 'kafka2:9092'],
+  // authenticationTimeout: 1000,
+  // reauthenticationThreshold: 10000,
+  ssl: true,
+  sasl: {
+    mechanism: 'oauthbearer',
+    oauthBearerProvider: async () => {
+      // Use an unsecured token...
+      const token = jwt.sign({ sub: 'test' }, 'abc', { algorithm: 'none' })
+
+      // ...or, more realistically, grab the token from some OAuth endpoint
+
+      return {
+        value: token
+      }
+    }
+  },
+})
+```
+
+The sasl object must include a property named oauthBearerProvider, an
+async function that is used to return the OAuth bearer token.
+
+The OAuth bearer token must be an object with properties value and
+(optionally) extensions, that will be sent during the SASL/OAUTHBEARER
+request.
+
+The implementation of the oauthBearerProvider must take care that tokens are
+reused and refreshed when appropriate.
+
 ### AWS IAM Example
 
 ```javascript
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -91,7 +91,7 @@ new Kafka({
 })
 ```
 
-The sasl object must include a property named oauthBearerProvider, an
+The `sasl` object must include a property named `oauthBearerProvider`, an
 async function that is used to return the OAuth bearer token.
 
 The OAuth bearer token must be an object with properties value and
--- a/src/broker/saslAuthenticator/oauthBearer.spec.js
+++ b/src/broker/saslAuthenticator/oauthBearer.spec.js
@@ -1,0 +1,18 @@
+const { newLogger } = require('testHelpers')
+const OAuthBearer = require('./oauthBearer')
+
+describe('Broker > SASL Authenticator > OAUTHBEARER', () => {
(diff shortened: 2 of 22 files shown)


src/broker/saslAuthenticator/oauthBearer.spec.js
```diff
@@ -0,0 +1,18 @@
+const { newLogger } = require('testHelpers')
+const OAuthBearer = require('./oauthBearer')
+
+describe('Broker > SASL Authenticator > OAUTHBEARER', () => {
```
Nevon: Would be nice to have a success case test as well

uwburn:
I wrote a test using for OAUTHBEARER login, but the problem is that Kafka fails to start with exception `java.lang.IllegalArgumentException: Must supply exactly 1 non-null JAAS mechanism configuration (size was 3)` when configured to use the embedded OAUTHBEARER handler.

Unfortunately i found other users having the same issue, by looking at the Java class at https://github.com/apache/kafka/blob/b8651d4e82d45463d2c71798bd5852f8a605b440/clients/src/main/java/org/apache/kafka/common/security/oauthbearer/internals/unsecured/OAuthBearerUnsecuredValidatorCallbackHandler.java it looks like it doesn't properly lookup the JAAS entry but just checks that there's only one and pick it, so it gets confused by PLAN and SCRAM entries. I also tried declaring a dedicated listener for the test but it still refuse to start with the same error.

I opened a branch `oauthbearer-support-test` in my fork if you want to check it. From it you can manually check that it works, but you have to remove non OAUTHBEARER SASL mechanisms and remove non OAUTHBEARER entries from KafkaServer JAAS section. Should i include the test keeping it commented, in case in future we can use it?

In the meantime i added the docs.

uwburn:
Do you have any plan for merging the PR and having the funcionality available in GA releases?

uwburn:
Rebased.

Nevon:
> Do you have any plan for merging the PR and having the funcionality available in GA releases?

Sorry about the delay. It seems all the maintainers, including myself, are busy with life and work lately. 

> Should i include the test keeping it commented, in case in future we can use it?

I think that sounds good, but rather than commenting it out, let's do what we do for running tests with different versions of Kafka: https://github.com/tulios/kafkajs/blob/b13567d66da541b7049a9f4e0e981d15e28158bc/testHelpers/index.js#L193-L199

So something like:

```js
const testIfEnv = (key, value) => (description, callback, testFn = test) => {
  return value === process.env[key]
    ? testFn(description, callback)
    : test.skip(description, callback)
}

/**
 *  <Explanation why oauthbearer is conditionally enabled and how to enable it>
 */
const testIfOauthbearerEnabled = testIfEnv('OAUTHBEARER_ENABLED', '1')
```

Then you test would be just the following, and it would run if `OAUTHBEARER_ENABLED` is `1`, and be skipped otherwise.

```
testIfOauthbearerEnabled('support SASL OAUTHBEARER connections', async () => {
  // ...
})
```

docs/Configuration.md
```diff
@@ -66,6 +66,41 @@ new Kafka({
...
+
+      return {
+        value: token
+      }
+    }
+  },
+})
+```
+
+The sasl object must include a property named oauthBearerProvider, an
```
Nevon, line 94:
```suggestion
The `sasl` object must include a property named `oauthBearerProvider`, an
```

uwburn:
> Sorry about the delay. It seems all the maintainers, including myself, are busy with life and work lately.

Sorry, i was not meaning to be harsh, i am about to start a project in the company where i work that would include kafkajs, so i'm looking forward to avoid using git npm dependencies, but that's not really an issue.

Anyway, i followed your suggestion, also i rearranged a bit SASL tests to avoid code duplications.

By looking at it i noted that probably some other conditional tests about 1.1 and 0.11 must be rechecked, if i have some time to spare eventually i will submit another PR about that.

Nevon:
I added a job to the pipeline that runs the Oauthbearer tests. It's configured to only run in master, as it takes more than 5 minutes to search through all tests for the SASL OAUTHBEARER ones. Not great, and could be fixed by reorganizing the tests so that all the oauthbearer ones are in files with a certain suffix or something, but I'm not gonna do that right now. I did run the job through CI once, and it passed.

PR #685, opened 2020-04-02
Fix encoder instanceof issue with Encoder

Nevon:
@bfdill was having this issue where KafkaJS was failing to encode requests because `Encoder.writeEncoder` was throwing, because the value was not an instanceof `Encoder`, even though we could clearly see that it was in fact an encoder. I don't have an explanation for this, but we have seen issues with instanceof before. I'm guessing that his project was doing something strange at build time, since it was a Typescript project (maybe he's accidentally transpiling `node_modules`).

```
Error: value should be an instance of Encoder
        at Encoder.writeEncoder (/Users/user/projects/project/node_modules/kafkajs/src/protocol/encoder.js:177:13)
        at Object.<anonymous>.module.exports (/Users/user/projects/project/node_modules/kafkajs/src/protocol/request.js:10:6)
        at sendRequest (/Users/user/projects/project/node_modules/kafkajs/src/network/connection.js:276:30)
        at Connection.send (/Users/user/projects/project/node_modules/kafkajs/src/network/connection.js:303:53)
        at Broker.apiVersions (/Users/user/projects/project/node_modules/kafkajs/src/broker/index.js:153:20)
        at Broker.connect (/Users/user/projects/project/node_modules/kafkajs/src/broker/index.js:93:25)
        at /Users/user/projects/project/node_modules/kafkajs/src/cluster/brokerPool.js:79:9
        at Cluster.connect (/Users/user/projects/project/node_modules/kafkajs/src/cluster/index.js:104:5)
        at Object.connect (/Users/user/projects/project/node_modules/kafkajs/src/producer/index.js:205:7) {
      [stack]: 'Error: value should be an instance of Encoder\n' +
        '    at Encoder.writeEncoder (/Users/user/projects/project/node_modules/kafkajs/src/protocol/encoder.js:177:13)\n' +
        '    at Object.<anonymous>.module.exports (/Users/user/projects/project/node_modules/kafkajs/src/protocol/request.js:10:6)\n' +
        '    at sendRequest (/Users/user/projects/project/node_modules/kafkajs/src/network/connection.js:276:30)\n' +
        '    at Connection.send (/Users/user/projects/project/node_modules/kafkajs/src/network/connection.js:303:53)\n' +
        '    at Broker.apiVersions (/Users/user/projects/project/node_modules/kafkajs/src/broker/index.js:153:20)\n' +
        '    at Broker.connect (/Users/user/projects/project/node_modules/kafkajs/src/broker/index.js:93:25)\n' +
        '    at /Users/user/projects/project/node_modules/kafkajs/src/cluster/brokerPool.js:79:9\n' +
        '    at Cluster.connect (/Users/user/projects/project/node_modules/kafkajs/src/cluster/index.js:104:5)\n' +
        '    at Object.connect (/Users/user/projects/project/node_modules/kafkajs/src/producer/index.js:205:7)',
      [message]: 'value should be an instance of Encoder'
    }
```
I thought about adding a `name` property to every instance of Encoder, but it feels a bit ugly, so I decided to go for an approach where we instead just check that the value is "Encoder-like", i.e. has a `buffer` property with a Buffer inside it, since that's what we actually care about. Anyway the encoder is only an internal class, so it's mostly there to prevent us from shooting ourselves in the foot.

### DIFF
--- a/src/protocol/encoder.js
+++ b/src/protocol/encoder.js
@@ -56,6 +56,7 @@ module.exports = class Encoder {
 
   constructor() {
     this.buffer = Buffer.alloc(0)
+    this.name = 'Encoder'
   }
 
   writeInt8(value) {
@@ -173,7 +174,7 @@ module.exports = class Encoder {
   }
 
   writeEncoder(value) {
-    if (value instanceof Encoder !== true) {
+    if (value == null || value.name !== 'Encoder') {
       throw new Error('value should be an instance of Encoder')
     }
     this.buffer = Buffer.concat([this.buffer, value.buffer])
--- a/src/protocol/encoder.js
+++ b/src/protocol/encoder.js
@@ -56,7 +56,6 @@ module.exports = class Encoder {
 
   constructor() {
     this.buffer = Buffer.alloc(0)
-    this.name = 'Encoder'
   }
 
   writeInt8(value) {
@@ -174,15 +173,15 @@ module.exports = class Encoder {
   }
 
   writeEncoder(value) {
-    if (value == null || value.name !== 'Encoder') {
+    if (value == null || value.buffer == null) {
       throw new Error('value should be an instance of Encoder')
     }
-    this.buffer = Buffer.concat([this.buffer, value.buffer])
-    return this
+
+    return this.writeBuffer(value.buffer)
   }
 
   writeEncoderArray(value) {
-    if (!Array.isArray(value) || value.some(v => !(v instanceof Encoder))) {
+    if (!Array.isArray(value) || value.some(v => v == null || !Buffer.isBuffer(v.buffer))) {
       throw new Error('all values should be an instance of Encoder[]')
     }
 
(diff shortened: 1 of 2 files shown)


bfdill:
Based on a little bit of additional testing, I'm wondering if this stems from a bug in require?

https://github.com/nodejs/node/issues/13408

Thanks for the help working through this today!

Nevon:
That wouldn't explain why you are the only person that has this problem, though. This should affect literally everyone, since it happens on the very first request we make.

bfdill:
That's true, but it does explain the behavior I'm seeing.  What I'm not sure about is how my setup is forcing this issue.  We are in a monorepo, that project is using typescript, the test is running via ts-node.

When I finish the project I'm on right now, I'll try to see if I can get a minimal reproduction with the current version.

src/protocol/encoder.js
```diff
@@ -173,15 +173,15 @@ module.exports = class Encoder {
   }
 
   writeEncoder(value) {
-    if (value instanceof Encoder !== true) {
+    if (value == null || value.buffer == null) {
```
ankon, line 176: Nitpicking: `==` should be `===`; also below in `writeEncoderArray`.

Nevon: No, it's intentional to check for both null and undefined.

PR #686, opened 2020-04-05
Document the parameters of `writeArray` and `writeNullableArray`

ankon:
This fixes a bunch of warnings in smart IDEs when looking at code (for example in fetch/v7/request.js) that doesn't
provide the type and length parameters.

While there: Drop the case of 'typeof value' returning 'int32': That just cannot happen.


Nevon:
You can remove the same case on line 225.

PR #693, opened 2020-04-07
Use a try-catch block to make the code look similar to similar places

ankon:


src/consumer/runner.js
```diff
@@ -94,7 +94,11 @@ module.exports = class Runner {
       return
     }
 
-    return this.join().catch(this.onCrash)
+    try {
```
Nevon, line 97: These are semantically different. I'm not sure if it matters in this case, but if `this.onCrash` rejects in the previous version, that promise rejection would be available to the caller of `scheduleJoin` to handle. In the new version, if onCrash rejects you have an unhandled promise rejection.

ankon: Hmm, good one!

This is an async function, so if `onCrash` is a regular function and throws that just gets translated into a rejection of `scheduleJoin`. If `onCrash` is an async function/returns a promise, then indeed this now leads to an UnhandledPromiseRejection. From what I see the only `onCrash` ever provided is an async function.
There is also another difference: If `onCrash` would return something, then the previous code would resolve the promise returned by `scheduleJoin` to that, while the try-catch refactoring drops the return value onto the floor. It seems that the only `onCrash` ever provided doesn't do that though.

Another `await` then?

```js
try {
  return await this.join()
} catch (e) {
  await this.onCrash(e)
}
```

Based on your comment I looked closer at what `start` actually does: It does use what looks like the same try-catch-onCrash pattern, except that it seems that it wouldn't call `onCrash` when `scheduleFetch` fails (no `return await`). Looking at that even closer then I noticed that `scheduleFetch` calls `onCrash` already internally, so all-in-all this is hard to read, but it seems that everything is just fine -- maybe a comment here and there would be nice for the reader :)

The initial reason for this PR was though to make code look more similar and use the "modern" syntax, and I'm not so sure that that is actually a good idea anymore: Sticking the `.catch(this.onCrash)` is used in more than one place as well, and the existing try-catch is actually doing quite some "magic".

I'll close this, thanks for the review!

PR #694, opened 2020-04-07
Delete the entry for the waiter when the timeout is reached

ankon:
When the lock times out the promise rejects, and the calling code will do "something sensible". Something sensible
could be to retry later -- but that actually requires the lock to be able to be acquired again. Right now this is
not possible, because the failed `tryToAcquire` is still inside the waiting queue -- and will never release the lock.

### DIFF
--- a/src/utils/lock.js
+++ b/src/utils/lock.js
@@ -39,10 +39,12 @@ module.exports = class Lock {
       }
 
       this[PRIVATE.WAITING].add(tryToAcquire)
-      timeoutId = setTimeout(
-        () => reject(new KafkaJSLockTimeout(this[PRIVATE.TIMEOUT_ERROR_MESSAGE]())),
-        this[PRIVATE.TIMEOUT]
-      )
+      timeoutId = setTimeout(() => {
+        // The message should contain the number of waiters _including_ this one
+        const error = new KafkaJSLockTimeout(this[PRIVATE.TIMEOUT_ERROR_MESSAGE]())
+        this[PRIVATE.WAITING].delete(tryToAcquire)
+        reject(error)
+      }, this[PRIVATE.TIMEOUT])
     })
   }
 
--- a/src/utils/lock.spec.js
+++ b/src/utils/lock.spec.js
@@ -40,6 +40,26 @@ describe('Utils > Lock', () => {
     ).rejects.toHaveProperty('message', 'Timeout while acquiring lock (2 waiting locks)')
   })
 
+  it('allows lock to be acquired after timeout', async () => {
+    const lock = new Lock({ timeout: 60 })
+    const resource = jest.fn()
+    const callResource = async () => {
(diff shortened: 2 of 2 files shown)


ankon:
This should fix one aspect of the problem mentioned in https://github.com/tulios/kafkajs/pull/666#issuecomment-610343946: We should be able to continue, even after a client has hit the timeout. 

I still need to find the reason for timeout, but I think this is independent and might be more caused by what my application code is doing with KafkaJS.

src/utils/lock.js
```diff
@@ -39,10 +39,10 @@ module.exports = class Lock {
       }
 
       this[PRIVATE.WAITING].add(tryToAcquire)
-      timeoutId = setTimeout(
-        () => reject(new KafkaJSLockTimeout(this[PRIVATE.TIMEOUT_ERROR_MESSAGE]())),
-        this[PRIVATE.TIMEOUT]
-      )
+      timeoutId = setTimeout(() => {
+        reject(new KafkaJSLockTimeout(this[PRIVATE.TIMEOUT_ERROR_MESSAGE]()))
```
Nevon, line 43: Does order matter here? When `reject` is called, will any `catch` be invoked synchronously or will it happen in a subsequent tick? I'm thinking if we need to delete the item from the waiting list before rejecting, in case someone tries to reacquire the lock directly in a catch?

ankon: As far as I understand the order doesn't matter. 

But, that said: I also would feel more "comfortable" with having the reject as the last statement -- but that adds one more line that also makes it look a bit more ugly:
```js
      timeoutId = setTimeout(() => {
        const error = new KafkaJSLockTimeout(this[PRIVATE.TIMEOUT_ERROR_MESSAGE]()) 
        this[PRIVATE.WAITING].delete(tryToAcquire)
        reject(error)
      }, this[PRIVATE.TIMEOUT])
```

The reason here is that I didn't want to change the count in the error message -- right now it shows the number of waiters _including the one failing right now_. If you change that to show the waiters _left after this one failed_:
```js
      timeoutId = setTimeout(() => {
        this[PRIVATE.WAITING].delete(tryToAcquire)
        reject(new KafkaJSLockTimeout(this[PRIVATE.TIMEOUT_ERROR_MESSAGE]()))
      }, this[PRIVATE.TIMEOUT])
```

(Or of course just bump the count in the error message helper function, but that then would require a comment -- which is a bit "meh")

Nevon: I think the first option is fine. It is a bit unintuitive since there's nothing to indicate that the error will contain a reference to the waiting list, but it's fine.

ankon:
> I think the first option is fine. It is a bit unintuitive since there's nothing to indicate that the error will contain a reference to the waiting list, but it's fine.

Changed. I added a comment as well, so that readers of the code won't be too confused :)

Issue #465, opened 2019-08-14
Add ACL functions to admin client


Brazier85:
Hello, following up from #111. Can you add the following functions to the admin client?
- CreateAcls
- DeleteAcls
- DescribeAcls

This would really help me administrating my secure kafka cluster.

shubhanilBag:
I have started looking into the issue 🔍
@tulios

aakrem:
+1

terios:
+1

tulios:
This feature was merged and is finally included in the latest pre-release `1.15.0-beta-8`.

Issue #649, opened 2020-03-03
Allow describe/alter config for broker configs


jlandersen:
Currently the `describeConfigs`and `alterConfigs` methods on admin client fails for broker configs when the cluster has more than one broker. Config requests for topic resources can be served by any broker, but broker config requests must be served by the specific broker.

Example test that will break (unless you are lucky and the request actually goes to the broker being requested):
```javascript
test('describe broker config', async () => {
  const cluster = createCluster()
  admin = createAdmin({ cluster: cluster, logger: newLogger() })
  await admin.connect()
  const metadata = await cluster.brokerPool.seedBroker.metadata()
  const brokers = metadata.brokers
  const brokerToDescribeConfig = brokers[1].nodeId.toString()

  await admin.describeConfigs({
    resources: [
      {
        type: RESOURCE_TYPES.CLUSTER,
        name: brokerToDescribeConfig,
      },
    ],
  })
})
```

- Change `describeConfigs` and `alterConfigs` to split requests up and send any requests for resource type broker to the specific broker. The Java client does this as well [0].

 - Current config related methods on admin client uses `RESOURCE_TYPE` [1] to facilitate requests. Unfortunately this does not represents configs correctly. The linked resource type in Java is ACL related, not for configs. The correct resource type values for configs are available in: https://github.com/apache/kafka/blob/a99e0111114d1cb8c762494ac195cf84e6425bb3/clients/src/main/java/org/apache/kafka/common/config/ConfigResource.java#L36. In the above code example, I would have expected something like `RESOURCE_TYPES.BROKER` as type value.

Possible solutions would be to either rework the current resource type object to match that of ConfigResource, or create a new ConfigResourceType module and use that for all config related methods. This is potentially a breaking change in both cases, especially for Typescript users, as a different type is introduced. Alternatively, keep RESOURCE_TYPE and accept the deviation from the Java client.
 
I am happy to make the necessary changes for this to work, especially the first part to get the feature working, but input on handling the second point would be good. My preference is to introduce the new ConfigResourceType module, and keep the existing resource type for any future functionality where it is needed.

[0] https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L1891

[1] https://github.com/tulios/kafkajs/blob/master/src/protocol/resourceTypes.js

Nevon:
I suggest splitting this into two efforts. The first one seems like a straight up bug that I would be happy to review a PR for.

[...]

PR #697, opened 2020-04-12
Add ACL functions to admin client

shubhanilBag:
Resolves issue: https://github.com/tulios/kafkajs/issues/465


src/admin/__tests__/createAcls.spec.js
```diff
@@ -0,0 +1,228 @@
+const createAdmin = require('../index')
+// const { KafkaJSProtocolError } = require('../../errors')
```
Nevon, line 2: Delete these unused imports.

src/admin/__tests__/createAcls.spec.js
```diff
@@ -0,0 +1,228 @@
...
+  const topicName = `test-topic-${secureRandom()}`
+  let admin
+
+  afterEach(async () => {
+    await admin.disconnect()
+  })
+
+  describe('createAcls', () => {
+    test('create new topic and apply acl', async () => {
+      const saslConnectionOpts = () =>
```
Nevon, line 28: No need to assign a new connection options object here. Just pass in `saslConnectionOpts()` from testHelpers into `createCluster` directly.  Same as in https://github.com/tulios/kafkajs/blob/master/src/admin/__tests__/connection.spec.js#L28-L32 for example.

src/admin/__tests__/createAcls.spec.js
```diff
@@ -0,0 +1,228 @@
...
+
+const {
+  secureRandom,
+  createCluster,
+  newLogger,
+  sslConnectionOpts,
+  saslBrokers,
+} = require('testHelpers')
+
+const RESOURCE_TYPES = require('../../protocol/resourceTypes')
```
Nevon, line 13: See https://github.com/tulios/kafkajs/issues/649#issuecomment-597131598 for some issues regarding `RESOURCE_TYPES`.

Basically, we have been misusing ResourceTypes for configs, when they are supposed to be used in the way that you are here, for ACLs. So before we start having more people use ResourceTypes, I would like to figure out a good way to split the two so that we have `CONFIG_RESOURCE_TYPES` and `ACL_RESOURCE_TYPES`.

The old `RESOURCE_TYPES` still needs to be exported for backwards compatibility until we decide to release a major version, but it would be good if we can prepare by having most people already now import either CONFIG_RESOURCE_TYPES or ACL_RESOURCE_TYPES, so that the break affects fewer people.

So what I would suggest is that we within this PR introduce `ACL_RESOURCE_TYPES`, which is just the old `RESOURCE_TYPES` re-exported under a different name, and use that in all examples and tests.

Nevon: In fact, we should probably prefix all the ACL-related protocol types with `ACL_` just to avoid this in the future.

src/admin/__tests__/createAcls.spec.js
```diff
@@ -0,0 +1,228 @@
...
+        cluster: createCluster(saslConnectionOpts(), saslBrokers()),
+        logger: newLogger(),
+      })
+
+      await admin.connect()
+
+      await expect(admin.getTopicMetadata({ topics: [topicName] })).resolves.toBeTruthy()
+    })
+
+    test('Delete ACL', async () => {
```
Nevon, line 164: This test should not depend on the other tests having run, since the order is not guaranteed and it should still pass if run with `.only`. If you need an ACL to be present before deleting it, create it in the beginning of the test.

src/admin/index.js
```diff
@@ -7,6 +7,9 @@ const { events, wrap: wrapEvent, unwrap: unwrapEvent } = require('./instrumentat
 const { LEVELS } = require('../loggers')
 const { KafkaJSNonRetriableError, KafkaJSDeleteGroupsError } = require('../errors')
 const RESOURCE_TYPES = require('../protocol/resourceTypes')
+// const OPERATION_TYPES = require('../protocol/operationsTypes')
```
Nevon, line 10: Delete

src/admin/index.js
```diff
@@ -772,6 +775,280 @@ module.exports = ({
...
 
+  /**
+   * @param {array} ACL
+   * @return {Promise<void>}
+   */
+  const createAcls = async ({ acl }) => {
+    if (!acl || !Array.isArray(acl)) {
+      throw new KafkaJSNonRetriableError(`Invalid ACL array ${acl}`)
+    }
+    if (acl.length === 0) {
```
Nevon, line 786: Should we really throw on this? I think in other situations where the user basically calls us with a no-op (create 0 items) we just resolve.

src/admin/index.js
```diff
@@ -772,6 +775,280 @@ module.exports = ({
...
+      throw new KafkaJSNonRetriableError(`Empty ACL array`)
+    }
+
+    // Validate principal
+    if (acl.filter(({ principal }) => typeof principal !== 'string').length > 0) {
```
Nevon, line 791: A simpler way of expressing this is `acl.some(({ principal }) => typeof principal !== 'string')`. Same goes for the other validations.

src/admin/index.js
```diff
@@ -772,6 +775,280 @@ module.exports = ({
...
+      throw new KafkaJSNonRetriableError('Invalid ACL array, the hosts have to be a valid string')
+    }
+    // Validate resourceName
+    if (acl.filter(({ resourceName }) => typeof resourceName !== 'string').length > 0) {
+      throw new KafkaJSNonRetriableError(
+        'Invalid ACL array, the resourceNames have to be a valid string'
+      )
+    }
+    // Validate operation
+    if (acl.filter(({ operation }) => typeof operation !== 'number').length > 0) {
```
Nevon, line 807: Rather than checking type, can we check if it is a member of the operations? Same for resourcePatternType, permissionType and resourceType. It's quite hard for the user to understand what is a valid operation if all they are told is that it has to be a number, rather than telling them: "Operation should be one of [1,2,3]"

src/broker/index.js
```diff
@@ -703,6 +703,77 @@ module.exports = class Broker {
...
+   * @returns {Promise<void>}
+   */
+  async createAcls({ acl }) {
+    const createAcls = this.lookupRequest(apiKeys.CreateAcls, requests.CreateAcls)
+    return await this.connection.send(createAcls({ creations: acl }))
+  }
+
+  /**
+   * @public
+   * @param {number} resourceType The type of resource
```
Nevon, line 729: Rather than using primitive types here, declare an enum type in the corresponding protocol file and reference that here. For example:

```js
@param {AclOperation} The type of ACL operation
```

src/protocol/requests/deleteAcls/v0/response.js
```diff
@@ -0,0 +1,73 @@
...
+  const filterResponses = decoder.readArray(decodeFilterResponse)
+
+  return {
+    throttleTime,
+    filterResponses,
+  }
+}
+
+const parse = async data => {
+  console.debug('%o', data)
```
Nevon, line 50: Some debug stuff left over.

src/protocol/requests/deleteAcls/v0/response.js
```diff
@@ -0,0 +1,73 @@
...
+  const filterResponsesWithError = data.filterResponses.filter(({ errorCode }) =>
+    failure(errorCode)
+  )
+  if (filterResponsesWithError.length > 0) {
+    throw createErrorFromCode(filterResponsesWithError[0].errorCode)
+  }
+  for (let i = 0; i < data.filterResponses.length; i++) {
```
Nevon, line 57: We try to avoid numeric for loops unless necessary. A simple `for (let filterResponse of data.filterResponses)` would be preferred.

shubhanilBag:
@Nevon Thanks for the thorough review! I opened the draft PR to check if the basic ACL test is working in the CI. I am still working to on it to make it ready for review, that's includes code cleanup, separating the test case into separate files, more proper validation checks in the admin client and others including your suggestions :smile:

src/admin/__tests__/createAcls.spec.js
shubhanilBag: @Nevon  SGTM, let's do that, also I would suggest to keep types in a separate folder like: `src/protocol/types/ACL/Resource.js` or `src/protocol/types/ACLResources.js`

iwt-janbrauer:
What's the situation here? Is this PR ready to be merged?

EricMCornelius:
Anything I can do to assist in getting this feature moving along?

tulios:
Hey @EricMCornelius, I will take over this PR soon and make sure it gets merged. We had other priorities but now it's the time to get this one in.

tulios:
Step 1 to get this to master (#898)

tulios (approved):

tulios:
@Nevon take a final look, it's all good for me now.

tulios:
I forgot about the Typescript types, that's still missing. I can add it later.

src/protocol/requests/createAcls/v1/response.js
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * CreateAcls Response (Version: 1) => throttle_time_ms [creation_responses]
+ *   throttle_time_ms => INT32
+ *   creation_responses => error_code error_message
+ *     error_code => INT16
+ *     error_message => NULLABLE_STRING
+ */
+
+module.exports = {
+  decode,
```
Nevon, line 12: v1 of all the `*acl` APIs return the responses before throttling, meaning that we should support client side throttling. That's already been implemented on the network layer, so all you have to do on the protocol level is to copy the `throttleTime` property to a `clientSideThrottleTime` property on the decoded object and the rest is handled automatically (see https://github.com/tulios/kafkajs/blob/900c60a7473d16b7f678be4284ed95f21587b789/src/protocol/requests/heartbeat/v2/response.js#L11-L18 for example)

tulios (changes requested):
The PR is also missing protocol and broker tests, I will add those before I merge.

Issue #703, opened 2020-04-19
Test scram credentials shell script shadows the 'username' variable causing tests to fail


galesky:
My goal was to run the tests locally so I could work on some issues, I followed the steps as mentioned here: [https://kafka.js.org/docs/contribution-guide](https://kafka.js.org/docs/contribution-guide), e.g.:

```
./scripts/dockerComposeUp.sh
./scripts/createScramCredentials.sh
yarn test:local:watch
```
When running `./scripts/createScramCredentials.sh` I expected the user `testscram` to be created, but I've actually got `root` (as my docker daemon was running with `sudo`) : 
```
zookeeper    | [2020-04-19 16:27:11,138] INFO Successfully authenticated client: authenticationID=test;  authorizationID=test. (org.apache.zookeeper.server.auth.SaslServerCallbackHandler)
zookeeper    | [2020-04-19 16:27:11,138] INFO Setting authorizedID: test (org.apache.zookeeper.server.auth.SaslServerCallbackHandler)
zookeeper    | [2020-04-19 16:27:11,139] INFO adding SASL authorization for authorizationID: test (org.apache.zookeeper.server.ZooKeeperServer)
kafka3       | [2020-04-19 16:27:11,544] INFO Processing notification(s) to /config/changes (kafka.common.ZkNodeChangeNotificationListener)
kafka3       | [2020-04-19 16:27:11,549] INFO Processing override for entityPath: users/root with config: Map(SCRAM-SHA-256 -> [hidden], SCRAM-SHA-512 -> [hidden]) (kafka.server.DynamicConfigManager)
kafka1       | [2020-04-19 16:27:11,550] INFO Processing notification(s) to /config/changes (kafka.common.ZkNodeChangeNotificationListener)
kafka3       | [2020-04-19 16:27:11,550] INFO Removing PRODUCE quota for user root (kafka.server.ClientQuotaManager)
kafka3       | [2020-04-19 16:27:11,550] INFO Removing FETCH quota for user root (kafka.server.ClientQuotaManager)
```

This, of course, causes any test that relies on SCRAM authentication to fail, as the user credentials expect the user definitions as set up on `testHelpers`

Seems to me that this was caused by shadowing the `USERNAME` variable on the `createScramCredentials` shell script.

I did some testing renaming the `USERNAME` variable, which fixed the issue, and will file a PR to exemplify;

My current environment is:
```
Docker version 19.03.8, build afacb8b7f0
Node v13.5.0
[...]

PR #704, opened 2020-04-19
fix(test): fixed username shadowing in test scram setup

galesky:
## Issue
https://github.com/tulios/kafkajs/issues/703

In addition to fixing the related issue, I've also included user feedback in the execution which can help to detect bugs such as this one.

### DIFF
--- a/docs/DevelopmentEnvironment.md
+++ b/docs/DevelopmentEnvironment.md
@@ -36,7 +36,7 @@ kafkajs_kafka3_1   start-kafka.sh                   Up      0.0.0.0:9098->9098/t
 kafkajs_zk_1       /bin/sh -c /usr/sbin/sshd  ...   Up      0.0.0.0:2181->2181/tcp, 22/tcp, 2888/tcp, 3888/tcp
 ```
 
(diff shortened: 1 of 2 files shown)


Nevon:
Thanks for looking into the issue. How about instead of relying on environment variables, we move to taking the username and password as parameters that are passed in when you execute the script? `./createScramCredentials foo bar`

We can still have default values, same as now, but not using environment variables would remove the risk of someone running into the same issues as you are now.

galesky:
That's a good idea, kinda moves in a way so it can be implemented as part of a CLI in the future.

I'll take a look at updating `./createScramCredentials` later today;

galesky:
Hi @Nevon , I've added some flags, basic docs and behaviors; 

Just want to check with you if that aligned to what you were expecting if so I'll also update the docs at least to indicate that those new commands are available
 
output

Nevon:
That's quite a bit more advanced than what I had in mind, but sure, it seems to work great.

galesky:
Yeah, I tried to stay as lean as possible so as to handle only the `username` and `password` but while also doing some checks and falling back to default values. 

I've updated the docs now to reference the new controls and it seems ready to me.

Issue #702, opened 2020-04-19
KafkaJS has UnhandledPromiseRejection somewhere in the code


goriunov:
Was running stress tests on the kafkajs and picked up this issue (error does not matter most important part is UnhandledPromiseRejectionWarning):

```
(node:78646) UnhandledPromiseRejectionWarning: KafkaJSProtocolError: The coordinator is not aware of this member
    at createErrorFromCode (/Users/dmitrii/Desktop/Projects/at_connector_common/node_modules/kafkajs/src/protocol/error.js:537:10)
    at Object.parse (/Users/dmitrii/Desktop/Projects/at_connector_common/node_modules/kafkajs/src/protocol/requests/heartbeat/v0/response.js:20:11)
    at Connection.send (/Users/dmitrii/Desktop/Projects/at_connector_common/node_modules/kafkajs/src/network/connection.js:317:35)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:78646) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2017)
```

Looks like `heartbeat` has unhandled promise or throwing inside async function somewhere

Was running latest `BETA` version  ( "kafkajs": "1.13.0-beta.15")

goriunov:
You should be able to easily reproduce that by changing `heartbeat/v0/response.json` to always throw and run consumer.

```js
const parse = async data => {
  // if (failure(data.errorCode)) {
  throw createErrorFromCode()
  // }
  return data
}
```

Looks like it may be related to reties:

```
Passed here
[34mdebug[39m: Suppressed error:  {"timestamp":"2020-04-19T21:17:58.870Z","logger":"kafkajs","message":"Response Heartbeat(key: 12, version: 1)","broker":"xyz:9092","clientId":"kafkajs-87503","error":"Unknown error code undefined","correlationId":5,"size":10}
[34mdebug[39m: {"timestamp":"2020-04-19T21:17:58.870Z","logger":"kafkajs","message":"Response Heartbeat(key: 12, version: 1)","broker":"xyz:9092","clientId":"kafkajs-87503","error":"Unknown error code undefined","correlationId":5,"payload":{"type":"Buffer","data":"[filtered]"}}
Passed here
[...]

PR #705, opened 2020-04-21
Fix for unhandled rejections (simpler runner)

goriunov:
Fix for https://github.com/tulios/kafkajs/issues/702

@Nevon I think this is good solution for the issues above. After adding this code, was able to run chaos tests without having unhandled rejections yet.


Nevon:
I would like to get @tulios input on this, as I'm not very familiar with how this part of the application works anymore or what the purpose of the locking here is.

goriunov:
@Nevon i have found that lock has some issues in this context, i am preparing new changes to fix up that hopefully within next few days should have properly working solution....

ankon:
Interesting. So if I understand correctly these unhandled rejections come from a situation where an exception triggers the rejection (`unlockWithError`), but then no one is waiting for the lock? As far as I can tell in that case we must have had batches, and therefore the number of expected executions should be >0 though? I'm probably missing something, but indeed this area looks quite fragile.

As it may help a bit: https://github.com/tulios/kafkajs/pull/671 has a side-effect of at least removing one level of promise-waiting from this function (no more "enqueued tasks"), so maybe it helps to take that in and see if stuff becomes clearer?

goriunov:
@ankon no, based on the current code we would have never reached await lock as exception would have been thrown in the

```
await Promise.all(enqueuedTasks.map(fn => fn()))
```

as it was awaited before lock and we actually calling `unlockWithError` in `enqueuedTasks` so if one of the `enqueuedTasks` takes long to process and another one at the same time throws error we will get unhandled rejection. I have tested this scenarios on my local.

@Nevon @tulios i have wrapped up my change and simplified runner. Looks like there was no need in having lock as we can flat enqueuedTasks queue and await all. I have done chaos tests on my local no more unhandled rejections.

src/consumer/runner.js
```diff
@@ -344,27 +340,19 @@ module.exports = class Runner {
...
+
     if (!this.running) {
       return
     }
 
-    await Promise.all(enqueuedTasks.map(fn => fn()))
-    if (expectedNumberOfExecutions > 0) {
-      await lock
-    }
+    await Promise.all([].concat(...enqueuedTasks.map(fn => fn())))
```
WaleedAshraf, line 355: Aren't you awaiting for this already at line 349?

goriunov: In line 349 we are creating an array `enqueuedTasks` we need to await for allResults array to make sure we fetch all `const batches = await result.value` without errors in case if it fails it will throw up the chain. In this line we are actually awaiting every single batch process

goriunov, line 355: I am using `[].concat(...enqueuedTasks)` to flatten `enqueuedTasks` as enqueuedTasks looks like [[batch, batch, batch], [batch, batch]] above one will flatten it and we will await for every single batch to complete of cause taking in account `concurrently` logic

goriunov:
@tulios @Nevon Have done more chaos testing and running this branch in our test environment for the last few days. No issues yet observed !

tulios:
I will review this today, sorry for the late reply @goriunov

tulios:
@goriunov I think this fixes the issue, but it defeats the purpose of the original change. With the change, the consumer will wait for all requests before it starts processing the batches, which was the situation before, that's why you are able not to check the number of executions and remote the lock. I will look at the issue and propose a change.

goriunov:
@tulios May be i am missing something but from my local running behaviour of the code has not actually changed...

goriunov:
@tulios but we need to wait other wise `expectedNumberOfExecutions` count will be wrong and lock may get unlocked early...

tulios:
Maybe `unLockWithError` was wrong, and breaking the loop too early, I will look into that. The lock was introduced to allow the loop to unpack and calculate the `expectedNumberOfExecutions`. To test this locally you need several partitions in the topic, I will run the code in the branch and try to confirm what I said.

goriunov:
@tulios i see the point now... yeah need to think around `unlockWithError`....

goriunov:
Move in to #714

PR #714, opened 2020-04-29
Fix unhandled rejections

goriunov:
This is slightly different approach to fix unhandled rejections (first PR is in https://github.com/tulios/kafkajs/pull/705), creating separate PR to run tests agains this branch and verify that everything works as expected. Also i have added `requestsCompleted` to make sure that we do not unlock too early... 

@tulios would like to hear your opinion about this approach.

I have not yet done tests from my side will ping in this PR when everything is checked.

### DIFF
--- a/src/consumer/__tests__/runner.spec.js
+++ b/src/consumer/__tests__/runner.spec.js
@@ -274,6 +274,64 @@ describe('Consumer > Runner', () => {
       expect(consumerGroup.join).toHaveBeenCalledTimes(1)
     })
 
+    it('correctly catch exceptions in parallel "eachBatch" processing', async () => {
+      runner = new Runner({
+        consumerGroup,
+        instrumentationEmitter: new InstrumentationEventEmitter(),
+        eachBatchAutoResolve: false,
+        eachBatch: async () => {
+          throw new Error('Error while processing batches in parallel')
+        },
+        onCrash,
+        logger: newLogger(),
+        partitionsConsumedConcurrently: 10,
+      })
+
+      const batch = new Batch(topicName, 0, {
+        partition,
+        highWatermark: 5,
+        messages: [{ offset: 4, key: '1', value: '2' }],
+      })
+
+      const longRunningRequest = new Promise(resolve => {
+        setTimeout(() => resolve([]), 100)
+      })
+
+      consumerGroup.fetch.mockImplementationOnce(() =>
+        BufferedAsyncIterator([longRunningRequest, Promise.resolve([batch])])
+      )
+
+      runner.scheduleFetch = jest.fn()
+      await runner.start()
+
+      await expect(runner.fetch()).rejects.toThrow('Error while processing batches in parallel')
+    })
+
+    it('correctly catch exceptions in parallel "heartbeat" processing', async () => {
+      const batch = new Batch(topicName, 0, {
+        partition,
+        highWatermark: 5,
+        messages: [{ offset: 4, key: '1', value: '2' }],
+      })
+
+      const longRunningRequest = new Promise(resolve => {
+        setTimeout(() => resolve([]), 100)
+      })
+
+      consumerGroup.heartbeat = async () => {
+        throw new Error('Error while processing heartbeats in parallel')
+      }
+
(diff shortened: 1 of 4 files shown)


goriunov:
@tulios @Nevon have run some tests on my local and everything looks good. I think we can close #705 and process with this branch. 

Change is basically that lock promise does not actually throwing an error but stores that in resolved and when we are awaiting for lock it will return first resolved value which could be error and we can easily propagate it up. @tulios what do you thing about this approach ?

goriunov:
@tulios have been running this branch in our test env for the last week everything looks pretty good. Hope you will have some time to check it out :)

ankon:
@Nevon @tulios what's the next steps here? Anything we can do to help?

Nevon:
I barely understood this code a year ago, and now after having barely looked at it for quite a while, I understand even less. 😔

@goriunov: Do you think you could walk me through why there was an unhandled rejection in the first place, and how this change solves it? To me, it looks strange that we have a promise where we resolve with an error, only to immediately re-throw the error, so clearly I'm missing something.

ankon:
From what I understand here is for the unhandled rejection to occur you need the processing in `fetch` to

1) produce an error (either in the batch handler, or in some code like the heartbeat function), *and*
2) do that before all pending batches have been enqueued and the lock has actually been `await`-ed. This condition is the harder one to hit it seems, but having either many batches pending (many subscribed topics for the group) and/or having slow brokers and/or startup situations where additional metadata updates are needed to work out what to request is what would easily be making these appear in our own tests.

In case both conditions are met the `unlockWithError` will reject the promise, but because no one is `await`-ing we end up (rightfully) with the unhandled rejection.

The fix essentially changes the logic so that the lock will _resolve_ to either `undefined` when all was good, or to the _error_ if one was caught via `unlockWithError` so that there is no chance ever for a rejection in the first place, *and* then modifies the flow such that the unlocking may happen later. @goriunov do you remember why you introduced the `requestsCompleted` part -- from reading this I somehow think that it is actually not needed now.

ankon:
> I barely understood this code a year ago, and now after having barely looked at it for quite a while, I understand even less. pensive

A couple of commenting lines might indeed be in order here, it looks very fragile right now. Ideally of course we could get rid of the lock thing in the first place ... 

I'm right now working on bringing #671 back, which at least will remove one level of promise-magic in a related area, and reduce the indentation level needed here as well.

ankon:
> 1. produce an error (either in the batch handler, or in some code like the heartbeat function), and
> 2. do that before all pending batches have been enqueued and the lock has actually been await-ed. 

I'm wondering whether this would make it testable: Create a `eachBatch` function that throws on the first call, and delays "for a long time" in the second call. Then make the test simulate at least two batches. Expectation would be that this throws in the call to `fetch` (and presumably triggers a crash instrumentation event).

goriunov:
@ankon you are right for most part except of  `requestsCompleted`. This one is needed mainly for `if (requestsCompleted && numberOfExecutions === expectedNumberOfExecutions)` in concurrently part. So basic example:  You have 2 requests going out and now awaiting for both of them to process in `Promise.all` then first one get processed from this point  `expectedNumberOfExecutions` increased based on the first request  then it will go inside of `batches.map(batch =>`  and if for some reason second request will not complete before all batches has been process on the first one it would resolve promise as `expectedNumberOfExecutions` would maximum be equal to the numbers of batches on the first request. `requestsCompleted` makes sure we do not resolve promise from concurrent batches logic before all requests have been processed and correct `expectedNumberOfExecutions` is set.

Nevon:
> I'm wondering whether this would make it testable:

Would either of you be up for giving this a shot? Although the explanation makes sense, I'm hesitant to merge something that I don't feel confident about understanding completely while *also* not having any test to reproduce the issue. I understand that the issue might not be easy to reproduce, but I would be concerned about accidentally introducing it again in the future.

goriunov:
@Nevon i have added some basic tests which expose this unhanded rejection on old branch and on this one  those errors will be caught correctly.

Nevon:
Great! That makes it a lot easier to merge, since now we'll know if we introduce this issue again in the future.

PR #718, opened 2020-05-06
Add listTopics to Admin

dashie:


src/admin/index.js
```diff
@@ -69,6 +69,16 @@ module.exports = ({
     instrumentationEmitter.emit(DISCONNECT)
   }
 
+  /**
+   * @return {Promise}
+   */
+  const listTopics = async () => {
```
Nevon, line 75: Could you please add a test for this change as well?

src/admin/index.js
```diff
@@ -69,6 +69,16 @@ module.exports = ({
...
+    await cluster.refreshMetadata()
```
Nevon, line 76: This should not be necessary. `cluster.metadata` will refresh if it needs to. `refreshMetadata` will also just refresh metadata for topics it knows about, so it wouldn't be able to discover new topics anyway.

src/admin/index.js
```diff
@@ -69,6 +69,16 @@ module.exports = ({
...
+    const { topicMetadata } = await cluster.metadata({ topics: [] })
```
Nevon, line 77: `{ topics: []}` is already the default behavior for `cluster.metadata`, so no need to pass it in.

Nevon:
Thank you for the contribution!

Issue #709, opened 2020-04-24
Consumer throws KafkaJSNumberOfRetriesExceeded after 5 retries when default is 10


WaleedAshraf:
**Describe the bug**
Consumer throws KafkaJSNumberOfRetriesExceeded after 5 retries where the default is 10 as here:
https://kafka.js.org/docs/configuration#a-name-retry-a-default-retry

**To Reproduce**
- Start consumer without providing retry config (let it use default 10)
- Throw error 5 times while consuming
- The consumer will restart with KafkaJSNumberOfRetriesExceeded

**Expected behavior**
The consumer should retry 10 by d times by default.

**Observed behavior**
![image](https://user-images.githubusercontent.com/8335457/80212835-83751380-8638-11ea-889d-6c79821a7e39.png)

**Environment:**
 - OS: Mac OS 10.15.4
 - KafkaJS version: 1.12.0
 - Kafka version: 2.4.0
 - NodeJS version: 10.16.3

**Additional context**
It's happening because we send empty object here,
https://github.com/tulios/kafkajs/blob/aec71606610b39bbdb9afc5a6610b40c840b3120/src/index.js#L137-L140

And inside the consumer, the default value is not assigned because we received the value as an empty object.
https://github.com/tulios/kafkajs/blob/aec71606610b39bbdb9afc5a6610b40c840b3120/src/consumer/index.js#L37-L39

tulios:
Good catch, can you attempt a PR?

WaleedAshraf:
@tulios Yes. Will do that 👍

PR #719, opened 2020-05-07
fix default retry for consumer

WaleedAshraf:
closes https://github.com/tulios/kafkajs/issues/709

### DIFF
--- a/src/admin/index.js
+++ b/src/admin/index.js
@@ -47,7 +47,7 @@ const findTopicPartitions = async (cluster, topic) => {
 module.exports = ({
   logger: rootLogger,
   instrumentationEmitter: rootInstrumentationEmitter,
-  retry = { retries: 5 },
+  retry,
   cluster,
 }) => {
   const logger = rootLogger.namespace('Admin')
--- a/src/consumer/index.js
+++ b/src/consumer/index.js
@@ -26,6 +26,7 @@ const specialOffsets = [
 module.exports = ({
   cluster,
   groupId,
+  retry,
   logger: rootLogger,
   partitionAssigners = [roundRobin],
   sessionTimeout = 30000,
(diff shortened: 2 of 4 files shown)


src/admin/index.js
```diff
@@ -47,7 +47,7 @@ const findTopicPartitions = async (cluster, topic) => {
 module.exports = ({
   logger: rootLogger,
   instrumentationEmitter: rootInstrumentationEmitter,
-  retry = { retries: 5 },
+  retry,
```
Nevon, line 50: This change looks a bit suspicious. Am I correct in assuming that `createRetry` will fall back to the default, so that the default here was pointless?

WaleedAshraf: Right,

As per documentation for [Admin Client](https://kafka.js.org/docs/admin), Admin Client doesn't provide any default value for itself. It uses the default value provided by the [Client Configuration](https://kafka.js.org/docs/configuration#a-name-retry-a-default-retry).
So, having a default here doesn't make sense.

Nevon:
Thanks for the fix! 👏

WaleedAshraf:
@Nevon I forgot to discuss something,
With this change, we are updating the retries from 5 to 10. Can that have an impact on the users? Should we change the default value to 5?

Nevon:
Oh, yes, good call. I missed that. We should definitely not change that behavior. Will you prepare a PR changing the default back to what it was?

WaleedAshraf:
I can do that,
Just to be clear, because of the bug, the default was 5. according to docs it was 10.
Now after fixing the bug, we are making default value 5 so that there's no impact for users.

Nevon:
I think that is preferable, yes. Even if the docs said 10, people have been running with 5, so I think it's better we update the docs to reflect that the default is 5.

PR #720, opened 2020-05-07
update consumer default retries to 5

WaleedAshraf:
relates to https://github.com/tulios/kafkajs/issues/709
Because of the bug, default retires was always 5. We have fixed the bug in https://github.com/tulios/kafkajs/pull/719 and now updating default value to 5 from 10, so that it doesn't affect the users.


WaleedAshraf:
Trying to figure out what went wrong.

Nevon:
I'm pretty sure that's the pipeline being flaky. I'm re-running the failed job now.

PR #721, opened 2020-05-08
improved docs - fixing typos

WaleedAshraf:
Consumer docs are missing some params in the example. I have also added ~async~ `: Promise<void>` with methods that return Promise. Let me know if that is okay.
Also fixed some typos.

### DIFF
--- a/docs/Consuming.md
+++ b/docs/Consuming.md
@@ -94,10 +94,10 @@ await consumer.run({
 * `eachBatchAutoResolve` configures auto-resolve of batch processing. If set to true, KafkaJS will automatically commit the last offset of the batch if `eachBatch` doesn't throw an error. Default: true.
 * `batch.highWatermark` is the last committed offset within the topic partition. It can be useful for calculating lag.
 * `resolveOffset()` is used to mark a message in the batch as processed. In case of errors, the consumer will automatically commit the resolved offsets.
-* `async heartbeat()` can be used to send heartbeat to the broker according to the set `heartbeatInterval` value in consumer [configuration](#options).
-* `async commitOffsetsIfNecessary(offsets?)` is used to commit offsets based on the autoCommit configurations (`autoCommitInterval` and `autoCommitThreshold`). Note that auto commit won't happen in `eachBatch` if `commitOffsetsIfNecessary` is not invoked. Take a look at [autoCommit](#auto-commit) for more information.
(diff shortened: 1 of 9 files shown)


docs/Consuming.md
```diff
@@ -55,13 +55,22 @@ await consumer.run({
 
 ## <a name="each-batch"></a> eachBatch
 
-Some use cases require dealing with batches directly. This handler will feed your function batches and provide some utility functions to give your code more flexibility: `resolveOffset`, `heartbeat`, `isRunning`, and `commitOffsetsIfNecessary`. All resolved offsets will be automatically committed after the function is executed.
+Some use cases require dealing with batches directly. This handler will feed your function batches and provide some utility functions to give your code more flexibility: `resolveOffset`, `heartbeat`, `commitOffsetsIfNecessary`, `uncommittedOffsets`, `isRunning`, and `isStale`. All resolved offsets will be automatically committed after the function is executed.
```
Nevon, line 58: `uncommittedOffsets` isn't a thing, or did I miss something? :D

Nevon, line 58: I had completely forgotten that `uncommittedOffsets` is a thing 😅

docs/Consuming.md
```diff
@@ -82,11 +91,13 @@ await consumer.run({
 })
 ```
 
-* `batch.highWatermark` is the last committed offset within the topic partition. It can be useful for calculating lag.
 * `eachBatchAutoResolve` configures auto-resolve of batch processing. If set to true, KafkaJS will automatically commit the last offset of the batch if `eachBatch` doesn't throw an error. Default: true.
+* `batch.highWatermark` is the last committed offset within the topic partition. It can be useful for calculating lag.
 * `resolveOffset()` is used to mark a message in the batch as processed. In case of errors, the consumer will automatically commit the resolved offsets.
-* `commitOffsetsIfNecessary(offsets?)` is used to commit offsets based on the autoCommit configurations (`autoCommitInterval` and `autoCommitThreshold`). Note that auto commit won't happen in `eachBatch` if `commitOffsetsIfNecessary` is not invoked. Take a look at [autoCommit](#auto-commit) for more information.
+* `async heartbeat()` can be used to send heartbeat to the broker according to the set `heartbeatInterval` value in consumer [configuration](#options).
```
Nevon, line 97: Maybe instead of `async` we should replace this with a Typescript-esque type definition, since that's a standard way of expressing this ex. `heartbeat(): Promise<void>`

WaleedAshraf: done 👍

PR #724, opened 2020-05-13
fix(consumer): Runner#waitForConsumer uses consuming stop event instead of timers

### LINKED ISSUE #723: Runner#stop takes too much time because of hardcoded 1000ms timer
As you can see at [runner source](https://github.com/sempasha/kafkajs/blob/c51dda6fd3d1626535cd7a44e763426e317f2d3a/src/consumer/runner.js#L144), call of Runner#stop first change `this.running` state, then wait consumer to be done and then leave consumer group.

When runner wait consumer to be done it check `this.consuming` flag is set or not. When flag is not set - we done, but when flag is not set, next flag check will be scheduled one second later.

So every time we need to stop runner (that mean every time we need to stop consumer) we may get unnecessary one second delay, caused by waiter based on `setTimeout`.

```js
  async stop() {
    if (!this.running) {
      return
    }
    ...
    this.running = false

    try {
      if (!isTestMode) {
        await this.waitForConsumer()
      }
      await this.consumerGroup.leave()
    } catch (e) {}
  }

  waitForConsumer() {
    return new Promise(resolve => {
      const scheduleWait = () => {
        ...
        setTimeout(() => (!this.consuming ? resolve() : scheduleWait()), 1000)
      }
      if (!this.consuming) {
        return resolve()
      }
      scheduleWait()
    })
  }
```

If runner could emit events on each change of `this.consuming`, we will use this events instead of `setTimeout` with hardcoded one second interval.

```js
  waitForConsumer() {
    return new Promise(resolve => {
      if (!this.consuming) {
        resolve()
      } else {
        this.once('consuming flag becomes off', () => resolve());
      }
    })
  }
```

sempasha:
fix #723

### DIFF
--- a/src/consumer/runner.js
+++ b/src/consumer/runner.js
@@ -1,3 +1,4 @@
+const EventEmitter = require('events')
 const Long = require('long')
 const createRetry = require('../retry')
 const limitConcurrency = require('../utils/concurrency')
@@ -15,8 +16,10 @@ const isRebalancing = e =>
 
 const isKafkaJSError = e => e instanceof KafkaJSError
 const isSameOffset = (offsetA, offsetB) => Long.fromValue(offsetA).equals(Long.fromValue(offsetB))
+const CONSUMING_START = 'consuming-start'
+const CONSUMING_STOP = 'consuming-stop'
 
-module.exports = class Runner {
+module.exports = class Runner extends EventEmitter {
   constructor({
     logger,
     consumerGroup,
@@ -30,6 +33,7 @@ module.exports = class Runner {
     retry,
     autoCommit = true,
   }) {
+    super()
     this.logger = logger.namespace('Runner')
     this.consumerGroup = consumerGroup
     this.instrumentationEmitter = instrumentationEmitter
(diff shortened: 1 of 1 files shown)


src/consumer/runner.js
```diff
@@ -135,20 +150,14 @@ module.exports = class Runner {
...
       if (!this.consuming) {
         return resolve()
       }
-
-      scheduleWait()
+      this.logger.debug('waiting for consumer to finish...', {
+        groupId: this.consumerGroup.groupId,
+        memberId: this.consumerGroup.memberId,
+      })
+      this.once(CONSUMING_STOP, () => resolve())
```
ankon, line 160: This could be simplied to `this.once(CONSUMING_STOP, resolve)`

sempasha: Yes could be. The reason I use wrapper function is safety. Promise must be resolved with no value, while passing `resolve` as event handler directly may have side effects, promise will be resolved with data passed into event. Ofcourse right now this event has no attached data so expression could be simplified (right now). But what if someone will modify event in future?

ankon: Fair enough, thanks!

sempasha:
Could you, please, clarify ability of this PR to be merged and released next days?

Nevon:
Thanks for fixing this. I somehow missed both the original PR and your comment yesterday.

Although I instinctively feel strange about making the runner an EventEmitter only so that it can listen to its own event, the solution is simple and contained enough that if we later decide to refactor it it's very easy to do so. As soon as the pipeline has run, I will merge, after which it will be available in a pre-release build.

PR #730, opened 2020-05-25
docs: remove "Experimental" from Instrumentation Events

WaleedAshraf:


Nevon:
@tulios: Do you agree with my assessment that we can consider these part of the stable API at this point? I don't see us breaking backwards compatibility here at any point soon.

Issue #733, opened 2020-06-01
Unhandled Promise Rejection when consumer config doesn't specify `retry`


ankon:
**Describe the bug**

**To Reproduce**

1. Run the tests (see https://github.com/tulios/kafkajs/issues/654#issuecomment-601112954)
2. Check for UnhandledPromiseRejectionWarnings in output

With current master (+https://github.com/tulios/kafkajs/pull/732):

```
 PASS  src/broker/__tests__/produce.spec.js
 PASS  src/protocol/encoder.spec.js
(node:50706) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'restartOnFailure' of undefined
(node:50706) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 68)
(node:50706) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
 PASS  src/consumer/__tests__/instrumentationEvents.spec.js (30.416s)
```

**Expected behavior**

No such warnings :)

**Environment:**
 - OS: [e.g. Mac OS 10.15.3]: Fedora 31 (x86_64), podman instead of docker
 - KafkaJS version [e.g. 1.12.0]: https://github.com/tulios/kafkajs/pull/732
 - Kafka version [e.g. 2.3.1]: 2.2
 - NodeJS version [e.g. 10.13.0]: v12.16.2

ankon:
Looking around: The only way this could happen is in this line: https://github.com/tulios/kafkajs/blob/4b7e79b2361f20472303f28f6946930ce2cc9908/src/consumer/index.js#L258

Given that the types define `retry` to be optional, I think this needs a proper null/undefined check.

ankon:
~~~~
(node:53291) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'initialRetryTime' of undefined
[...]

PR #734, opened 2020-06-01
Check that `retry` is provided before dereferencing it

ankon:
Fixes https://github.com/tulios/kafkajs/issues/733


Nevon:
Right, this is only an issue in tests because there we might import the consumer directly and instantiate it without `retry`, whereas in normal use you would get the default retry.

Once you've merged latest master into this branch, we can merge.

ankon:
Done!

PR #742, opened 2020-06-03
Add `admin.describeGroups`

### LINKED ISSUE #731: Admin API- Describe consumer groups
I'm trying to find the [Describe consumer groups](https://kafka.apache.org/24/javadoc/org/apache/kafka/clients/admin/Admin.html#describeConsumerGroups-java.util.Collection-) call in the admin API but I can't see it.
My use-case is [I want to see the consumers for a specific topic](https://stackoverflow.com/questions/55937806/apache-kafka-get-list-of-consumers-on-a-specific-topic).

I'd like a `describeGroups` function in the admin api.

**Additional context**
I can see that it [was added to `protocol/requests`](https://github.com/tulios/kafkajs/commit/e589500441847b18f5f56f854f680fa34ec1fae0) but not to the admin api.

Nevon:
Implements a new method on the Admin client, `describeGroups`. It works the same as the equivalent method on the Consumer, except it allows you to query by multiple groupIds.

Fixes #731

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -387,14 +387,6 @@ List groups available on the broker.
 await admin.listGroups()
 ```
 
-Example:
-
-```javascript
-const { ResourceTypes } = require('kafkajs')
-
-await admin.listGroups()
-```
-
 Example response:
 
 ```javascript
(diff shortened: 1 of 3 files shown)


Nevon:
And before I forget, updating the TS type definitions.

Nevon:
> The only thing I think we might be missing is providing the leader / partition assignor in the result. If I remember correctly the memberId of the consumer who was elected is what's used elsewhere as well, would that be appropriate?

We don't provide this in `consumer.describeGroup`, and I don't think we actually have a way to get it, since it's not part of the DescribeGroups response. The only place I can find with that information is in the JoinGroup response, which isn't relevant to the admin client.

tzachshabtay:
Thanks for the quick turnaround! When can I expect to see a new package on npm? @Nevon @JaapRood

Nevon:
There's a new beta release out already. I can't make any promises regarding when we'll make the next stable release.

PR #747, opened 2020-06-08
Add a `jsconfig.json` file

ankon:
This helps Visual Studio Code to work out how to resolve imports/requires, which in turn
helps developers by indicating types better.

----
While there is right now no consensus about having "compile" steps, this file can already help a developer working on the KafkaJS sources by telling Visual Studio code to analyse content. The more we would add JSDoc information then, the more the IDE would be able to hint at problems. For example applying this diff makes it possible to know what the various functions on the cluster are inside the sendMessages function:

```diff
diff --git a/src/producer/sendMessages.js b/src/producer/sendMessages.js
index c194cbe..de61493 100644
--- a/src/producer/sendMessages.js
+++ b/src/producer/sendMessages.js
@@ -12,6 +12,13 @@ const staleMetadata = e =>
     e.type
   )
 
+/**
+ * @param {Object} options
+ * @param {any} options.logger
+ * @param {import("../cluster")} options.cluster
+ * @param {any} options.partitioner
+ * @param {any} options.eosManager
+ */
 module.exports = ({ logger, cluster, partitioner, eosManager }) => {
   const retrier = createRetry({ retries: TOTAL_INDIVIDUAL_ATTEMPTS })
 
``` 

WDYT, would it be worth-while to continue in this direction and start applying JSDoc information?

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,0 +1,14 @@
+{
+  "compilerOptions": {
+    "baseUrl": ".",
+    "module": "commonjs",
+    "target": "es6",
+    "paths": {
+      "testHelpers": ["./testHelpers"]
+    }
+  },
+  "include": [
+    "src",
+    "testHelpers"
+  ]
+}


Nevon:
I don't think we will want to start including IDE specific config files in the repo itself, and I can't find any reference to or specification for jsconfig other than in the vscode documentation. We'd end up with various project directories for every conceivable editor eventually.

How about adding the file to `.gitignore` by default, and then adding the config along with instructions to a "Visual Studio Code Integration" article under "Developing KafkaJS" in the docs though?

PR #748, opened 2020-06-08
Use addMultipleTargetTopics instead of looping over multiple calls to addTargetTopic

ankon:
`addTargetTopic` internally calls `addMultipleTargetTopics` by building a singular array anyways, and
`addMultipleTargetTopics` itself is fairly expensive and requires locking: Simply calling
it directly should make things faster when sending out batches for multiple topics.

---
This should help when batch-producing messages to multiple topics at the same time. Looking at the produce path it seems there could be quite some additional potential for "improvements", but it is difficult to quantify how much all of that would help.

### DIFF
--- a/src/producer/sendMessages.js
+++ b/src/producer/sendMessages.js
@@ -18,9 +18,8 @@ module.exports = ({ logger, cluster, partitioner, eosManager }) => {
   return async ({ acks, timeout, compression, topicMessages }) => {
     const responsePerBroker = new Map()
 
-    for (const { topic } of topicMessages) {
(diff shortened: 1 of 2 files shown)


Nevon:
Nice catch! Could you merge master and we'll get this merged.

Btw, when you create a fork, you can allow maintainers of the forked repo to push to the PR branch, which would allow us to press the nice "update with latest master" button in the PR UI instead of having to bug you.

ankon:
Done. 

I'll have a look for that setting, thanks!

PR #749, opened 2020-06-08
Fix the return value of Cluster.findTopicPartitionMetadata

ankon:
The return value is not an Object, but an Array (of PartitionMetadata objects). This also means we can
simply check the number of entries and don't need to calculate the keys when sending messages.

Note that this exposed a bunch of problems in the tests as well: the mocked calls not just returned the
wrong type, but in fact returned a completely unrelated value.

### DIFF
--- a/src/producer/sendMessages.js
+++ b/src/producer/sendMessages.js
@@ -29,7 +29,7 @@ module.exports = ({ logger, cluster, partitioner, eosManager }) => {
       for (const { topic, messages } of topicMessages) {
         const partitionMetadata = cluster.findTopicPartitionMetadata(topic)
 
-        if (keys(partitionMetadata).length === 0) {
+        if (partitionMetadata.length === 0) {
           logger.debug('Producing to topic without metadata', {
             topic,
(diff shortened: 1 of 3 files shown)


src/producer/sendMessages.js
```diff
@@ -30,7 +30,7 @@ module.exports = ({ logger, cluster, partitioner, eosManager }) => {
       for (const { topic, messages } of topicMessages) {
         const partitionMetadata = cluster.findTopicPartitionMetadata(topic)
 
-        if (keys(partitionMetadata).length === 0) {
```
Nevon, line 33: The fact that this works is kind of a funny Javascript-ism. 🙈

Nevon:
Thanks. If you fix the conflict I'll merge this.

ankon:
Turns out I didn't split another PR properly, and the actual change in the type was already merged with 78fefe2, so this PR is only updating the users to match as well.

(Rebased onto current master)

Nevon:
And now this needs a merge as well.

ankon:
There you go :)

PR #754, opened 2020-06-10
removed maxInFlight option from Default retry and moved into Producer…

sujayneglur123:
… and Consumer


Nevon:
Thanks for the fix!

PR #758, opened 2020-06-10
Fix a typo in a type name

ankon:
Note that this technically is a backwards-incompatible change (the type `TopicPartitionOffsetAndMedata` is no longer exported), which feels ok to me. If it isn't this could be "fixed" by exporting the old type name as well as

```typescript
export type TopicPartitionOffsetAndMedata = TopicPartitionOffsetAndMetadata
```

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -522,7 +522,7 @@ export type GroupDescription = {
 }
 
 export type TopicPartitions = { topic: string; partitions: number[] }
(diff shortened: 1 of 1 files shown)


Nevon:
I definitely think your proposal of exporting the old type as well is the way to go here. Don't want to break people's build on a non-major version release. You can add a `TODO` comment to remove on the next major release.

ankon:
Fair enough :)

PR #761, opened 2020-06-12
Add documentation on vsode integration

Nevon:
Closes #747

WDYT @ankon? I think it strikes an OK middle ground between sharing editor config without tying the codebase itself to an editor-specific format.

This is what it ends up looking like:

Screenshot 2020-06-12 at 10 22 38


ankon:
I still would prefer the file to be in the sources, mainly to avoid things getting out-of-sync. Then again, I don't expect too much changes really, and the documentation approach definitely will help already!

I think the main part of the #747 issue is the second part: I'd like to start documenting functions and types with JSDoc more aggressively and file them as PRs, so that over time the code base becomes less affected by typing errors, and refactoring becomes easier. While we have by now moved to TypeScript completely, we started out using this approach to get to "kinda-but-not-quite typed JavaScript", and we saw improvements very quickly in our perceived code quality and developer productivity.

So, in short: Yes, by all means, docs are good!

Nevon:
> I'd like to start documenting functions and types with JSDoc more aggressively and file them as PRs

Please do. I'm not a huge fan of these type annotations that aren't actually validated against the code itself, since the code might diverge from the type annotations without you noticing, but I do think they are better than nothing, and we're already half-way in with JSDoc, so why not make use of it more. I think the initial ambition would be to cover the public APIs at least.

PR #765, opened 2020-06-15
Remove unneeded hardcoded `node_modules/.bin` references

ankon:
The PATH environment variable will contain the $(npm bin) directory when running package.json scripts, and that
might be different from `node_modules/.bin` depending on the environment.

For `test:debug` we need a path (we're running `node` here, which then calls the real jest), so use `$(npm bin)`
for safety.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   "scripts": {
     "jest": "export KAFKA_VERSION=${KAFKA_VERSION:='2.3'} && NODE_ENV=test echo \"KAFKA_VERSION: ${KAFKA_VERSION}\" && KAFKAJS_DEBUG_PROTOCOL_BUFFERS=1 jest",
     "test:local": "yarn jest --detectOpenHandles",
-    "test:debug": "NODE_ENV=test KAFKAJS_DEBUG_PROTOCOL_BUFFERS=1 node --inspect-brk $(npm bin)/jest --detectOpenHandles --runInBand --watch",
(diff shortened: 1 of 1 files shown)


Nevon:
Given that we are preferring `yarn` for installing dependencies, we are technically not requiring `npm`, so it would feel weird to suddenly require it for running the scripts.

The difference might be kind of academic, as probably 99.999% of people who have `yarn` installed will have `npm` as well, but maybe we can have a simple bash script wrapper that runs `yarn "$@"` if yarn exists and `npm $@` if it doesn't?

ankon:
Actually, good point -- but given that you require yarn (and that is documented!), I guess the easier thing would be to use `$(yarn bin)` instead. The result is the same, with an additional message from yarn about being offline in my case:

~~~~
$ npm run test:debug

> kafkajs@1.12.0 test:debug /home/andreas/project/kafkajs
> NODE_ENV=test KAFKAJS_DEBUG_PROTOCOL_BUFFERS=1 node --inspect-brk $(yarn bin)/jest --detectOpenHandles --runInBand --watch

warning You don't appear to have an internet connection. Try the --offline flag to use the cache for registry queries.                                        
Debugger listening on ws://127.0.0.1:9229/4a32ac6f-c017-4cd5-987a-c6affdafb739
For help, see: https://nodejs.org/en/docs/inspector
[...]
~~~~

ankon:
Rebased onto current master.

Issue #713, opened 2020-04-28
Support for KIP-392: Allow consumers to fetch from closest replica


jgao54:
We are using kafkajs in production and recently upgraded our cluster to support Kafka 2.4, which include [KIP-392](https://cwiki.apache.org/confluence/display/KAFKA/KIP-392%3A+Allow+consumers+to+fetch+from+closest+replica) that supports consumers to fetch from replica instead of the leader. It looks like this feature is not supported by kafkajs yet. The latest `fetch` protocol implemented is 7, but the feature is supported in 11. I was wondering if there are any active initiatives around updating protocol support right now (or support for 2.4 in general)? And if not, would it be okay to directly create a PR introducing v11, or are there any dependencies that requires prior versions to be supported first? Thanks!

tulios:
Hi @jgao54 my plan for the beginning of the year was to upgrade the protocols and potentially reach version 2; the idea was to finalize https://github.com/tulios/kafkajs/projects/2, but a lot of things happened, and I can promise on that timeline anymore.

> would it be okay to directly create a PR introducing v11
The short answer is no. We would like to have the other versions.
The protocol was structured in a way that you can easily maintain a fork just adding the protocol version you want.

Of course, we would love to receive contributions implementing the other versions. We can support you on the way, and as we normalize our time we will get back to the protocol update.

ankon:
Mostly for the record: We're also interested in this feature, and also willing to contribute time and resources to implementation of it.

ankon:
Small update here: We still find this quite relevant to have, so I'm now going to try to implement the missing pieces (protocol support, for a start). To make things easier to merge I'll try to create smaller PRs for each part.

ankon:
Took a while, but: https://github.com/tulios/kafkajs/pull/810 provides an implementation that works in my tests so far.

ankon:
#810 has landed, so I guess we can close this one?

PR #776, opened 2020-06-23
Support Fetch v8 protocol (including client-side throttling)

### LINKED ISSUE #161: Support protocol fetch v5-v8
v5, v6 and v7 implemented on PR #348 

The groundwork for fetch v9 which allows ZSTD compression

ankon:
This is a pre-condition for #713 (we need Fetch v11 for KIP-392).

v8 is exactly the same as v7, the difference is purely in how quota throttling is to be understood. See https://cwiki.apache.org/confluence/display/KAFKA/KIP-219+-+Improve+quota+communication

This provides the last puzzle piece to resolve #161.


Nevon:
If we start supporting v8, then we should also start honoring the throttle time by not sending any more fetch requests during that time, if I'm reading the KIP right. To make things more complicated, we should only introduce this backoff if the throttle time in the response came from a Fetch response v8, since for v7 the broker will have already throttled the client by delaying the response.

ankon:
Fair enough. My ultimate goal is to get to KIP-392 as quickly as possible, so I would be looking into a minimum implementation for throttling here. :)

I see that KIP-219 changes a lot of requests/responses so that each of these can indicate client-side throttling, but from what I understand this still means we can only support it for _some_ requests. IOW: It looks like supporting it for Fetch v8 is fine, as long as there is a reasonable path forward to add similar behavior to other APIs.

What the java client does looks quite reasonable:
1. Identify per response whether this is a "client-side throttling response", and then if so provide a default "throttling time" value (https://github.com/apache/kafka/commit/1facab387f8c2e513c8b7397430251dc44970e35#diff-bcbf176f64322c655e3016ca72cc3f58R165-R177)
2. When processing any response check for whether that is applicable, and if there is needed throttling mark the connection state for that broker accordingly
3. Delay sending stuff on that connection

The Java client does use a version callback, looking at JavaScript and the KafkaJS approach to requests it seems easier to have a flag for "should client-side throttle" and a value for the "how long aspect". One could even put these into the same value (`clientSideThrottleTimeMs`), and define that a value > 0 changes the connection state. 

For the delaying then the `Connection` could be having such a "throtte until" flag, and if that is set it would basically not send out messages until that time is hit. I'm still reading through these areas.

WDYT, would such an approach be acceptable for merging?

Nevon:
I think that sounds very reasonable. And as you say, we can opt into this API-per-API as we upgrade versions.

ankon:
For the test failures: These now come from the various places that check responses against a pre-defined "expected" response. Depending on the Kafka version though we will now get a different response.

I'm going to fix these in the next commit:
1. Always set `throttleTime` to 0 for Fetch V8 responses: This is the server-side throttling applied to the response, and with Fetch V8 the broker won't apply any. Older code can check for that though, and because we never implemented tests for throttling the value will be "just fine".
2. Modify the tests to accept the `clientSideThrottleTime` as an optional field: When running with older versions of Kafka this won't be there, and with never versions it will be -- but then set to the 0 value as there aren't any quotas defined.

src/network/connection.spec.js
```diff
@@ -157,6 +157,22 @@ describe('Network > Connection', () => {
       await expect(connection.send(protocol)).rejects.toThrowError(KafkaJSRequestTimeoutError)
     })
 
+    test.only('respect the client-side throttling', async () => {
```
Nevon, line 160:
```suggestion
    test('respects the client-side throttling', async () => {
```

src/network/connection.js
```diff
@@ -292,28 +302,50 @@ module.exports = class Connection {
           this.failIfNotConnected()
           const entry = { apiKey, apiName, apiVersion, correlationId, resolve, reject }
 
+          const sendRequest = () => {
```
Nevon, line 305: I'm not sure that this is the place to handle this. I think it should be inside the requestQueue instead. Unless I'm reading this wrong, the request queue will now think that a request has been sent even if it's actually being throttled. Conceptually, it feels like it makes more sense for the requestQueue to be the one to do the actual throttling by simply pushing the request onto the queue instead of executing it directly, and then not executing any more requests until the throttle time has passed - same as if `maxInFlightRequests` has been exceeded. From reading the Java implementation, that looks more or less how they've implemented it (`NetworkClient.maybeThrottle`).

Since the requestQueue doesn't decode the response, it still makes sense for the connection to be the one to actually decide for how long to throttle and based on what, but the actual enforcement of the throttling should be moved into the requestQueue.

ankon: I can try that. 

When looking at this yesterday the request queue approach looked like considerable more intrusive changes, while putting into the actual connection was basically trivial and nicely symmetric.

Nevon: It is, but putting it there has a serious drawback, which is that if you are throttled, any requests you make will look like they are "completed" even though they have actually just been scheduled. If you make subsequent requests for which you don't expect a response, the "send" call will resolve immediately even though the request has not been made. Potentially you could even pile up requests in this limbo state. Using the request queue makes it very clear that the request has not been made yet, and the "send" call will not resolve until the throttling is over and the request has actually been made.

ankon:
#786 fixes two issues with tests that lead to errors now. I rather not mix up the commits though, so for this branch it's somewhat expected that the tests mentioned in that PR fail here.

ankon:
I rebased and shuffled the changes a bit around: 
1. First some cosmetics
2. Then implement infrastructure for client-side throttling
3. Then use that infrastructure in the Fetch v8 request/response definitions

src/network/connection.js
ankon: Ok, the changes were indeed a bit more intrusive, but ultimately I agree that putting this in the request queue feels more "right" now.

src/network/requestQueue/index.js
```diff
@@ -67,6 +84,18 @@ module.exports = class RequestQueue {
...
 
+  maybeThrottle(clientSideThrottleTime) {
+    if (clientSideThrottleTime) {
+      const minimumThrottledUntil = Date.now() + clientSideThrottleTime
+      if (this.throttledUntil < minimumThrottledUntil) {
+        this.throttledUntil = minimumThrottledUntil
+
+        // No need to schedule anything here: We will do that when there's
+        // actual requests that are affected by this throttling.
+      }
```
Nevon, line 95: Nitpicky, but I think this would read more clearly as:

```suggestion
      const minimumThrottledUntil = Date.now() + clientSideThrottleTime
      this.throttledUntil = Math.max(minimumThrottledUntil, this.throttledUntil)
```

Issue #767, opened 2020-06-16
Process won't shutdown even after consumer.disconnect()


dhulke:
**Describe the bug**
I'm trying to get the last message of a topic, from partition 0 and after retrieving the message I cant disconnect the consumer, the process hangs forever. I tried calling `await consumer.disconnect();` in the `eachMessage` callback but the disconnect call hangs. Then I tried calling it with `setImmediate` and "end" is printed, but the process still hangs.
This is the error I get:

This is the code:
```
require("dotenv").config();

const { Kafka } = require('kafkajs');

(async () => {

    const kafka = new Kafka({
        // clientId: 'my-app',
        brokers: process.env.KAFKA_BROKER_LIST.split(","),
        ssl: true,
        sasl: {
            mechanism: process.env.KAFKA_SASL_MECHANISM,
            username: process.env.KAFKA_SASL_USERNAME,
            password: process.env.KAFKA_SASL_PASSWORD
        },
    })

    const admin = kafka.admin();

    await admin.connect();
    const offsets = await admin.fetchTopicOffsets(process.env.KAFKA_TOPIC);
    await admin.disconnect();

    const consumer = kafka.consumer({ groupId: 'my-group' })

    await consumer.connect()
    await consumer.subscribe({ topic: process.env.KAFKA_TOPIC })

    consumer.run({ autoCommit: false, eachMessage: async ({ topic, message }) => {
            const event = JSON.parse(message.value.toString());
            console.log(topic, event.event.replayId);
            setImmediate(async () => {
                await consumer.disconnect();
                console.log('end');
            });
        }
    })
    consumer.seek({ topic: process.env.KAFKA_TOPIC, partition: 0, offset: offsets[0]['offset'] })

})(...process.argv.slice(2)).catch(console.log);
```

goriunov:
@dhulke @Nevon i can confirm that this issue persists even on latest beta, it happens only if connection to one or more brokers failed and and in brokerPool file `this.seedBroker` has been rotated in `retrier`. I believe we are missing some disconnect/clean up for old/rotated failed connections somewhere but that is as far as i have tracked for now (will look at it more later). TBH i was able to reproduce that few times with just admin client

dhulke:
[...]

PR #780, opened 2020-06-23
Fix application lock in case of errors before actual connection established 

goriunov:
This is basically fix for #767,  after running some debugging looks like we do not close socket connections properly. One of the examples when timeout hit before connection is established in this case `this.connected` will be `false` and we won't `end` and `unref` socket correctly you can see those handlers with

```js
 console.log(process._getActiveHandles());
```

There is no particular point on having 
```js
 if (!this.connected) {	
      return true	
  }
```
as calling the same `unref` and `end` function multiple times won't affect socket


Nevon:
Good find!

PR #781, opened 2020-06-24
Consistently quote values in the docker-compose configurations

ankon:
This commit aligns the quoting style to always use single quotes rather than an mix
of double quotes, single quotes, and no quotes.

The "no quotes" also introduced a rather ugly other problem: at least podman-compose locally
gets confused by environment variables like `KAFKA_BROKER_ID: 0`, and basically drops these
out, probably because '0' is a falsy value in YAML. This leads to Kafka starting up without
a broker id, and thereby it generates one for itself -- which the tests then are very unhappy about:

~~~~
 FAIL  src/cluster/__tests__/brokerPool.spec.js
  ● Cluster > BrokerPool › #refreshMetadata › updates the list of brokers

    - Expected  - 1
    + Received  + 1

      Array [
    -   "0",
        "1",
    +   "1001",
        "2",
      ]

      178 |       expect(brokerPool.brokers).toEqual({})
      179 |       await brokerPool.refreshMetadata([topicName])
    > 180 |       expect(Object.keys(brokerPool.brokers).sort()).toEqual(['0', '1', '2'])
          |                                                      ^
      181 |       expect(Object.values(brokerPool.brokers)).toEqual(
      182 |         expect.arrayContaining([expect.any(Broker), expect.any(Broker), expect.any(Broker)])
      183 |       )

      at Object.<anonymous> (src/cluster/__tests__/brokerPool.spec.js:180:54)
~~~~

As environment variables should always be strings anyways: Fix that by just always quoting all environment
variables, to reduce the mental load of having all of YAMLs arcane rules in ones head.

### DIFF
--- a/docker-compose.0_10.yml
+++ b/docker-compose.0_10.yml
@@ -5,11 +5,11 @@ services:
     hostname: zookeeper
     container_name: zookeeper
     ports:
-      - "2181:2181"
+      - '2181:2181'
     environment:
-      ZOOKEEPER_CLIENT_PORT: 2181
-      ZOOKEEPER_TICK_TIME: 2000
-      KAFKA_OPTS: "-Djava.security.auth.login.config=/etc/kafka/server-jaas.conf -Dzookeeper.authProvider.1=org.apache.zookeeper.server.auth.SASLAuthenticationProvider"
+      ZOOKEEPER_CLIENT_PORT: '2181'
+      ZOOKEEPER_TICK_TIME: '2000'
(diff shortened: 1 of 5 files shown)


Nevon:
Fun, only slightly related fact, I once worked on a product that had different features enabled in different countries. To manage this configuration, we were using YAML and simple lists of country codes. Something like this:

```yaml
newFeatureEnabled:
  - se
  - fi
  - gb
```

At some point we added support for Norway:

```yaml
newFeatureEnabled:
  - se
  - fi
  - gb
  - no
```

But to our great surprise, it didn't work in Norway. In fact, I seem to recall us having some issue after introducing Norway into this list. Turns out, our YAML parser helpfully converted the string `no` to the boolean value `false`, because that makes a lot of sense.

I don't love YAML.

Issue #782, opened 2020-06-24
Offsets get committed with autoCommit: false if a rejection happens


rafalneves:
`consumer.run({ autoCommit: false, eachMessage: () => Promise.reject(new Error('Error')) });`

On the consumer above, message offsets should never be committed. The first message of the topic should be retried.

That doesn't happen. Offset gets committed and opens the possibility of data loss.

**To Reproduce**
1. Create a consumer with autoCommit disabled
2. On eachMessage trigger a rejection
3. Check topic offsets for your consumer group - if it's a new topic should be 0, but that's not the case.

**Expected behavior**
The topic offset should not be changed on message rejection. **Much less committed with autoCommit: false.**

**Environment:**
 - OS: Not relevant
 - KafkaJS version 1.12.0
 - Kafka version: Not relevant
 - NodeJS version: Not relevant

Nevon:
This was fixed by #866 and is out now in version 1.14.0.

PR #783, opened 2020-06-24
#782 Commit offsets on eachMessage failure only if needed.

rafalneves:


ankon:
I guess the test there should be cloned to have one for both "auto-commit enabled" and "auto-commit not enabled"?

rafalneves:
> I guess the test there should be cloned to have one for both "auto-commit enabled" and "auto-commit not enabled"?

Hmm... should the offset be committed on error? I wouldn't expect so...

ankon:
> Hmm... should the offset be committed on error? I wouldn't expect so...

Intuitively: No, I don't think so either -- the app should control that behavior, and if it considers errors not really problematic it should probably just catch them.

That being said, it seems this behavior was explicitly added and there's a test for it, so something smells like changing it might be a problem: https://github.com/tulios/kafkajs/commit/3aa2d1a44af19cc9febf22d59f34c68170d48670

@tulios do you remember by chance why you added this behavior?

goriunov:
@rafalneves @ankon reading the code i can see that `commitOffsets` only commits previously consumed messages and not the message which failed . Check comment near `resolveOffset` and `commitOffsets`. I believe if message failed we want to always commit what ever has been already processed and then retry things.

```
async processEachMessage(batch) {
    const { topic, partition } = batch

    for (const message of batch.messages) {
      if (!this.running || this.consumerGroup.hasSeekOffset({ topic, partition })) {
        break
      }

      try {
        await this.eachMessage({ topic, partition, message })
      } catch (e) {
        if (!isKafkaJSError(e)) {
          this.logger.error(`Error when calling eachMessage`, {
            topic,
            partition,
            offset: message.offset,
            stack: e.stack,
          })
        }

      
        // here is our error and we try to commit previously resolved messages
        await this.consumerGroup.commitOffsets()
        throw e
      }

       // In this place we resolve offsets, only resolved on should get committed
      this.consumerGroup.resolveOffset({ topic, partition, offset: message.offset })
      await this.consumerGroup.commitOffsetsIfNecessary()
    }
```

rafalneves:
That's not my experience @goriunov. I didn't mix resolved offsets in the experiences that I did, only rejected offsets, but with rejected offsets, the offset keeps increasing. Meaning that it's committed.

Nevon:
@goriunov is correct, at least about the intention of the code. On error, we should commit _resolved_ offsets, to avoid re-processing already processed messages. However, as you correctly identified, there was a bug where we didn't respect the `autoCommit` flag in this case, so good catch on that. It would be good to introduce a test that fails without your change, so that we can avoid introducing this issue again in the future.

Now the second thing is another ballgame. As I mentioned above, we are _not_ supposed to be committing unresolved offsets. So if you are not resolving anything, and you are still advancing the offset, then that's a separate bug that we should look into. However, that's separate from this change and can be dealt with separately.

rafalneves:
@Nevon you are right. I can't replicate the second issue, must've made some confusion has I was trying to get arround the issue on my use case.

So... it's just a matter of update current failing test, right?

goriunov:
@rafalneves regarding second issue i believe i may have explanation. So when you create new consumer group it will consume from latest offset. So if you never allowed for connection to properly commit offsets then kafka will never know what is your current offset and every time you start your consumer it will always look for latest message, which is correct behaviour. To confirm that you can try running describe on kafka side and see if you have current offset registered.

Nevon:
> So... it's just a matter of update current failing test, right?

Correct, or rather, adding a test to prove that the issue exists and that your change fixes it.

kafkajs.iml
```diff
@@ -0,0 +1,11 @@
...
+<module type="JAVA_MODULE" version="4">
+  <component name="NewModuleRootManager" inherit-compiler-output="true">
+    <exclude-output />
+    <content url="file://$MODULE_DIR$">
+      <sourceFolder url="file://$MODULE_DIR$/src/producer/partitioners/defaultJava" isTestSource="false" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
```
ankon, line 11: This seems to be an IntelliJ Idea project file? In any case, it shouldn't be here. :)

src/consumer/runner.js
```diff
@@ -181,8 +181,8 @@ module.exports = class Runner extends EventEmitter {
           })
         }
 
-        // In case of errors, commit the previously consumed offsets
-        await this.consumerGroup.commitOffsets()
+        // In case of errors, commit the previously consumed offsets if needed
+        await this.consumerGroup.commitOffsetsIfNecessary()
```
goriunov, line 185: I am not sure if this one should be `commitOffsetsIfNecessary` as in case if we get an error we want to commit offsest always (except if we have `this.autoCommit` false) but `commitOffsetsIfNecessary` also checks last commit time which means that we may not commit all processed messages in case if previous commit was too near...

kafkajs.iml
rafalneves: Ups, miss that.

src/consumer/runner.js
rafalneves: I see. You are right. Maybe we need a third commitOffsetIfEnabled function?

goriunov: There are 2 functions available on runner which basically already checking all the things properly (not sure why we don't just use them...) I am wondering if we can rename those to more appropriate names and use those or just move `autoCommit` check logic in to `consumerGroup.commitOffsets` and `consumerGroup.commitOffsetsIfNecessary` functions and delete below one completely, this should make code cleaner and simpler  @Nevon what do you think ?

```js
 autoCommitOffsets() {
    if (this.autoCommit) {
      return this.consumerGroup.commitOffsets()
    }
  }

  autoCommitOffsetsIfNecessary() {
    if (this.autoCommit) {
      return this.consumerGroup.commitOffsetsIfNecessary()
    }
  }
```

Nevon: Gif of Tina Fey saying "Well, here we are 32 years later"

`consumerGroup.commitOffsets` is exposed to the `eachBatch` interface as well as via `consumer.commitOffsets` as a way to manually commit offsets, so that really shouldn't really need to know about `autoCommit`. I think the offending line (185) should just be changed to calling  `autoCommitOffsetsIfNecessary` and then the bug will be fixed.

tulios:
@rafalneves thanks for the PR, we took too long to get this one in so I ended up fixing it in a smaller PR (#866).
The issue is fixed in the latest pre-release version.

PR #784, opened 2020-06-24
Only disconnect() the consumer & producers when they were created

ankon:
This test sometimes fails without creating the consumers, and then we get an additional misleading
error in `afterEach` as well. Avoid these.

Related-to: #654


ankon:
I looked through other consumer tests, and found that some already had a similar fix -- so I went ahead and aligned them all onto that pattern.

PR #785, opened 2020-06-24
Resolve socket requests without response immediately when they have been queued

ankon:
Found while working on #776 and reading the code in the request queue: When we push a request that doesn't need a reply and we can execute it immediately we will do so, and then properly complete the request. But: When the request would have to be enqueued we will not check the "does it need a response", and so likely this request will just get stuck somehow.

### DIFF
--- a/src/network/requestQueue/index.js
+++ b/src/network/requestQueue/index.js
@@ -122,16 +122,23 @@ module.exports = class RequestQueue {
       return
     }
 
+    this.sendSocketRequest(socketRequest)
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


goriunov:
Nice catch ! :)

Nevon:
I could have sworn I already commented on this yesterday, but yeah, nice catch. I am however perplexed as to how this ever worked, then. Maybe it's just rare that you make a request where you don't expect a response while you are also hitting the maxInFlightRequests limit, so no one noticed.

PR #786, opened 2020-06-24
Fix the requestQueue.fulfillRequest test for when there are pending requests

ankon:
The test was essentially testing its own setup: The outer `beforeEach` would put the `request`
into the queue, which means the `send` method would get called (no pending/inflight requests
at that point). Then the inner `beforeEach` would saturate the queue, and then the test would
add the same request again (this time queued as pending).

When fulfilling the request with this correlation id then we would complete the first request, and
then pop the request added by the test into "in-flight" mode: At this point this.inflight[correlationId]
actually points to the second request, because `send` will have overwritten the entry.

fulfillRequest then does `this.inflight.delete(correlationId)` (assuming to delete the just-fulfilled
entry), but effectively clears out the inflight-ness of the second request.

This commit fixes the problem in the test:
1. Explicitly push the request into the queue in the tests, not in the `beforeEach` call, and
2. Pick one of the dummies created by the inner `beforeEach` to fulfill, and
3. Fix the inflight size check: We had N, and this test removes one and then checks that another
   takes its place -- so the expected size is the same size as before.

Note that this also points to another (theoretical) problem here: If a request is arriving with
the same correlation id as another request, and both of these end up being "in-flight" at the same
time, we would notice the conflict.

### DIFF
--- a/src/network/requestQueue/index.spec.js
+++ b/src/network/requestQueue/index.spec.js
@@ -294,8 +294,11 @@ describe('Network > RequestQueue', () => {
       }
 
       requestQueue.push(request)
+
+      // Pick one of the inflight requests to fulfill
+      const correlationId = requestQueue.inflight.keys().next().value
       requestQueue.fulfillRequest({
-        correlationId: request.entry.correlationId,
+        correlationId,
         payload,
         size,
       })
--- a/src/network/requestQueue/index.spec.js
+++ b/src/network/requestQueue/index.spec.js
@@ -133,11 +133,10 @@ describe('Network > RequestQueue', () => {
         entry: createEntry(),
         expectResponse: true,
       }
-
-      requestQueue.push(request)
     })
 
     it('deletes the inflight request and calls completed on the request', () => {
+      requestQueue.push(request)
       expect(requestQueue.inflight.size).toEqual(1)
 
       requestQueue.fulfillRequest({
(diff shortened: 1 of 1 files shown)


ankon:
> Note that this also points to another (theoretical) problem here: If a request is arriving with
the same correlation id as another request, and both of these end up being "in-flight" at the same
time, we would notice the conflict.

One idea would be something like this:
```diff
diff --git a/src/network/requestQueue/index.js b/src/network/requestQueue/index.js
index a41fd1e..12abe2a 100644
--- a/src/network/requestQueue/index.js
+++ b/src/network/requestQueue/index.js
@@ -94,6 +94,9 @@ module.exports = class RequestQueue {
       instrumentationEmitter: this.instrumentationEmitter,
       requestTimeout,
       send: () => {
+        if (this.inflight.has(correlationId)) {
+          throw new Error('Heeeeelp!')
+        }
         this.inflight.set(correlationId, socketRequest)
         pushedRequest.sendRequest()
       },
```

... but I'm not sure really whether this is a relevant issue, or whether maybe just a debug-mode-assertion would be fine, or maybe just log it?

ankon:
A similar issue also exists in the test for the "emits event when pending queue is reduced", https://github.com/tulios/kafkajs/pull/786/commits/9ac2b17ef2c7290feac9e4cd8b57e4a0fec809bf fixes that one.

Nevon:
> ... but I'm not sure really whether this is a relevant issue, or whether maybe just a debug-mode-assertion would be fine, or maybe just log it?

If we have the same correlation id, something has gone horribly wrong, since the correlation id is supposed to be a sequence of integers that wrap around once it reaches the max value. If we have a request in flight with the same correlation id, then there's a bug in how we generate the correlation id and probably the right thing is to throw a non-retriable error saying that there's a duplicate correlation id - basically an invariant.

PR #787, opened 2020-06-25
fix: add isNaN check for concurrency limit

### LINKED ISSUE #788: Consumer hangs if concurrency is `NaN`
**Describe the bug**
Consumer hands up if `partitionsConsumedConcurrently` is `NaN`

**To Reproduce**
Set `partitionsConsumedConcurrently = NaN`

**Environment:**
 - KafkaJS version: 1.12.0

WaleedAshraf:
Closes #788

### DIFF
--- a/src/utils/concurrency.js
+++ b/src/utils/concurrency.js
@@ -6,7 +6,7 @@ const REJECTED_ERROR = new KafkaJSNonRetriableError(
 function NOOP() {}
 
 const concurrency = ({ limit, onChange = NOOP } = {}) => {
-  if (typeof limit !== 'number' || limit < 1) {
+  if (isNaN(limit) || typeof limit !== 'number' || limit < 1) {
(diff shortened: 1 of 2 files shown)


ankon:
Out of interest: How did you manage to get this to be `NaN` in the first place? I thinking there's probably more places that don't specifically look at NaN in the code base because there's no reasonable path to get into that state.

Nevon:
I'm generally not a big fan of these runtime type checks when there isn't a reasonable way that you'd expect to possibly receive weird values, but given that it's already checking that the limit is a number I supposed there's no harm in also checking that it's not NaN (love that NaN is a number and also neither equal to, larger or smaller than 1). But yeah, how did you get it to be NaN to begin with? 😃

WaleedAshraf:
Before we pass config values to kafkajs, we do parseInt. And, There is WTF JS 😆 
```
Number.parseInt(undefined) // NaN
```

PR #778, opened 2020-06-23
Support Fetch v9 protocol (minimal)

ankon:
NOTE: This is a draft on top of the earlier #776, and merely here as a "intent to get this merged". Fetch v9 is needed as precondition for #713.

PR #792, opened 2020-06-25
Support Fetch v10 protocol (minimal) 

ankon:
This provides the protocol definitions for Fetch v10 ([KIP-110: ZStd Support](https://cwiki.apache.org/confluence/display/KAFKA/KIP-110%3A+Add+Codec+for+ZStandard+Compression)), without any additional changes to actually support consuming or producing zstd compressed content.

For producing: This will require updating the produce APIs as well, see https://github.com/tulios/kafkajs/projects/2 for the bigger picture.
For consuming: This might actually work just fine, if one provides a suitable codec through the regular extension mechanism.

This is mainly here as a prerequisite for KIP-392 support (#713), which requires Fetch v11 messages.

~Note: This PR is floating on top of the Fetch v9 changes (#778), and should stay a draft until we can merge that one first.~


ankon:
Test failures are unrelated, see #786

src/broker/__tests__/fetch.spec.js
```diff
@@ -21,6 +21,16 @@ const maxBytesPerPartition = 1048576 // 1MB
 const maxWaitTime = 100
 const timestamp = 1509827900073
 
+expect.extend({
```
Nevon, line 24: Add this to `testHelpers/setup.js` instead, where we already introduce custom matchers.

ankon: Good one, done with aacf5c2.

Nevon:
So this PR is pretty simple an uncontroversial, but I want to make sure that I understand what the implications are. By supporting Fetch V10, the only thing that will happen is that if a partition is using zstd compression, we will get back data (that we can't decompress without a codec), rather than an `UNSUPPORTED_COMPRESSION_TYPE` error?

ankon:
Yes, this is my understanding as well. KafkaJS supports the v2 format from the KIP, so the discussion about down-conversion shouldn't apply at all. But sending a Fetch v10 request we're indicating that we understand a response that talks about zstd (which we do-ish, assuming we had a codec).

Issue #796, opened 2020-06-26
UnhandledPromiseRejection in beta.38


goriunov:
**Describe the bug**

FIY, Have been running some chaos testing and stumbled on unhandled rejection in beta.38, also got the same issue during our deployment (looks like it is there for a while): 

```
(node:55902) UnhandledPromiseRejectionWarning: KafkaJSRequestTimeoutError: Request Fetch(key: 1, version: 7) timed out
    at SocketRequest.timeoutRequest (/node_modules/kafkajs/src/network/requestQueue/socketRequest.js:104:19)
    at inflight.forEach.request (/node_modules/kafkajs/src/network/requestQueue/index.js:59:21)
    at Map.forEach (<anonymous>)
    at Timeout.requestTimeoutIntervalId.setInterval (/node_modules/kafkajs/src/network/requestQueue/index.js:57:23)
    at ontimeout (timers.js:436:11)
    at tryOnTimeout (timers.js:300:5)
    at listOnTimeout (timers.js:263:5)
    at Timer.processTimers (timers.js:223:10)
```

This usually happens after disconnect, basically we call disconnect and for some reason we get timeout after which is throwing `UnhandledPromiseRejection` i am doing more investigations around this area at the moment as this may cause issues during our zero downtime deployment. Also looks like this stack trace is useless as it only shows where the error has been created....

**Expected behavior**
To do not have unhandled rejections

**Environment:**
 - OS: [e.g. Mac OS 10.15.3] Mac OS 10.15.4
 - KafkaJS version [e.g. 1.12.0] beta.38
 - Kafka version [e.g. 2.3.1] 
 - NodeJS version [e.g. 10.13.0] 10.16.3

Nevon:
Fixed by #797

PR #797, opened 2020-06-26
Fix unhandled rejections, improve fetch flow

goriunov:
This is fix for #796 and probably other possible issues. Main reason for unhandled rejections was iterator `BufferedAsyncIterator` so when we disconnect  consumer there is high chance in slow systems that we may still be waiting for fetch 

https://github.com/tulios/kafkajs/blob/fce9c09a00e0d7969f7a79adff73257a557d2af3/src/consumer/runner.js#L269 

which returns array of requests (Promises), in the current implementation we will check 

https://github.com/tulios/kafkajs/blob/fce9c09a00e0d7969f7a79adff73257a557d2af3/src/consumer/runner.js#L320

and basically exit fetch loop but requests on that iterator are not yet rejected or processed and we are not awaiting them any where which causes unhandledRejection.

so the nearest place where we are awaiting for requests is at 

https://github.com/tulios/kafkajs/blob/fce9c09a00e0d7969f7a79adff73257a557d2af3/src/consumer/runner.js#L375

we have basically sync code between await fetch and this `promise.all` so there is no need for `this.running` check:

https://github.com/tulios/kafkajs/blob/fce9c09a00e0d7969f7a79adff73257a557d2af3/src/consumer/runner.js#L320

Also i have moved
```js
 const batches = await result.value
 expectedNumberOfExecutions += batches.length
```

up as we want to always `await result` to make sure we catch all errors properly

the same applies for 

```js
await Promise.all(enqueuedTasks.map(fn => fn()))
```

---------------------------------------------------------
---------------------------------------------------------
---------------------------------------------------------
Another slightly unrelated fix for request queue, we want to destroy loop even if we have no inflight requests:
```js
if (!this.isConnected()) {
          this.destroy()
 }
```


goriunov:
With above implementation we basically allow every single request to be processed as expected after disconnect and cleanly exit the loop before committing and heartbeat 
```js
  ...
   await Promise.all(enqueuedTasks.map(fn => fn()))
    requestsCompleted = true

    if (expectedNumberOfExecutions === numberOfExecutions) {
      unlock()
    }

    const error = await lock
    if (error) {
      throw error
    }

    if (!this.running) {
      return
    }

    await this.autoCommitOffsets()
```


Also we allow batches to be processed as expected but don't pass any data to the user and don't heartbeat, basically will loop through fetched batches and exit cleanly:

```js
enqueuedTasks.push(async () => {
        const batches = await result.value
        expectedNumberOfExecutions += batches.length

        batches.map(batch =>
          concurrently(async () => {
            try {
              if (!this.running) {
                return
              }

              if (batch.isEmpty()) {
                return
              }

              await onBatch(batch)
            } catch (e) {
              unlockWithError(e)
            } finally {
              numberOfExecutions++
              if (requestsCompleted && numberOfExecutions === expectedNumberOfExecutions) {
                unlock()
              }
            }
          }).catch(unlockWithError)
        )
      })
```

goriunov:
Have been running this branch for some time, could not trigger those unhandled rejections any more. Unfortunately summed on 
https://issues.apache.org/jira/browse/KAFKA-9478 in kafka 2.4 (sever side). Should have nothing to do with kafkajs though.

Nevon:
There must be something I don't fully understand about this diff. So the original problem is that we call `consumerGroup.fetch()`, and while awaiting that, someone calls `consumer.disconnect()` which sets `this.running = false`. What happens after this, with the current implementation, is that we resolve `runner.fetch` without awaiting the promises returned from `consumerGroup.fetch()`, which could cause an unhandled promise rejection.

So far I'm with you.

Now the intention of your solution, if I'm reading it right, is to await all the requests before resolving `runner.fetch`, and just short-circuiting the actual processing of those responses. That way, if there is an error with the ongoing requests, that error would bubble up and be handled.

The downside of this is that we now have to wait for all the promises returned from `consumerGroup.fetch()` before we're able to stop. I'm not actually sure that we care about errors from `consumerGroup.fetch()` when we are stopping the runner. For example, if we got a `REBALANCE_IN_PROGRESS` error, the runner would attempt to handle this by re-joining the group, which there's no point to doing if we're in the process of stopping. So instead of delaying the stopping, maybe we can just catch and ignore any errors that happen if we're no longer running?

```js
while (true) {
  if (result.done) {
    break
  }

  if (!this.running) {
    // Note that we don't actually await this to resolve, we just ignore the error
    result.value.catch(error => {
      this.logger.debug('Ignoring error in fetch request while stopping runner', {
        error: error.message || error,
        stack: error.stack
      })
    })
    return
  }
```

I'm not 100% sure about this. It might be cleaner to actually await all the ongoing requests, but it just seems pointless since we don't actually care about the results.

src/consumer/__tests__/runner.spec.js
```diff
@@ -347,5 +347,27 @@ describe('Consumer > Runner', () => {
...
+            }, 10)
+          )
+      )
+
+      runner.scheduleFetch = jest.fn()
+      await runner.start()
+      runner.running = false
+
+      runner.fetch()
+      await sleep(100)
```
goriunov, line 370: Interesting that `jest` automatically wraps unhandled rejections if you don't exit your test too early. With this sleep we are making sure that if for any reason we get unhandled rejection this test will fail

goriunov:
@Nevon good point i have updated PR so if consumer is not running it will ignore errors on every request. Small change in your code as we have to loop though every single `result`. So if consumer is not running `enqueuedTasks` will be empty and we will just fall though the rest of the the function and will we exit at   https://github.com/tulios/kafkajs/pull/797/files#diff-0706eb2c0060b2da86157638eb477a86R388

Issue #770, opened 2020-06-19
[question] admin.createTopics Timeout


axxop:
The timeout setting has no effect when creating a topic:

my code:
```
  await admin.createTopics({
    topics: [{ topic }],
    waitForLeaders: true,
   timeout: 1000000
  })
```

It will fail after waiting for about 10 seconds:
```
[KafkaJSTimeout: Timed out while waiting for topic leaders
    at Timeout._onTimeout (/xxxxx/node_modules/kafkajs/src/utils/waitFor.js:41:23)
    at listOnTimeout (internal/timers.js:549:17)
    at processTimers (internal/timers.js:492:7) {
  name: 'KafkaJSTimeout',
  retriable: false,
  helpUrl: undefined
}
](url)
```
I checked the kafkajs code:
(https://github.com/tulios/kafkajs/blob/master/src/admin/index.js#L108-L121), in L118, timeout is not passed as maxWait parameter.

Is this expected?

Nevon:
That is indeed a bug.

PR #800, opened 2020-06-27
fix admin.createTopics timeout .

axxop:
#770


Nevon:
Thanks for the fix, and welcome to the community!

PR #801, opened 2020-06-29
Add JoinGroup protocol v3 and v4

goriunov:
This will add protocol support for JoinGroup v3 and v4, based on https://cwiki.apache.org/confluence/display/KAFKA/KIP-394%3A+Require+member.id+for+initial+join+group+request v4 protocol we need to retry join with provided member id if we get MEMBER_ID_REQUIRED


src/broker/index.js
```diff
@@ -349,7 +349,8 @@ module.exports = class Broker {
     groupProtocols,
   }) {
     const joinGroup = this.lookupRequest(apiKeys.JoinGroup, requests.JoinGroup)
-    return await this.connection.send(
+
```
Nevon, line 352: This will throw a `KafkaJSProtocolError` if we get a `MEMBER_ID_REQUIRED` error. You will need to catch that error and retry the request with the new memberId, rather than looking manually at `response.errorCode`.

```js
const joinGroup = this.lookupRequest(apiKeys.JoinGroup, requests.JoinGroup)
const makeRequest = (assignedMemberId = memberId) => this.connection.send(
  joinGroup({
    groupId,
    sessionTimeout,
    rebalanceTimeout,
    memberId: assignedMemberId,
    protocolType,
    groupProtocols
  })
)

try {
  return await makeRequest()
} catch(error) {
  if (e.type === 'MEMBER_ID_REQUIRED') {
    return makeRequest(response.memberId)
  }

  throw error
}
```

src/protocol/requests/joinGroup/v3/request.js
```diff
@@ -0,0 +1,33 @@
...
+module.exports = ({
+  groupId,
+  sessionTimeout,
+  rebalanceTimeout,
+  memberId,
+  protocolType,
+  groupProtocols,
+}) =>
+  Object.assign(
+    requestV1({
```
Nevon, line 24: The V2 request is apparently identical to V1, so it doesn't really matter from a practical perspective, but just for symmetry it would be nice to have v3 import from v2.

src/protocol/requests/joinGroup/v3/request.spec.js
```diff
@@ -0,0 +1,21 @@
+const RequestV2Protocol = require('./request')
```
Nevon, line 1: This is all referencing v2, rather than v3.

src/protocol/requests/joinGroup/v4/request.js
```diff
@@ -0,0 +1,33 @@
...
+module.exports = ({
+  groupId,
+  sessionTimeout,
+  rebalanceTimeout,
+  memberId,
+  protocolType,
+  groupProtocols,
+}) =>
+  Object.assign(
+    requestV1({
```
Nevon, line 24: Same with this. This should reference v3.

src/protocol/requests/joinGroup/v4/response.js
```diff
@@ -0,0 +1,50 @@
...
+    members: decoder.readArray(decoder => ({
+      memberId: decoder.readString(),
+      memberMetadata: decoder.readBytes(),
+    })),
+  }
+}
+
+const parse = async data => {
+  // KIP-394: Require member.id for initial join group request
+  if (data.errorCode !== 79 && failure(data.errorCode)) {
```
Nevon, line 40: Oh, I see now why it wasn't throwing in broker.joinGroup. This is pretty nasty in my opinion. We always throw protocol errors rather than returning with error codes, and calling code shouldn't have to interact with protocol error codes and stuff like that. It should just be protocol errors, so that you don't have to know magic numbers like 79.

goriunov: Yeah, i thought about that at first but at the moment if we throw error we will not propagate response with new member_id so we wont be able to use this one at the broker level, we can try to change parse function to add member id to the error response, what do you think about that ?

src/broker/index.js
goriunov: Updated, i have changed v4 parser to pass  new `memberId` with error

src/protocol/requests/joinGroup/v4/response.js
```diff
@@ -0,0 +1,33 @@
...
+ *   members => member_id member_metadata
+ *     member_id => STRING
+ *     member_metadata => BYTES
+ */
+
+const parse = async data => {
+  if (failure(data.errorCode)) {
+    const error = createErrorFromCode(data.errorCode)
+    if (error.type === 'MEMBER_ID_REQUIRED') {
+      error.memberId = data.memberId
```
goriunov, line 21: This looks like the cleanest solution, i was also thinking on changing inside `createErrorFromCode` function but that function does not get full data set and will require updating every single parser or passing second param which looks ugly to only support single case

Nevon: Right, I didn't think about the fact that the protocol error doesn't include the response. Take a look at what we're doing with `KafkaJSOffsetOutOfRange`, which has a similar problem. I think that solution is appropriate here as well. Basically create a new `MemberIdRequired` class that extends `KafkaJSProtocolError` and adds a a `memberId` field, and then throw that instead of a generic protocol error if the type is `MEMBER_ID_REQUIRED`.

goriunov: Eventually got some time, have updated PR based on above !

PR #810, opened 2020-07-10
Support Fetch v11 (KIP-392 follower fetching)

ankon:
This is a working implementation of KIP-392, on top of the earlier fetch v9 and v10 PRs.

To use this: Set the `rackId` parameter when creating a consumer, and make sure that the Kafka cluster is configured accordingly (minimally: set `broker.rack` and `replica.selector.class`)


ankon:
Rebased and squashed the unneeded fixups.

src/consumer/consumerGroup.js
```diff
@@ -566,4 +614,65 @@ module.exports = class ConsumerGroup {
...
+    const partitionMetadata = this.cluster.findTopicPartitionMetadata(topic)
+    const preferredReadReplicas = this.preferredReadReplicasPerTopicPartition[topic]
+    return partitions.reduce((result, id) => {
+      const partitionId = parseInt(id, 10)
+      const metadata = partitionMetadata.find(p => p.partitionId === partitionId)
+      if (!metadata) {
+        return result
+      }
+
+      if (metadata.leader === null || metadata.leader === undefined) {
```
Nevon, line 637: This short-hand is equivalent:

```suggestion
      if (metadata.leader == null) {
```

src/consumer/consumerGroup.js
```diff
@@ -566,4 +614,65 @@ module.exports = class ConsumerGroup {
...
+        throw new KafkaJSError('Invalid partition metadata', { topic, partitionId, metadata })
+      }
+
+      // Pick the preferred replica if there is one, and it isn't known to be offline, otherwise the leader.
+      let nodeId = metadata.leader
+      if (preferredReadReplicas) {
+        const { nodeId: preferredReadReplica, expireAt } = preferredReadReplicas[partitionId] || {}
+        if (Date.now() > expireAt) {
+          this.logger.info('Preferred read replica information has expired, using leader', {
```
Nevon, line 646: It feels like this should be a debug log, not an info log. WDYT?

src/consumer/consumerGroup.js
```diff
@@ -566,4 +614,65 @@ module.exports = class ConsumerGroup {
...
+
+      if (metadata.leader === null || metadata.leader === undefined) {
+        throw new KafkaJSError('Invalid partition metadata', { topic, partitionId, metadata })
+      }
+
+      // Pick the preferred replica if there is one, and it isn't known to be offline, otherwise the leader.
+      let nodeId = metadata.leader
+      if (preferredReadReplicas) {
+        const { nodeId: preferredReadReplica, expireAt } = preferredReadReplicas[partitionId] || {}
+        if (Date.now() > expireAt) {
```
Nevon, line 645: This is a bit nitpicky, but given the name of `expireAt`, I would say that we shouldn't use an entry that is expiring right now.

```suggestion
        if (Date.now() >= expireAt) {
```

src/consumer/consumerGroup.js
```diff
@@ -566,4 +614,65 @@ module.exports = class ConsumerGroup {
...
+            leader: metadata.leader,
+          })
+          // Drop the entry
+          delete preferredReadReplicas[partitionId]
+        } else if (preferredReadReplica != null) {
+          // Valid entry, check whether it is not offline
+          // Note that we don't delete the preference here, and rather hope that eventually that replica comes online again
+          const offlineReplicas = metadata.offlineReplicas
+          if (Array.isArray(offlineReplicas) && offlineReplicas.includes(nodeId)) {
+            this.logger.info('Preferred read replica is offline, using leader', {
```
Nevon, line 661: My gut tells me this should probably also be `debug`, but I'm not sure.

Nevon:
Don't forget to add `rackId` to the docs.

ankon:
Thanks for the review!

> Don't forget to add rackId to the docs.

Good one, will do.

> [...] should be debug

I initially had the messages on debug, and found it a bit too silent -- seeing that the preferred reader is picked made me feel good. But, after running with this for a while in production I agree: Things work, and it is just noise to see these "progress" log messages.
I do feel a bit for the "leader is gone, help!" message to be more important, but then again the ops team will likely know that there brokers is gone and doesn't need the developers start screaming :)

So yeah, will change both to debug.

I'll fix these up together with the nit-picks, and rebase to master -- I see there's a lot of things moving today :100:

Nevon:
> I see there's a lot of things moving today 💯

Heh, yeah, once I accumulate enough guilt and my unread emails starts to span multiple pages, I do a push to go through everything. 😅

Nevon:
This is a really cool feature. I'm excited to see how much this can improve both costs and performance, especially in multi-AZ/multi-cloud setups.

jgao54:
@ankon super excited about this! We are looking to enable this for our cluster. Are there any other dependencies or is it sufficient to cherry-pick the 4 mentioned PRs from the blog post? appreciate the guidance!

Nevon:
> Are there any other commit dependencies or is it sufficient to cherry-pick the 4 mentioned PRs from the blog post?

That should be sufficient, although I would probably recommend to just run `1.13.0-beta.68` unless you want to maintain a divergent fork yourself.

PR #811, opened 2020-07-10
Avoid repeated costly copies of buffers when working with encoders

ankon:
Instead of using `Buffer.concat` whenever writing into the encoder, and thereby copying the byte
contents of earlier added content potentially many times, this commit changes the encoder to use
a bigger pre-allocated buffer to write into directly. If that buffer fills up it a new one will be created
and the old contents copied into that.

---
This is somewhat related to https://github.com/tulios/kafkajs/pull/475. In our production environment I can see improvements with this change in the order of maybe ~10% less GC reclaimed bytes/s, which does help.

### DIFF
--- a/src/protocol/encoder.js
+++ b/src/protocol/encoder.js
@@ -54,44 +54,77 @@ module.exports = class Encoder {
     return Encoder.sizeOfVarInt(size) + size
   }
 
-  constructor() {
-    this.buffer = Buffer.alloc(0)
+  static nextPowerOfTwo(value) {
+    return 1 << (31 - Math.clz32(value) + 1)
+  }
+
+  /**
+   * Construct a new encoder with the given initial size
+   *
+   * @param {number} [initialSize] initial size
+   */
+  constructor(initialSize = 511) {
+    this.buf = Buffer.alloc(Encoder.nextPowerOfTwo(initialSize))
+    this.offset = 0
+  }
+
+  /**
+   * @param {Buffer} buffer
+   */
+  writeBufferInternal(buffer) {
+    const bufferLength = buffer.length
+    this.ensureAvailable(bufferLength)
+    buffer.copy(this.buf, this.offset, 0)
+    this.offset += bufferLength
+  }
+
+  ensureAvailable(length) {
+    if (this.offset + length > this.buf.length) {
+      const newLength = Encoder.nextPowerOfTwo(this.offset + length)
+      const newBuffer = Buffer.alloc(newLength)
+      this.buf.copy(newBuffer, 0, 0, this.offset)
+      this.buf = newBuffer
+    }
+  }
+
+  get buffer() {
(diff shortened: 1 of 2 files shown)


src/protocol/encoder.js
```diff
@@ -54,44 +54,77 @@ module.exports = class Encoder {
...
+    return 1 << (31 - Math.clz32(value) + 1)
+  }
+
+  /**
+   * Construct a new encoder with the given initial size
+   *
+   * @param {number} [initialSize] initial size
+   */
+  constructor(initialSize = 511) {
+    this.buf = Buffer.alloc(Encoder.nextPowerOfTwo(initialSize))
```
Nevon, line 67: This means that if I set an initialSize of 512, we will actually allocate 1024 bytes. The following test would fail:

```js
it('does not over-allocate unless necessary', () => {
  let encoder = new Encoder(512)
  expect(Buffer.byteLength(encoder.buf)).toEqual(512)
})
```

Replacing `nextPowerOfTwo` with the following and updating all references would solve it:

```
static roundToPowerOfTwo(value) {
  value--
  value |= value >> 1
  value |= value >> 2
  value |= value >> 4
  value |= value >> 8
  value |= value >> 16
  value++

  return value
}
```

The question is, does it make sense? By over-allocating on the first pass, we do avoid an initial resize in what I'm guessing is most cases. On the other hand, if I actually _know_ that the encoder is going to need 512 bytes of space, it doesn't make sense that we allocate 1024.

Nevon:
Cool! I've been wanting to do this forever, but never got around to finishing it up. I just have one question about the implementation, but I'm fine with merging either way.

src/protocol/encoder.js
ankon: > This means that if I set an initialSize of 512, we will actually allocate 1024 bytes. The following test would fail:

Indeed. That's why the initial size is 511 :)
This is a bit modelled after NodeJS's logic in the connection backlog area:
https://github.com/nodejs/node/blob/0edeeec3ec317bb8ed3bc9f6289eb9e7bfe0c61b/lib/net.js#L1282-L1285

To be honest: I don't mind either way. There's not that many places that know the exact size ahead of time, and those that do won't know the protocol overhead they need to account for. On the other hand it does feel more natural to say "512", rather than having to know "511, because it will add one", so your `roundUpToPowerOfTwo` wins on readability IMHO.

I played a bit with the tests and some verbose logging of resizing operations happening with 256, 512, and 1024 as initial sizes, and 512 "felt ok". Probably you can science a lot more out of it :)

Nevon: > Probably you can science a lot more out of it :)

Yeah, I was thinking about that and was gonna do exactly what you just mentioned, but in reality I'm sure the sizes will differ wildly depending on your use case. Sounds fun, but it's good enough as is.

PR #813, opened 2020-07-10
Improve type information of internal functions

ankon:
These were "fallout" while working on KIP-392 (see #810), but can be treated as independent to hopefully simplify reviewing/merging.


Nevon:
Interesting. I didn't know you could import Typescript types from within JSDoc comments.

ankon:
While trying to find a reference: There's probably a caveat here that this syntax comes from TypeScript, and right now the JSDoc people haven't implemented it (but it seems they're not actively against it either):

* TypeScript reference for the syntax: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#import-types
* JSDoc plugin providing it when building documentation: https://github.com/SoftwareBrothers/better-docs#typedefimport
* JSDoc issue tracking the implementation and discussing the approaches: https://github.com/jsdoc/jsdoc/issues/1645

Nevon:
Since we're not generating docs from the JSDoc comments, and only rely on it for IDE support, I think it's fine. VSCode seems to mostly get it (although amusingly hovering over `@type` shows you [this person's github profile](https://github.com/type) 😂)

Issue #279, opened 2019-02-13
Version documentation website


Nevon:
At the moment, the website that is deployed to https://kafka.js.org is based on latest master. However, most people will be using the latest stable version of KafkaJS, but there's no indication of which capabilities are available in that version vs what's in latest master.

Docusaurus has versioning support, so we should be able to generate one version per tag and then default to the latest stable, but I haven't looked into how to do it yet.

PR #835, opened 2020-08-11
Adding versioning to docs.

rob3000:
This PR takes a first run at https://github.com/tulios/kafkajs/issues/279.

I currently only have the following versions:
* 1.12.0 (current)
* 1.11.0
* 1.10.0

I have a question around the tags, should i also be doing the minor updates docs as well e.g 1.9.0, 1.9.1, 1.9.2?

Not sure if we will need to look at adding in a job to `azure-pipelines.yml` to allow adding a versioned document based on a tag e.g: `cd website && yarn run version ${TAG}`


Nevon:
Thanks for taking a stab at this. This is something that's been highly requested, and will reduce the number of incoming questions by quite a bit.

First, let's see if I understand Docusaurus versioning correctly.

* By running `docusaurus-version <version>` you create a new versioned set of docs, which essentially copies the contents  of `/docs` into `/website/versioned_docs/version-<version>` and adds a reference in `/pages/en/versions.json`.
* Each version of the docs will be available at `/docs/<version>/`
* The last created version (regardless of the name of the version) will be available at `/docs/` as well as `/docs/<version>`.
* The docs in `master` will be available at `/docs/next/`

If this is correct, then that seems good to me. One concern that I had was breaking existing links, but that shouldn't be the case since the current version will always be available at `/docs/`.

One thing that we'll need to address, probably, is how we generate URLs to the docs, since ideally we'd want to point to the docs for _that version_ of KafkaJS, so we'll need to read that from the package.json. This does not need to be addressed within the scope of this PR, however.
https://github.com/tulios/kafkajs/blob/7d7a765c05f419f44108e03085d80c01ebc41d79/src/utils/websiteUrl.js#L3

> I have a question around the tags, should i also be doing the minor updates docs as well e.g 1.9.0, 1.9.1, 1.9.2?

Yes, I believe we should create a version whenever we create a new (stable) release, basically. It seems a lot simpler than trying to conditionally create versions only for minor and major releases. I don't think we should create versions for the beta releases. If you're using the beta version of the package, you can just rely on `/docs/next`.

However, I don't think you need to go back and create versions for all of those now. I'm perfectly fine with just having `0.12.0` and `next` for now.

> Not sure if we will need to look at adding in a job to azure-pipelines.yml to allow adding a versioned document based on a tag e.g: cd website && yarn run version ${TAG}

That would be nice. I would say it's another job with similar conditions to the release job:

https://github.com/tulios/kafkajs/blob/master/azure-pipelines.yml#L211-L235

The steps would be similar to the website build, except that instead of invoking the Docusaurus "publish-to-gh-pages" action we would invoke the `docusaurus-version` action and then commit and push that back to master. It's a little tricky to get this right, especially since you won't have direct access to azure pipelines to test it out, so I would leave that part to us (I hear @tulios loves Azure Pipelines).

One thing you could do, however, would be to [revert this commit](https://github.com/tulios/kafkajs/commit/1a6c3d51faeb2faabcce5d85d1162aa239e1f95f), since it will no longer be necessary.

rob3000:
Hey @Nevon,

> By running docusaurus-version <version> you create a new versioned set of docs, which essentially copies the contents of /docs into /website/versioned_docs/version-<version> and adds a reference in /pages/en/versions.json

Correct.

> Each version of the docs will be available at /docs/<version>/
Correct. And you can view all the versions by visiting /versions
image

> The last created version (regardless of the name of the version) will be available at /docs/ as well as /docs/<version>.
Not really, the latest version will always be at /docs and wont be available at /docs/<version> until a newer version is published.

> The docs in master will be available at /docs/next/
Correct.

> That would be nice. I would say it's another job with similar conditions to the release job....

I thought as much :) 

I'll revert that commit now as well.

Nevon:
Let me know if there's anything else you want to address before we merge this.

rob3000:
I think that's all from me 😄

Nevon:
Then let's give it a whirl and see what happens.

PR #839, opened 2020-08-15
Improve types

Nevon:
This does two things:

## Constrain the parameter types for `cluster.fetchTopicOffsets`

`cluster.fetchTopicOffsets` now has two optional parameters, `fromBeginning` and `fromTimestamp`. In reality, these are mutually exclusive. There is no case in which you would want to pass in both. Simultaneously, there is also no case in which you would not want to pass in one of them. I.e. they are mutually exclusive but required.

```ts
// Would have compiled but failed at runtime
await cluster.fetchTopicOffsets([{
  topic: 'foo',
  partitions: [{ partition: 1 }]
}])

// Would compile but behave strangely at runtime
await cluster.fetchTopicOffsets([{
  topic: 'foo',
  partitions: [{ partition: 1 }]
  fromBeginning: true,
  timestamp: 1597492814792
}])
```

With the updated type definition, neither of those cases are possible anymore.

Screenshot 2020-08-15 at 14 04 06

Screenshot 2020-08-15 at 14 05 21

## Add JSDoc annotations for internal factory methods for Consumer, Producer and Admin

Some of these previously had return annotations but not all of them, and none of them had parameter annotations so everything was `any`, pretty much.

Unfortunately I couldn't find a way to get JSDoc to understand that `params.instrumentationEmitter` should be an instance of `InstrumentationEventEmitter`. The `typedef` in there didn't make any sense, so I removed that. It should just be a matter of importing the module and then just using the classname, but it doesn't seem to work in VSCode at least.

Also, there is absolutely no way to reference an enum in a `@param`, which I noticed when trying to type `isolationLevel`. Best I can do is `number`, which is the underlying type. You can constrain the possible values by putting literals in there (`@param {(0|1|2)} isolationLevel`), but that is ridiculous.

### DIFF
--- a/src/admin/index.js
+++ b/src/admin/index.js
@@ -47,7 +47,7 @@ const findTopicPartitions = async (cluster, topic) => {
 /**
  *
  * @param {Object} params
- * @param {import("../../types").Logger=} params.logger
+ * @param {import("../../types").Logger} params.logger
  * @param {import('../instrumentation/emitter')=} params.instrumentationEmitter
  * @param {import('../../types').RetryOptions} params.retry
  * @param {import("../../types").Cluster} params.cluster
--- a/src/admin/index.js
+++ b/src/admin/index.js
@@ -44,6 +44,16 @@ const findTopicPartitions = async (cluster, topic) => {
     .sort()
 }
 
+/**
+ *
+ * @param {Object} params
+ * @param {import("../../types").Logger=} params.logger
+ * @param {import('../instrumentation/emitter')=} params.instrumentationEmitter
+ * @param {import('../../types').RetryOptions} params.retry
+ * @param {import("../../types").Cluster} params.cluster
+ *
+ * @returns {import("../../types").Admin}
+ */
 module.exports = ({
   logger: rootLogger,
   instrumentationEmitter: rootInstrumentationEmitter,
--- a/src/admin/index.js
+++ b/src/admin/index.js
@@ -48,7 +48,7 @@ const findTopicPartitions = async (cluster, topic) => {
  *
  * @param {Object} params
  * @param {import("../../types").Logger} params.logger
- * @param {import('../instrumentation/emitter')=} params.instrumentationEmitter
+ * @param {import('../instrumentation/emitter')} [params.instrumentationEmitter]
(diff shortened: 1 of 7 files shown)


src/admin/index.js
```diff
@@ -44,6 +44,16 @@ const findTopicPartitions = async (cluster, topic) => {
     .sort()
 }
 
+/**
+ *
+ * @param {Object} params
+ * @param {import("../../types").Logger=} params.logger
```
ankon, line 50: This trailing `=` seems to make the parameter optional, but then later it is assumed it is defined:

```
const logger = rootLogger.namespace('Admin')
```

(On the nit-picky side: I find the `@param {type} [optionalParameter]` syntax easier to understand and more visible than a trailing `=` ...)

src/consumer/index.js
```diff
@@ -24,6 +24,24 @@ const specialOffsets = [
 ]
 
 /**
+ * @param {Object} params
+ * @param {import("../../types").Cluster} params.cluster
+ * @param {String} params.groupId
+ * @param {import('../../types').RetryOptions} params.retry
+ * @param {import('../../types').Logger=} params.logger
```
ankon, line 31: Same here: the logger seems to be assumed to be defined in the body.

src/admin/index.js
Nevon: > This trailing = seems to make the parameter optional, but then later it is assumed it is defined:

Whoops! For some reason my brain read `logger: rootLogger` as `logger` having a default value. 🙈

> (On the nit-picky side: I find the @param {type} [optionalParameter] syntax easier to understand and more visible than a trailing = ...)

I think I agree. I'm gonna refactor to use that syntax instead. I think I adopted the trailing equals because I wanted to add the default value, but then I decided not to because vscode doesn't use the default values anyway and it's so much duplication.

ankon: > I think I adopted the trailing equals because I wanted to add the default value, but then I decided not to because vscode doesn't use the default values anyway and it's so much duplication.

Like this: `@param {Foo} [optionalParameter=defaultValue]`?

Nevon: Fixed the syntax in `f6e215f` and the issue with the logger not being required in `048cb66`.

PR #843, opened 2020-08-19
Define the options for the runner constructor

ankon:
This is only relevant when developing, as it provides more type information (and incidentally makes it possible to ctrl-click on things to navigate! :D)

### DIFF
--- a/src/consumer/runner.js
+++ b/src/consumer/runner.js
@@ -30,7 +30,7 @@ module.exports = class Runner extends EventEmitter {
    * @param {(payload: import("../../types").EachBatchPayload) => Promise<void>} options.eachBatch
    * @param {(payload: import("../../types").EachMessagePayload) => Promise<void>} options.eachMessage
    * @param {number} [options.heartbeatInterval]
-   * @param {(reason: any) => PromiseLike<never>} [options.onCrash]
+   * @param {(reason: Error) => void} options.onCrash
    * @param {import("../../types").RetryOptions} [options.retry]
    * @param {boolean} [options.autoCommit=true]
    */
--- a/src/consumer/runner.js
+++ b/src/consumer/runner.js
@@ -20,6 +20,20 @@ const CONSUMING_START = 'consuming-start'
 const CONSUMING_STOP = 'consuming-stop'
 
 module.exports = class Runner extends EventEmitter {
+  /**
+   * @param {object} options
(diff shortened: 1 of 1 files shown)


src/consumer/runner.js
```diff
@@ -20,6 +20,20 @@ const CONSUMING_START = 'consuming-start'
...
+   * @param {object} options
+   * @param {import("../../types").Logger} options.logger
+   * @param {import("./consumerGroup")} options.consumerGroup
+   * @param {import("../instrumentation/emitter")} options.instrumentationEmitter
+   * @param {boolean} [options.eachBatchAutoResolve=true]
+   * @param {number} [options.partitionsConsumedConcurrently]
+   * @param {(payload: import("../../types").EachBatchPayload) => Promise<void>} options.eachBatch
+   * @param {(payload: import("../../types").EachMessagePayload) => Promise<void>} options.eachMessage
+   * @param {number} [options.heartbeatInterval]
+   * @param {(reason: any) => PromiseLike<never>} [options.onCrash]
```
Nevon, line 33: Pretty sure `reason` should be `Error`.

ankon: Good one. I think I copied that one from the suggestions, and didn't do much checking. Thinking about it further we probably should define what we expect a crash handler to do:

* Setting the parameter to `Error` seems reasonable, with the _minuscule_ risk that somewhere deep down the stack some misbehaved library rejects a promise with a non-error thing. 
* It seems that the return value of onCrash handler is never looked at, so maybe `void` is enough?

```
@param {(reason: Error) => void} [options.onCrash]
```

WDYT?

Nevon: Completely agree.

ankon: Based on the types in lib.es5.d.ts: "Anything goes, really": 

```
    catch<TResult = never>(onrejected?: ((reason: any) => TResult | PromiseLike<TResult>) | undefined | null): Promise<T | TResult>;
```

I do notice that onCrash isn't optional though: in `scheduleJoin` undefined would be ok, but in `start` just below we call it unconditionally.

Nevon: > "Anything goes, really"

That should be the slogan for Javascript 😄 

> I do notice that onCrash isn't optional though: in scheduleJoin undefined would be ok, but in start just below we call it unconditionally.

Right, yes, it should not be optional.

PR #851, opened 2020-08-28
Fix fall-back retry config for producer

pawelrychlik:
See https://github.com/tulios/kafkajs/blob/master/src/producer/index.js#L56
`this.retry` in cluster constructor was not assigned, so referring to `cluster.retry` from producer constructor will always yield undefined.


Nevon (approved):

Nevon:
Thanks for the fix.

PR #854, opened 2020-09-02
Broker rediscovery with async getConnection method

pimpelsang:


pimpelsang:
Fair criticism! I agree with the brokers parameter value to be either list or async method to get that list and nothing weird like I proposed initially. I've already updated code and it does look better.

For the other part of the question - why the need to have sasl parameter dynamic? It's not about having different credentials per kafka host, but more about getting credentials asynchronously from another service and the possibility that those credentials could get changed during runtime.

pimpelsang:
Fixed tests by reverting to original broker rotation instead of taking random

src/cluster/__tests__/connectionBuilder.spec.js
```diff
@@ -42,18 +42,18 @@ describe('Cluster > ConnectionBuilder', () => {
...
   })
 
-  test('when called without host and port iterates throught the seed brokers', () => {
-    const connections = Array(brokers.length)
-      .fill()
-      .map(() => {
-        const { host, port } = builder.build()
-        return `${host}:${port}`
-      })
+  test('when called without host and port picks random seed broker', async () => {
```
Nevon, line 45: Isn't this exactly the same as before, except with a for-loop instead of a map?

src/cluster/connectionBuilder.js
```diff
@@ -1,10 +1,9 @@
 const Connection = require('../network/connection')
-const { KafkaJSNonRetriableError } = require('../errors')
-const shuffle = require('../utils/shuffle')
+const { KafkaJSConnectionError } = require('../errors')
 
 const validateBrokers = brokers => {
   if (!brokers || brokers.length === 0) {
-    throw new KafkaJSNonRetriableError(`Failed to connect: expected brokers array and got nothing`)
+    throw new KafkaJSConnectionError(`Failed to connect: expected brokers array and got nothing`)
```
Nevon, line 6: I don't think this should be a `KafkaJSConnectionError`, since it's actually a validation error that's not retriable. `KafkaJSConnectionError` is retriable by default. I see three options:

1. Keep it retriable and live with the fact that it's weird to validate the same static list N times.
2. Make `validateBrokers` a bit smarter and have it throw a retriable error if `brokers` is a function and a non-retriable error if it's not.
3. Wrap `brokers` in an async function that does retrying internally instead. Something along the lines of

```js
const retrier = createRetry({ ...retry })
const getBrokers = () => retrier(async bail => {
  const static = Array.isArray(brokers)
  const list = static ? brokers : await brokers()
  try {
    validateBrokers(list) // Throws retriable error
    return list
  } catch (err) {
    if (static) { // if it's a static list, don't bother retrying
      bail(err)
    } else {
      throw err
    }
  }
})
```

src/cluster/connectionBuilder.js
```diff
@@ -22,27 +21,26 @@ module.exports = ({
...
+
+        host = randomBroker.split(':')[0]
+        port = Number(randomBroker.split(':')[1])
       }
 
       return new Connection({
         host,
         port,
         rack,
+        sasl: typeof sasl === 'function' ? await sasl() : sasl,
```
Nevon, line 42: Could we remove the sasl renewal functionality from this PR and handle it separately instead. I don't feel like I've fully thought through the ramifications of this, and once we make it a public interface, it's difficult to change if we later realize it should change.

For example, I'm not sure that on connection build is the right time to do this. If this call fails, how should we handle that? It may very well be that this is the way we should handle it, but I'd rather not have to merge both changes in the same PR.

types/index.d.ts
```diff
@@ -14,10 +14,13 @@ export class Kafka {
   logger(): Logger
 }
 
+export type BrokersFunction = () => Promise<string[]>;
```
Nevon, line 17: The tests seem to indicate that this could also be a non-async function, so a more correct type would be 

```ts
export type BrokersFunction = () => string[] | Promise<string[]>
```

src/cluster/__tests__/connectionBuilder.spec.js
pimpelsang: Yes it is except for the loop can handle async/await

src/cluster/connectionBuilder.js
pimpelsang: I suggest second option

src/cluster/connectionBuilder.js
pimpelsang: yes, removed

types/index.d.ts
pimpelsang: Indeed, fixed

src/cluster/connectionBuilder.js
Nevon: Sounds good!

src/cluster/connectionBuilder.js
```diff
@@ -1,11 +1,18 @@
...
-const shuffle = require('../utils/shuffle')
+const { KafkaJSConnectionError, KafkaJSNonRetriableError } = require('../errors')
 
-const validateBrokers = brokers => {
-  if (!brokers || brokers.length === 0) {
-    throw new KafkaJSNonRetriableError(`Failed to connect: expected brokers array and got nothing`)
+const getBrokers = async brokers => {
+  const dynamic = typeof brokers === 'function'
+
+  const list = dynamic ? await brokers() : brokers
```
Nevon, line 7: Make sure to catch this so that you can wrap the error in a `new KafkaJSConnectionError(error)` if it rejects.

pimpelsang: yes, updated getBrokers method to be more explicit

Nevon:
Overall this looks good to me. But note that ##878 adds a new `removeBroker` method to the BrokerPool that will need to be updated now that `brokers` can be a function rather than an array.

pimpelsang:
I think it's fine - this.brokers inside BrokerPool is always array. Only inside connectionBuilder it can be function.

Nevon:
Ah yes, you're right. My mistake. 👍

Nevon (approved):

Nevon:
Thanks for the contribution! I'll go about updating the documentation separately.

Issue #849, opened 2020-08-24
Ensure fair allocation of fetch responses across partitions


Nevon:
When receiving a Fetch request, the broker processes partitions in order they appear in request. Otherwise, for each partition except the first one server fetches up to corresponding partition limit max_bytes, but not bigger than remaining response limit. For the first partition, server always fetches at least one message.

This means that if we make a fetch request for 1 megabyte across 3 partitions, we may receive data for only 1 partition and none for the rest. This can be circumvented by configuring the client with a higher `maxBytes` than `maxBytesPerPartition`:

```
const numPartitions = 3
const maxBytes = 1000000
const maxBytesPerPartition = maxBytes / numPartitions
```

However, in the case of unbalanced partitions where only one partition has data available, you would receive a maximum of `maxBytes/numPartitions` per fetch, which would be wasteful.

A better approach is to do what they recommend in the KIP and introduce a shuffling mechanism so that it's not always the same partition that is prioritized.

https://cwiki.apache.org/confluence/display/KAFKA/KIP-74%3A+Add+Fetch+Response+Size+Limit+in+Bytes

> Since new fetch request processes partitions in order and stops fetching data when response limit is hit, client should use some kind of partition shuffling to ensure fairness.
> 
> Consider following example - suppose client want to fetch from 4 partitions: A, B, C, D. Suppose that partitions A and B are growing much faster than C and D.
> 
> If client is always fetching partitions in order A,B,C,D than it is possible that response limit is hit before any messages were fetched from C and D.
> 
> In this scenario client won't  get any messages from C and D until it catches up with A and B.
> 
> The solution is to reorder partitions in fetch request in round-robin fashion to continue fetching from first empty partition received or to perform random shuffle of partitions before each request.

PR #859, opened 2020-09-08
Ensure fair fetch response allocation across topic-partitions

Nevon:
See #849 for details, but this fixes a big issue with uneven consumption when the consumer has fallen behind on several partitions and `maxBytesPerPartition * numPartitions > maxBytes`.

### DIFF
--- a/src/broker/__tests__/fetchPartitionAllocation.spec.js
+++ b/src/broker/__tests__/fetchPartitionAllocation.spec.js
@@ -49,18 +49,27 @@ describe('Broker > Fetch', () => {
    */
   test('it should randomize the order of the requested topic-partitions', async () => {
     const topics = [createFetchTopic('topic-a', 3), createFetchTopic('topic-b', 2)]
+    const flattened = [
+      { topic: topics[0].topic, partition: topics[0].partitions[0] },
+      { topic: topics[0].topic, partition: topics[0].partitions[1] },
+      { topic: topics[0].topic, partition: topics[0].partitions[2] },
+      { topic: topics[1].topic, partition: topics[1].partitions[0] },
+      { topic: topics[1].topic, partition: topics[1].partitions[1] },
+    ]
 
     mockShuffle.mockImplementationOnce(() => [
-      { topic: topics[0].topic, partitions: [topics[0].partitions[0]] },
-      { topic: topics[1].topic, partitions: [topics[1].partitions[0]] },
+      flattened[0],
+      flattened[3],
       // test that consecutive partitions for same topic are merged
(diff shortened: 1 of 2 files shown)


src/broker/__tests__/fetchPartitionAllocation.spec.js
```diff
@@ -0,0 +1,83 @@
...
+   * the response will be allocated in the order
+   * the topic-partitions appear in the request. To avoid biasing consumption,
+   * the order of all topic-partitions is randomized on each request.
+   *
+   * @see https://cwiki.apache.org/confluence/display/KAFKA/KIP-74%3A+Add+Fetch+Response+Size+Limit+in+Bytes
+   */
+  test('it should randomize the order of the requested topic-partitions', async () => {
+    const topics = [createFetchTopic('topic-a', 3), createFetchTopic('topic-b', 2)]
+
+    mockShuffle.mockImplementationOnce(() => [
```
Nevon, line 57: I don't really love this test, since I had to module mock `shuffle` - i.e. reach into the internals of `broker.fetch` - but it's hard to assert on things that have randomness, and in this case I really do want a specific order to check that consecutive partitions for the same topic get merged.

JaapRood: Totally fair I'd say. My best attempt would asserting that the order is different, but even that's not guaranteed: ending up with the same order is still a valid shuffle result. Mocking this order is the closest I imagine you could get to verifying the logic in `broker.fetch`.

JaapRood (approved):
Looks good to me!

src/broker/__tests__/fetchPartitionAllocation.spec.js
JaapRood: Thinking about this a tiny bit more in case you'd really didn't like the test, I guess you could spy on `shuffle` instead of mocking it, so you'd know the order of topics it returned when called by `broker.fetch`. That would verify the interface between those two to work correctly, but since the surface area of that is so small and straightforward, I'm pretty sure that I wouldn't bother.

Nevon: The problem with that is that I need shuffle to actually return a very specific result in order to be able to validate that consecutive partitions for the same topic are merged. Meaning if we have topic-partitions:

```
topic-a | 1
topic-a | 2
topic-b | 1
topic-b | 2
```

And after shuffling it looks like:

```
topic-a | 1
topic-b | 1
topic-b | 2
topic-a | 2
```

We want the resulting request to be:

```
topic-a | 1
topic-b | 1, 2
topic-a | 2
```

PR #869, opened 2020-09-11
Add protocol produce v6

tulios:


src/protocol/requests/produce/v6/request.js
```diff
@@ -0,0 +1,38 @@
+const requestV3 = require('../v3/request')
```
Nevon, line 1: I don't think we should "skip" versions when we delegate to an older version, even if the versions in between are in turn also delegating, because then if we change something in an in between version that change will be lost. It might not matter in this specific case, without having looked at v4 and v5, but if we are consistent in always delegating v7 -> v6 -> v5 etc. there's no chance of that issue happening.

Nevon: Plus it saves the reviewer from having to go look in all those in-between versions to check that there isn't anything going on there that we're missing :P

tulios: Sure, I can change the reference.

src/protocol/requests/produce/v6/request.js
```diff
@@ -0,0 +1,38 @@
+const requestV3 = require('../v5/request')
```
Nevon, line 1: My dude, now I'm even more triggered 😂

tulios: 🙈

ankon:
This is a change related to [KIP-219](https://cwiki.apache.org/confluence/display/KAFKA/KIP-219+-+Improve+quota+communication), right? 

In that case I think the `throttleTime` should be changed to `clientSideThrottleTime` here, so that the throttling will be done. See https://github.com/tulios/kafkajs/pull/776, specifically https://github.com/tulios/kafkajs/pull/776/commits/29779f276556b966eb76b2464a5970ced0564f34

tulios:
Yes, I noticed that. I made the change locally but we have to change the client to wait for the pending requests when disconnecting otherwise we could lose data with the producer. I am on it.

tulios:
@Nevon @ankon I updated the PR to use the client-side throttling and for the request queue to wait for pending requests before it shut down and disconnect.

tulios:
There is one more detail, the request queue has to wait for the current inflight requests. I am making the change.

tulios:
I spot a bug in the way the request queue process pending requests
https://github.com/tulios/kafkajs/blob/c025e05725a8b1badf813ea6a5231d0b8f77647a/src/network/requestQueue/index.js#L232
it is processing that most recent request in the queue rather than the oldest one, I am switching from `this.pending.pop()` to `this.pending.shift()` since we want FIFO behavior.

tulios:
I ran many tests and I am now happy with the solution, please take a look.

JaapRood (approved):
Looks solid. Reading through the KIP and the comments on this PR makes me think you've covered all the scenarios, most importantly covering the fact that throttled messages should be flushed before disconnecting. Code and tests backs this up!

src/network/connection.js
```diff
@@ -317,8 +324,13 @@ module.exports = class Connection {
 
     try {
       const payloadDecoded = await response.decode(payload)
-      // KIP-219: If the response indicates that the client-side needs to throttle, do that.
+
+      /**
```
ankon, line 328: Somewhat nit-picky, but this line wouldn't be a JSDoc-y comment, but just a multiline comment, no?

src/cluster/__tests__/fetchTopicsOffset.spec.js
```diff
@@ -38,10 +38,12 @@ describe('Cluster > fetchTopicsOffset', () => {
       createPartitioner: createModPartitioner,
     })
 
+    await producer.connect()
     await sendSampleMessages()
   })
 
   afterEach(async () => {
+    await producer.disconnect()
```
ankon, line 46: This should use the same protection against `producer` not being assigned, to avoid noisy errors when tests fail due to problems with the cluster (see https://github.com/tulios/kafkajs/pull/795)

ankon:
LGTM!

src/network/connection.js
tulios: I like the doc block because it highlights the comment. I made the change to give the importance this comment deserve; I don't think we should argue about the comment block style; my intention was just to highlight a good comment 😄

PR #871, opened 2020-09-14
Update examples with pretty logger

tulios:
As someone that uses the examples a lot to test PRs, use cases, bugs, etc I would appreciate a better output. The pretty format helps a lot by highlighting the important parts and making it easier to spot what I want to test.

Producer:
producer_example

Consumer:
consumer_example


JaapRood:
👏 always great type of enhancement to make!

PR #883, opened 2020-09-22
Move compression to worker threads if available

tulios:
This is experimental work ⚠

The idea is to move the compression and decompression work to worker threads. This change could potentially improve performance as the operations would not block the event loop. This is still pending a proper benchmark, which will happen during the course of this draft.

* Note 1: worker threads require Node.js >= 12.11
* Note 2: If this is successful we could potentially move the encoding and decoding bits to worker threads.


Nevon:
Closing this since we didn't see any improvement and we're not planning on merging this.

PR #887, opened 2020-09-24
Include "restart" property in consumer.crash instrumentation event

### LINKED ISSUE #885: Include "restart" property in consumer.crash instrumentation event payload
When the consumer crashes, the payload contains the error that caused the crash. However, the logic for determining whether or not the consumer will restart itself is opaque to the user so it's not possible to know from the crash event whether or not the crash was fatal (i.e. it's not as simple as just checking if the error is retriable).

Not having access to this information makes it hard for the user to gracefully shut down the application, initiate their own restart or react in some other way to the fact that the consumer will no longer be running.

Add a `restart: Boolean` property to the crash payload to indicate whether or not the consumer will restart.

**Additional context**

This came from a Slack support request: https://kafkajs.slack.com/archives/CF6RFPF6K/p1600859082002600

> My consumers stop and fail to restart due to KafkaJSProtocolError: Replication-factor is invalid error(screenshot 1), which  I guess is the expected behaviour, because this is considered a non-retryable error(screenshot 2).
> My question is is there a good strategy to handle this crash on application level and restart the consumer?

tulios:
fixes #885


Nevon (approved):

Nevon:
Don't forget to add the property in the payload documentation for the crash event.

Nevon (approved):

PR #892, opened 2020-09-25
Fix flaky tests

Nevon:
Fixing two of the worst offenders according to Azure Devops Pipelines. The commit messages have information on why they were flaky.


tulios:
I noticed the same on the error recovery test, I fixed one of them, but the other one started working, so I left it alone. Great job.

tulios (approved):

PR #895, opened 2020-09-27
Feat: add option for admin.fetchOffsets to resolve the offsets

### LINKED ISSUE #205: Add option to make admin.fetchOffsets resolve the offsets

artysidorenko:
Closes #205

Had a stab at issue #205 as I don't think anyone was working on it any more.

Tweaking the admin to resolve the offsets I think was fairly quick, but let me know if you had other ideas for the implementation.

I might have gone a little OTT with some of the tests 🙈. Wanted to include test cases for:

(a) what happens when you fetch/resolve without having reset the consumer group offsets: it should return the consumer's latest offsets rather than the topic's highwatermark. That's where I'm sending two separate message batches and fetching the offsets before the second batch, and

(b) what happens when low watermark is greater than 0. Am simulating this by deleting some messages from the topic. I couldn't find that the kafkajs client implemented `deleteRecords` yet? (not sure if it's on the roadmap for the future, or a decision to leave it out, or I might have missed it), so I wrote a helper for the test to do it.

Let me know if you want me to simplify any of these.. (also, if you don't have deleteRecords but do want to implement it in the main client I'm happy to look into that as well)

### DIFF
--- a/src/admin/__tests__/fetchOffsets.spec.js
+++ b/src/admin/__tests__/fetchOffsets.spec.js
@@ -10,7 +10,10 @@ const {
   waitForConsumerToJoinGroup,
   generateMessages,
   waitForMessages,
+  testIfKafkaAtLeast_0_11,
 } = require('testHelpers')
+const Encoder = require('../../protocol/encoder')
+const Decoder = require('../../protocol/decoder')
 
 describe('Admin', () => {
   let admin, cluster, groupId, logger, topicName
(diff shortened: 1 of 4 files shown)


tulios:
Hi @artysidorenko, thanks for the PR. Adding this flag is something that I wanted for a while but end up forgetting about 😅, so it's definitely something we would like to have. We never had a need for `deleteRecords` and that's why we don't have it yet, but we would gladly review a PR adding the feature.

src/admin/__tests__/fetchOffsets.spec.js
```diff
@@ -64,5 +72,221 @@ describe('Admin', () => {
...
+    })
   })
 })
+
+/**
+ * Manually delete records up to the selected offset
+ * http://kafka.apache.org/protocol.html#The_Messages_DeleteRecords
+ * Only available from Kafka 0.11 upwards
+ */
+const deleteRecords = async ({ cluster, topicName, topicDetails }) => {
```
Nevon, line 235: I definitely think we should move this into the supported protocols rather than having it here, but we can certainly do that in a separate PR rather than bloating this one.

src/admin/index.js
```diff
@@ -355,7 +361,7 @@ module.exports = ({
    * @param {string} topic
```
Nevon, line 361:
```suggestion
   * @param {string} topic
   * @param {boolean} [resolveOffsets=false]
```

Nevon (changes requested):
Great contribution, thanks a lot! Don't forget to update the type definition for `Admin.fetchOffsets` in `types/index.d.ts`

artysidorenko:
> Great contribution, thanks a lot! Don't forget to update the type definition for `Admin.fetchOffsets` in `types/index.d.ts`

Cool thanks :) I'll have a go at building `deleteRecords` into the protocols, and made a note to re-use it in here/clean it up once that's done.

src/admin/index.js
```diff
@@ -368,12 +375,31 @@ module.exports = ({
...
 
-    return responses
+    if (resolveOffsets) {
+      const indexedOffsets = indexByPartition(await fetchTopicOffsets(topic))
+      consumerOffsets = consumerOffsets.map(({ topic, partitions }) => ({
+        topic,
+        partitions: partitions.map(({ offset, partition, ...props }) => ({
+          partition,
+          offset:
+            Number(offset) === EARLIEST_OFFSET
```
tulios, line 390: Can you change this to not be a double ternary?

tulios (changes requested):
Looks good to me, nice to finally have this in. Can you just fix the double ternary and we are ready to merge.

src/admin/index.js
```diff
@@ -368,12 +375,31 @@ module.exports = ({
...
+              ? indexedOffsets[partition].low
+              : Number(offset) === LATEST_OFFSET
+              ? indexedOffsets[partition].high
+              : offset,
+          ...props,
+        })),
```
artysidorenko, line 396:
```suggestion
        partitions: partitions.map(({ offset, partition, ...props }) => {
          let resolvedOffset = offset
          if (Number(offset) === EARLIEST_OFFSET) {
            resolvedOffset = indexedOffsets[partition].low
          }
          if (Number(offset) === LATEST_OFFSET) {
            resolvedOffset = indexedOffsets[partition].high
          }
          return {
            partition,
            offset: resolvedOffset,
            ...props,
          }
        }),
```

@tulios, checking before I commit, if this is what you had in mind to make it more readable, or you were thinking something different?

tulios: Yes, the suggestion works fine.

tulios (approved):

Nevon (approved):

Issue #886, opened 2020-09-24
Improve TS typing of error class


lafeuil:
The TS typing does not describe the properties of error classes. 

For example with the `KafkaJSProtocolError` class. This error class have two properties : `type` and `code`. This two properties are set in the constructor.

```javascript
class KafkaJSProtocolError extends KafkaJSError {
  constructor(e) {
    super(e, { retriable: e.retriable })
    this.type = e.type
    this.code = e.code
    this.name = 'KafkaJSProtocolError'
  }
}
```

But, in the TS typing, this properties does not appear :

```typescript
export class KafkaJSProtocolError extends KafkaJSError {
  constructor(e: Error | string)
}
```

I think this can be fix by declaring readonly property in KafkaJSProtocolError : 

```typescript
export class KafkaJSProtocolError extends KafkaJSError {
  readonly code: number;
  readonly type: string;
  constructor(e: Error | string)
} 
```

Nevon:
Fixed by #900

PR #900, opened 2020-10-02
Added properties to error classes typescript types

rob3000:
Adding error types as outlined in: https://github.com/tulios/kafkajs/issues/886

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -859,6 +859,11 @@ export var CompressionCodecs: {
 }
 
 export class KafkaJSError extends Error {
+  readonly message: Error["message"];
+  readonly name: string;
+  readonly retriable: boolean;
+  readonly helpUrl: string;
+
   constructor(e: Error | string, metadata?: KafkaJSErrorMetadata)
 }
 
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -751,6 +751,11 @@ export var CompressionCodecs: {
 }
 
 export class KafkaJSError extends Error {
+  readonly message: Error["message"];
+  readonly name: string;
+  readonly retriable: boolean;
+  readonly helpUrl: string;
```
Nevon, line 757: `retriable` and `helpUrl` may not be set, so they should be nullable, right?

```suggestion
  readonly retriable?: boolean;
  readonly helpUrl?: string;
```

types/index.d.ts
```diff
@@ -827,6 +851,7 @@ export class KafkaJSUnsupportedMagicByteInMessageSet extends KafkaJSError {
 }
 
 export class KafkaJSDeleteGroupsError extends KafkaJSError {
+  readonly groups: any[];
```
Nevon, line 854: This should be:

```suggestion
  readonly groups: DeleteGroupsResult[];
```

Except the `DeleteGroupsResult` is actually not correctly typed. It should be:

```
export type DeleteGroupsResult = {
  groupId: string
  errorCode?: number
  error?: KafkaJSProtocolError
}
```

rob3000: will fix up `DeleteGroupsResult` type as well.

types/index.d.ts
rob3000: `retriable` looks to be set as true by default: https://github.com/tulios/kafkajs/blob/master/src/errors.js#L2

azure-pipelines.yml
```diff
@@ -259,7 +259,6 @@ jobs:
         env:
           VERSION: $(Build.SourceBranch)
           GH_TOKEN: $(GH_TOKEN)
-          CONTACT_FORM_URL: $(CONTACT_FORM_URL)
```
Nevon, line 262: Something weird is going on here. Looking at the commit history, I can see two of my commits that are already in master, and this diff doesn't make sense either since this is already what master looks like. 😕

types/index.d.ts
```diff
@@ -935,6 +960,7 @@ export class KafkaJSUnsupportedMagicByteInMessageSet extends KafkaJSError {
 }
 
 export class KafkaJSDeleteGroupsError extends KafkaJSError {
+  readonly groups: DeleteGroupsResult[];
```
Nevon, line 963: Don't forget to update the `DeleteGroupsResult` interface.

azure-pipelines.yml
rob3000: hmm, i rebased from master. Let me fix it up.

rob3000: Fixed up the commit history which now doesn't include this file.

Nevon (approved):

Nevon:
Awesome! Thanks for the contribution.

Issue #789, opened 2020-06-25
Transaction documentation does not mention transactionalId


Nevon:
In order to create a transaction, the producer must be created with a transactionalId. However, this is not mentioned anywhere in the documentation for transactions.

See: https://kafka.js.org/docs/transactions

senseluo:
oh yes i didn't find it as well

megakoresh:
Yes, still not documented.

eliw00d:
Documentation and examples would be great! In the meantime, has anyone gotten transactions to work in a way that supports multiple instances of an app? I think Streams and other implementations use a combination of topic, group, and partition for the transactional.id, so knowing how we can get that to work would be awesome.

Nevon:
Fixed by #902

PR #902, opened 2020-10-02
Added transactionalId to transactions documentation

rob3000:
Adding `transactionalId` config to transactions documentation for https://github.com/tulios/kafkajs/issues/789


Nevon:
Thanks for giving this a shot!

While at least mentioning that a transactionalId needs to be set is a good start, I think we might have to be a little more helpful when it comes to setting the transactionalId, because it's not that easy to know what a good value is (and most likely the example value you set in the doc would be a bad one).

[This article from Confluent](https://www.confluent.io/blog/transactions-apache-kafka/) is pretty good at explaining how transactions work, and give some hints as to what a good value for what they're calling `transactional.id` is.

> [...] each transactional producer be assigned a unique identifier called the transactional.id. This is used to identify the same producer instance across process restarts.
> 
> The API requires that the first operation of a transactional producer should be to explicitly register its transactional.id with the Kafka cluster. When it does so, the Kafka broker checks for open transactions with the given transactional.id and completes them. It also increments an epoch associated with the transactional.id. The epoch is an internal piece of metadata stored for every transactional.id.
> 
> Once the epoch is bumped, any producers with same transactional.id and an older epoch are considered zombies and are fenced off, ie. future transactional writes from those producers are rejected.

So basically, it's a way for the broker to know that any future writes from a producer with that particular transactional id should be disregarded unless the epoch is a certain value, to avoid duplicate messages being produced by both the old instance and the new instance.

It's important to not set it to a random string, as then a new instance won't have the same transactional id and any open transactions that were left behind by the old producer will never be closed (I assume they time out at some point, but still).

In the case of a stream processor, where you're consuming from one topic and producing to another, a good candidate would be the topic-partition combination. For example: `producer-${topic}-${partition}`, since that will ensure that if some other consumer is assigned that topic-partition, the previously assigned one will not be able to produce any messages. That's essentially what kafka-streams does internally (I don't know what's up with the `ALPHA` and `BETA`, but `TaskId` is from my understanding corresponding to a topic-partition):

https://github.com/apache/kafka/blob/20e4a74c3579837df7b46ce1a5dcea37b6e1e452/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsProducer.java#L88-L123

[The Java Kafka client has terrible documentation for this](https://kafka.apache.org/documentation/#transactional.id)

So, my suggestion would be that we add a "Choosing a `transactionalId`" section just below where we say something like the following.

> ### Choosing a `transactionalId`
> 
> The `transactionalId` allows Kafka to fence out zombie instances by rejecting writes from producers with the same `transactionalId`, allowing only writes from the most recently registered producer. To ensure EoS semantics in a stream processing application, it is important that input topics and partitions in the read-process-write cycle is always the same for a given `transactionalId`.
> 
> The simplest way to achieve this is to encode the topic and partition in the `transactionalId` itself, for example: `"application-" + topic + "-" + partition`.
>
> [This article from Confluent](https://www.confluent.io/blog/transactions-apache-kafka/) goes into much greater detail on how transactions work and is a recommended read before deciding on a `transactionalId`.

docs/Transactions.md
```diff
@@ -17,14 +17,18 @@ You initialize a transaction by making an async call to `producer.transaction()`
 > - The producer must wait for acknowledgement from all replicas (acks=-1)
 > - The producer must have unlimited retries
 
-Configure the producer client with `maxInFlightRequests: 1` and `idempotent: true` to guarantee EOS. Configuring the two options will enable the settings mentioned above.
+Configure the producer client with `maxInFlightRequests: 1`, `idempotent: true` and a `transactionalId: 'name'` to guarantee EOS. Configuring the options will enable the settings mentioned above.
```
Nevon, line 20: Make sure that the value here corresponds to the one down in the example code, so that it's easy for people to make the connection. `transactional-producer` is probably a better example value than `name`.

PR #905, opened 2020-10-03
Feat: implement Admin.deleteTopicRecords

artysidorenko:
Not linked to an issue but was briefly discussed in PR #895 

Ended up being a fair few things to add/tests to write, but I tried to follow existing patterns where possible so hopefully not too out-of-the-ordinary 🤞 

**Recap:**
- Add `deleteRecords` to the protocols. I only went up to v1... there's also a v2 but I don't think it has anything of value, just this new `TAG_BUFFER` field I'm not familiar with.
- Add `deleteRecords` to the broker. 
- Implement as `deleteTopicRecords` in the admin. Called `deleteTopicRecords` instead of `deleteRecords` as it works on a single topic at a time (otherwise the logic becomes messy if you have to pass an array with multiple topics and nested multiple partitions; I think deleting from a single topic at a time should be sufficient for most people)

The test cases should describe most of the behaviours in detail, but there's a couple of things I wanted to highlight, as I made some assumptions about what "should" happen in these scenarios (let me know if you disagree):
- if one of the passed-in partition offsets is out-of-range, Kafka would throw an error anyway, so I've set it up to throw pre-emptively before sending any request (to avoid a possible confusing situation where it fails half-way through, after you've already deleted from one or more partitions).
- if one of the passed-in partitions has no broker/does not exist, it's a no-op and we log a warning and ignore. Perhaps an outright error is better though?
- if one of the passed-in partition offsets is below the LWM (but still above 0), Kafka allows this... it just means nothing gets deleted on that partition. So we log a warning.


src/admin/__tests__/deleteTopicRecords.spec.js
```diff
@@ -0,0 +1,269 @@
...
+  test('in case partition does not exist/has no lead broker, skip the partition with a warning', async () => {
+    // try to add to the delete request a partition that doesn't exist
+    recordsToDelete = [
+      { partition: 0, offset: '7' },
+      { partition: 2, offset: '5' },
+    ]
+    await admin.deleteTopicRecords({ topic: topicName, partitions: recordsToDelete })
+
+    expect(logger.warn).toHaveBeenCalledTimes(1)
+    expect(logger.warn).toHaveBeenCalledWith('Could not find broker/leader for the partition', {
```
Nevon, line 249: I think the "broker/leader" wording is a bit confusing. Just the word "leader" is better, as we already use this terminology for the "LEADER_NOT_AVAILABLE" error: "There is no leader for this topic-partition as we are in the middle of a leadership election."

src/admin/index.js
```diff
@@ -990,6 +994,86 @@ module.exports = ({
...
+        return obj
+      },
+      {}
+    )
+
+    const retrier = createRetry(retry)
+    return retrier(async () => {
+      for (const [nodeId, { topic, partitions }] of entries(seekEntriesByBroker)) {
+        const broker = await cluster.findBroker({ nodeId })
+        await broker.deleteRecords({ topics: [{ topic, partitions }] })
```
Nevon, line 1070: I think we need to handle some error codes here. Specifically if we have stale metadata, so that we try to make a request to a broker that used to be the leader for a partition, but is not anymore.

[`sendMessages.js`](https://github.com/tulios/kafkajs/blob/master/src/producer/sendMessages.js) is a good example of this: https://github.com/tulios/kafkajs/blob/02bc8a3c19504d182df131914ee9af4661bf060c/src/producer/sendMessages.js#L121-L138

src/admin/index.js
```diff
@@ -990,6 +994,86 @@ module.exports = ({
...
+      topic,
+      partitions.map(p => p.partition)
+    )
+
+    const partitionsFound = flatten(values(partitionsByBroker))
+    const topicOffsets = await fetchTopicOffsets(topic)
+
+    partitions.forEach(({ partition, offset }) => {
+      // warn if no broker/leader found for partition
+      if (!partitionsFound.includes(partition)) {
```
Nevon, line 1027: I think I disagree with this choice, as it means that without human intervention you would never know that the operation failed. If we fail to find a leader for the partition, I think it's preferable that the operation fails rather than partially completing without knowing what partitions actually failed. Otherwise, how do you react to this in a programmatic way?

src/admin/index.js
```diff
@@ -990,6 +994,86 @@ module.exports = ({
...
+        logger.warn('Could not find broker/leader for the partition', {
+          topic,
+          partition,
+        })
+        return
+      }
+      const { high, low } = topicOffsets.find(p => p.partition === partition)
+      // throw in case of offset out-of-range (otherwise could fail mid-process)
+      if (parseInt(offset) > parseInt(high) || parseInt(offset) < -1) {
```
Nevon, line 1036: Do we really need all these checks? AFAIK, they are already handled on the broker side. I'm not sure that I see a compelling enough reason to handle them on the client side when we already have to deal with partial failures regardless. The response should already contain all the information needed to know which partitions failed and for what reason.

src/protocol/requests/deleteRecords/v1/response.js
```diff
@@ -0,0 +1,13 @@
...
+ * DeleteRecords Response (Version: 1) => throttle_time_ms [topics]
+ *  throttle_time_ms => INT32
+ *  topics => name [partitions]
+ *    name => STRING
+ *    partitions => partition_index low_watermark error_code
+ *      partition_index => INT32
+ *      low_watermark => INT64
+ *      error_code => INT16
+ */
+module.exports = responseV0
```
Nevon, line 13: In V1 the broker is returning the response before throttling the client in case the client has exceeded the quota. This means that we should be throttling the client on the client side, instead of relying on the server to do it for us.

See [this comment](https://github.com/tulios/kafkajs/pull/697#discussion_r498360020) for more information and an example. It's a very simple change.

Nevon (changes requested):
Another quality contribution! 👍 

Lots of notes, but overall I think it's a great contribution. I'm glad to see you went back and addressed the bit of tech debt we took on in the previous PR.

Nevon:
> if one of the passed-in partition offsets is below the LWM (but still above 0), Kafka allows this... it just means nothing gets deleted on that partition. So we log a warning.

I agree with this choice. In a way, the caller's request has already been fulfilled if the lwm is below the requested offset.

Nevon:
> if one of the passed-in partition offsets is out-of-range, Kafka would throw an error anyway, so I've set it up to throw pre-emptively before sending any request (to avoid a possible confusing situation where it fails half-way through, after you've already deleted from one or more partitions).

Hmm, I see. I understand your thinking here. Since we would reject with an error as soon as one of the responses come back with an error, the caller wouldn't necessarily be able to tell which offsets were deleted.

Let's think about it for a bit and figure out what the best outcome for the user is. Say I'm deleting records from 3 different partitions that are on 3 different leaders. One of the requests fail because the requested offset is out of range.

```js
// Say we fail to delete partition 2
await admin.deleteTopicRecords({ topic, partitions: [
  { partition: 1, offset: 10 },
  { partition: 2, offset: 15 },
  { partition: 3, offset: 12 }
]})
```

Now how do I deal with this as a user? If we do the early check on the client side, I get a `KafkaJSOffsetOutOfRange` telling me that partition 2 offset 15 is not maintained by this server. Partition 1 and 3 have not been modified (the only reason I know this is because I read the code though). So I can fetch offsets and re-make the request with the correct offset. The downside is some duplication of logic that already lives on the broker, but overall not terrible. It's a bit annoying that if partition 3 is also out of range, I'll get another error and have to repeat the process, but again, not terrible.

If we let the broker deal with it, the call will reject with the same error as above. Assuming that the requests were made in the same order that we specified the partitions in, partition 1 would have been processed and partitions 2 and 3 not, but there's no way for the user to know that. Not great.

A third option is to let the broker deal with it but aggregate the errors ourselves. Basically we throw the requests into `Promise.allSettled` and if any of them errored, we reject with our own error that wraps the protocol errors. Something like:

```ts
interface DeleteTopicRecordsResult {
  topic: string,
  partitions: DeleteRecordsResult[]
}

interface DeleteRecordsResult {
  partition: string;
  offset: number;
  error: Error // This would be the original error from the broker
}

class KafkaJSDeleteTopicRecordsError extends KafkaJSError {
  constructor(result: DeleteTopicRecordsResult) {
    /*
     * This error is retriable if all the errors were retriable
     * Not sure if it's a good idea or not, but it seems right to me.
     */
    const retriable = result.partitions
      .filter(({ error }) => error != null)
      .every(({ error }) => e.retriable === true)

    super('Error while deleting records', { retriable })
    this.name = "KafkaJSDeleteRecordsError"
    this.partitions = result.partitions
  }
}
```

This would allow the user to know exactly which partitions failed to be deleted and why.

Do you have any thoughts?

artysidorenko:
Thanks for the feedback 😄  

> It's a bit annoying that if partition 3 is also out of range, I'll get another error and have to repeat the process, but again, not terrible

Ah yes I didn't factor that in; you're right.. if one partition is out of range it's possible/probable that another one might well be, making it a little annoying for the user to go through the same process again.

> A third option is to let the broker deal with it but aggregate the errors ourselves

I like the example you gave - makes sense. The only element that sits a little uncomfortably with me, conceptually, is the idea of a request that ultimately responds with an error, but nevertheless is allowed to partially modify the state of the data. But this is me coming from my risk-averse mindset where I try to make everything idempotent and wrap it all inside a postgres transaction 😂 . I guess in your example, partition 1 and 3 would have been modified anyway, if partition 2 was correct, so the request is unlikely to be corrupting the data... so I don't feel strongly about it.. happy to go with aggregating the errors (which will hopefully be more elegant than the early check).

I've got a couple of work deadlines this week but will have a go at this over the weekend, and respond to your other comments as well - I had a quick read but would like to set some proper time aside to respond fully and clean the PR up etc etc - if that's ok.

Nevon:
> The only element that sits a little uncomfortably with me, conceptually, is the idea of a request that ultimately responds with an error, but nevertheless is allowed to partially modify the state of the data.

Oh I agree completely. If there was a way we could guarantee atomicity here, I would be all for it, but since we are actually potentially making several calls to several brokers, there's no way to roll back the other operations if one of them fail. Even trying to check the offset early is just lulling you into a false sense of security, as some of the operations may still fail, either because the offsets have become out of range since you did the check or for some other reason, and then you're in the exact same situation where the operation may have partially succeeded.

> I've got a couple of work deadlines this week but will have a go at this over the weekend, and respond to your other comments as well - I had a quick read but would like to set some proper time aside to respond fully and clean the PR up etc etc - if that's ok.

Sure thing. There's no rush on my account, as long as we eventually get it merged.

docs/Admin.md
```diff
@@ -509,6 +509,31 @@ try {
 }
 ```
 
+## <a name="delete-topic-records"></a> Delete Topic Records
+
+Delete records for a selected topic. This will delete all records from the earliest up to the provided target offset for the given partition(s). To delete all records in a partition, use a target offset of `-1`.
```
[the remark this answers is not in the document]
artysidorenko: indeed, in the example here, it would delete 0-29, and then start consuming normally from #30. Just updated - hopefully clearer to the reader now.

src/admin/__tests__/deleteTopicRecords.spec.js
```diff
@@ -0,0 +1,269 @@
...
+
+  test('throws an error if the topic name is invalid', async () => {
+    await expect(admin.deleteTopicRecords({ topic: null })).rejects.toHaveProperty(
+      'message',
+      'Invalid topic null'
+    )
+
+    await expect(admin.deleteTopicRecords({ topic: ['topic-in-an-array'] })).rejects.toHaveProperty(
+      'message',
+      'Invalid topic topic-in-an-array'
```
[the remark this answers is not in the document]
artysidorenko: sure, updated in latest commit

src/admin/__tests__/deleteTopicRecords.spec.js
```diff
@@ -0,0 +1,269 @@
...
+    const messagesPartition0 = []
+    const messagesPartition1 = []
+    await admin.deleteTopicRecords({ topic: topicName, partitions: recordsToDelete })
+    consumer.run({
+      eachMessage: async event => {
+        if (event.partition === 0) messagesPartition0.push(event)
+        if (event.partition === 1) messagesPartition1.push(event)
+      },
+    })
+    // wait for the 2 partition batches to get processed
```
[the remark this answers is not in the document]
artysidorenko: cool, just updated to use `waitForMessages`. I had seen it, and initially I had leant towards waiting on the batch events as a way to simulate "wait for _every_ available message to get consumed" rather than waiting for a set number, because there's a bit of circularity in waiting for 13 messages to get consumed, then testing that each partition has the correct split of these 13 messages? (you will always have 13 because that's what you waited for)

With that in mind I've added an extra assertion to check that the first message consumed on partition 0 is offset 7, and the last message is offset 12, so that it's clear that the first 7 were skipped, hopefully that covers everything while not too clunky :)

src/admin/__tests__/deleteTopicRecords.spec.js
```diff
@@ -0,0 +1,269 @@
...
+        partitions: expect.arrayContaining([
+          { partition: 0, offset: '7' },
+          { partition: 1, offset: '0' },
+        ]),
+      },
+    ])
+  })
+
+  test('non-deleted messages are successfully consumed', async () => {
+    recordsToDelete = [{ partition: 0, offset: '7' }]
```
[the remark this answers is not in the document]
artysidorenko: yep this was a hangover from an earlier iteration where it wasn't initially being re-assigned. Updated it now in latest.

src/admin/index.js
artysidorenko: Tried to implement in the latest. Had to improvise a little bit because we're not waiting on all the requests to get resolved, so it will check if the "master" `KafkaJSDeleteTopicRecordsError` error is retriable + if at least 1 of the errors was related to stale metadata. Let me know if you had something else in mind (I also moved the `staleMetadata` helper function out into the errors file - but happy to move back if it doesn't belong there)

```javascript
} catch (e) {
  if (
    e.retriable &&
    e.brokers.some(
      ({ error }) => staleMetadata(error) || error.name === 'KafkaJSMetadataNotLoaded'
    )
  ) {
    await cluster.refreshMetadata()
  }
  throw e
}
```

src/admin/index.js
artysidorenko: tried to re-use the new `KafkaJSDeleteTopicRecordsError` structure, as a way to send back an error showing all cases of "leader not found" in the event it's more than 1. Let me know what you think (makes the code a bit less readable because it's collecting the errors in an array, but probably better for the user?)

src/admin/index.js
artysidorenko: (resolving as this comment has been covered by the updated KafkaJSDeleteTopicRecordsError pattern, I think?)

types/index.d.ts
```diff
@@ -938,12 +939,27 @@ export class KafkaJSDeleteGroupsError extends KafkaJSError {
...
 export interface KafkaJSDeleteGroupsErrorGroups {
   groupId: string
   errorCode: number
   error: KafkaJSError
 }
 
+
+export interface KafkaJSDeleteTopicRecordsErrorTopic {
+  topic: string,
+  brokers: KafkaJSDeleteTopicRecordsErrorBroker[]
```
artysidorenko, line 955: Small tweak on your initial suggestion as the requests and responses are at "broker-level". Might be a little less relevant to the end user as they're not exposed to the brokers directly? Though this makes it much more straightforward to map the error to the broker, rather than trying to figure out a way to map the error from the broker to the partition that caused the problem?

Let me know what you think..

Nevon: > Might be a little less relevant to the end user as they're not exposed to the brokers directly?

I agree that it's a bit of a pain in the ass to map, but I really think it's better for the user if we map it to partitions rather than brokers, as the user has sent in a list of partitions and now needs to know which partitions failed. Which broker the requests were made towards is pretty irrelevant.

I guess it's something like this:
```js
const deleteTopicRecordsResults = [
  { partitions: [{ partition: 0, offset: 10 }], error: new Error() },
  { partitions: [{ partition: 1, offset: 10 }, { partition: 2, offset: 10 }], error: new Error() },
]

const resultByPartition = deleteTopicRecordsResults.reduce((result, { partitions, error }) => {
  for (let partition of partitions) {
    result.partitions.push({ ...partition, error })
  }
  return result
}, { topic, partitions: [] })

const error = new KafkaJSDeleteTopicRecordsError({ result })
```

src/admin/index.js
Nevon: I agree 👍

src/protocol/requests/deleteRecords/v0/response.spec.js
```diff
@@ -0,0 +1,25 @@
+const { decode, parse } = require('./response')
+
+// FIXME - update with correct sample response
```
Nevon, line 3: Is this comment still relevant?

src/protocol/requests/deleteRecords/v0/response.js
```diff
@@ -0,0 +1,51 @@
...
+
+const parse = async data => {
+  const topicsWithErrors = data.topics
+    .map(({ partitions }) => ({
+      partitionsWithErrors: partitions.filter(({ errorCode }) => failure(errorCode)),
+    }))
+    .filter(({ partitionsWithErrors }) => partitionsWithErrors.length)
+
+  if (topicsWithErrors.length > 0) {
+    throw createErrorFromCode(topicsWithErrors[0].partitionsWithErrors[0].errorCode)
```
artysidorenko, line 42: Aha, I just realised.. the "mapping-errors-to-partitions" issue is not a kafka problem, it's just a result of how this response is being parsed. I had used other parts of the code as an example, but forgot about it until now.. it's calling `createErrorFromCode()` which picks only the first available error, but I could add a similar helper that works on multiple errors and includes some metadata when there's more than 1 partition?

Means another tweak, but I think it might be worthwhile as it would help clarify which partition is wrong? Also I guess the way it's written now there's a small chance that it could drop an error if there are multiple ones, especially if they're not the same error (though I guess this is a very unlikely/pessimistic scenario 😂)

src/protocol/requests/deleteRecords/v0/response.spec.js
artysidorenko: D'oh, apologies, indeed, I fixed the test but forgot to remove the fixme.

types/index.d.ts
artysidorenko: Yes, you're right, actually, I realised also if I change how the response is being parsed, I should be able to map the error to the partition precisely (explained in the other comment I just left, on `createErrorFromCode`)

artysidorenko:
Hopefully we're getting there.. it should be correctly matching up the errors to the partitions now, and will handle multiple errors from a single broker response.

There is a (small) bit of extra complexity in the response handler: am passing the _requested_ topic data into the response handler, so it can attach the requested offset to the error that's thrown back. Let me know if you prefer to leave it out.

To handle the possibility of multiple errors from a single broker, I'm also using the same `KafkaJSDeleteTopicRecordsError`  structure to throw these to the admin, but can create a separate one if it's preferable.

Nevon:
> (edit: forgot to re-request a review - not sure if you get pinged only when I push a PR update)

No, I don't get any notification unless you re-request a review. Even if I did, I wouldn't know if you were still working on it or if it's ready to be reviewed again, so requesting is greatly appreciated.

src/protocol/requests/deleteRecords/v0/response.js
```diff
@@ -0,0 +1,65 @@
...
+
+const parse = requestTopics => async data => {
+  const topicsWithErrors = data.topics
+    .map(({ partitions }) => ({
+      partitionsWithErrors: partitions.filter(({ errorCode }) => failure(errorCode)),
+    }))
+    .filter(({ partitionsWithErrors }) => partitionsWithErrors.length)
+
+  if (topicsWithErrors.length > 0) {
+    // at present we only ever request one topic at a time, so can destructure the arrays
```
Nevon, line 43: Can't wait for this to break when we introduce the ability to delete records from multiple topics 2 years from now... 😅

src/admin/index.js
```diff
@@ -990,6 +996,133 @@ module.exports = ({
...
+        if (
+          e.retriable &&
+          e.partitions.some(
+            ({ error }) => staleMetadata(error) || error.name === 'KafkaJSMetadataNotLoaded'
+          )
+        ) {
+          await cluster.refreshMetadata()
+        }
+        if (e.retriable) throw e
+        bail(e)
```
Nevon, line 1121: No need to handle this manually. The retrier already has logic inside for determining whether it should retry an error.

```suggestion
        throw e
```

Nevon (approved):
Just a minor suggestion as well as some merge conflict to handle, but then we should be good to go.

artysidorenko:
> Just a minor suggestion as well as some merge conflict to handle, but then we should be good to go.

Cool, thanks :)

Fortunately the conflict wasn't too bad - was just missing an extra error class.

PR #926, opened 2020-10-15
Prevent inflight's correlation id collisions

### LINKED ISSUE #793: Throw on duplicate correlation ids
While the correlation id used to match network requests with responses is always supposed to be unique at any given time, some tests were misbehaving because that wasn't the case.

Since this is one of those things that _should never happen_, we could add an invariant that throws a non-retriable error if this happens anyway, since something has gone terribly wrong and we won't get valid results afterwards.

_Originally posted by @ankon in https://github.com/tulios/kafkajs/pull/786#issuecomment-649094892_

ThomasFerro:
Resolves #793

This is my first PR here, please let me know if there is anything that does not fit with the project standards :)

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -178,6 +178,14 @@ class KafkaJSUnsupportedMagicByteInMessageSet extends KafkaJSNonRetriableError {
   }
 }
 
+class KafkaJSCorrelationIdAlreadyExists extends KafkaJSNonRetriableError {
+  constructor(correlationId) {
+    super(...arguments)
+    this.name = 'KafkaJSCorrelationIdAlreadyExists'
+    this.message = `Correlation id ${correlationId} already exists`
+  }
+}
+
 module.exports = {
   KafkaJSError,
   KafkaJSNonRetriableError,
@@ -201,4 +209,5 @@ module.exports = {
   KafkaJSLockTimeout,
   KafkaJSServerDoesNotSupportApiKey,
   KafkaJSUnsupportedMagicByteInMessageSet,
+  KafkaJSCorrelationIdAlreadyExists,
 }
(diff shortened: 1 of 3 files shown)


src/errors.js
```diff
@@ -178,6 +178,14 @@ class KafkaJSUnsupportedMagicByteInMessageSet extends KafkaJSNonRetriableError {
   }
 }
 
+class KafkaJSCorrelationIdAlreadyExists extends KafkaJSNonRetriableError {
+  constructor(correlationId) {
+    super(...arguments)
+    this.name = 'KafkaJSCorrelationIdAlreadyExists'
+    this.message = `Correlation id ${correlationId} already exists`
+  }
+}
```
Nevon, line 187: This error is a bit unusual in that it should _never_ happen. If it does happen, it likely means that there is a coding error somewhere. I would suggest we make this clear by creating a generic `KafkaJSInvariantViolation` and urge the user to file an issue:

```suggestion
const pkgJson = require('../package.json')
const issueUrl = pkjJson.bugs.url

class KafkaJSInvariantViolation extends KafkaJSNonRetriableError {
  constructor(e) {
    const message = e.message || e
    super(`Invariant violated: ${message}. This is likely a bug and should be reported.`)
    this.name = 'KafkaJSInvariantViolation'
    this.helpUrl = `${issueUrl}/new?assignees=&labels=bug&template=bug_report.md&title=${encodeURIComponent(`Invariant violation: ${message}`)}`
  }
}
```

```js
throw new KafkaJSInvariantViolation('Correlation id already exists')
```

ThomasFerro: Thanks for your feedback, I changed the error as requested.

Nevon:
Thanks for the contribution!

PR #929, opened 2020-10-19
Math.random with arguments?

pihentagy:
AFAIK Math.random doesn't accept arguments.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random


ankon (approved):

Nevon:
Thanks for noticing!

PR #933, opened 2020-10-29
Bump API versions for client-side throttling

ankon:
This is a fairly big, but rather mechanical piece of work: Review the existing requests and ensure that the bump for https://cwiki.apache.org/confluence/display/KAFKA/KIP-219+-+Improve+quota+communication is done (i.e. the throttleTime is reported as clientSideThrottleTime), and for all APIs used by KafkaJS that didn't support the client-side throttling yet add the needed API versions.

The first commit of the PR ensures that when adding the new APIs I didn't accidentally miss a bump somewhere, by moving the 'metadata about the API' test from the individual request.spec.js files into src/prototol/requests/index.spec.js, and making it generic over all requests defined.

### DIFF
--- a/src/protocol/requests/addOffsetsToTxn/v1/response.js
+++ b/src/protocol/requests/addOffsetsToTxn/v1/response.js
@@ -1,0 +1,24 @@
+const { parse, decode: decodeV0 } = require('../v0/response')
+
+/**
+ * Starting in version 1, on quota violation, brokers send out responses before throttling.
+ * @see https://cwiki.apache.org/confluence/display/KAFKA/KIP-219+-+Improve+quota+communication
+ *
+ * AddOffsetsToTxn Response (Version: 1) => throttle_time_ms error_code
+ *   throttle_time_ms => INT32
+ *   error_code => INT16
+ */
+const decode = async rawData => {
+  const decoded = await decodeV0(rawData)
+
+  return {
+    ...decoded,
+    throttleTime: 0,
+    clientSideThrottleTime: decoded.throttleTime,
+  }
+}
+
+module.exports = {
+  decode,
+  parse,
+}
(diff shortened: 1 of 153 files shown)


src/protocol/requests/addOffsetsToTxn/v1/response.js
```diff
@@ -0,0 +1,24 @@
...
+ * AddOffsetsToTxn Response (Version: 1) => throttle_time_ms error_code
+ *   throttle_time_ms => INT32
+ *   error_code => INT16
+ */
+const decode = async rawData => {
+  const decoded = await decodeV0(rawData)
+
+  return {
+    ...decoded,
+    throttleTime: 0,
```
Nevon, line 16: Why is it that you're overriding `throttleTime` here?

I can see the logic that you want to say that the request wasn't throttled, and that we are supposed to throttle on our side, but it's inconsistent with what we're doing everywhere else, where we return the original value for `throttleTime` and simply copy it to `clientSideThrottleTime`.

Nevon:
Thanks for taking on this unbelievably boring task. I had been meaning to do it for a while now, but I just never could bring myself to it. 😅

src/protocol/requests/addOffsetsToTxn/v1/response.js
Nevon: That said, I think I like your decision. I just wanna make sure we're consistent.

ankon: I think I actually changed all places to set throttleTime to 0 when returning a client-side value. :)

The thinking is that "throttleTime" greater than 0 means the broker throttled for that amount; while "clientSideThrottleTime" greater 0 indicates that the client did/should do so.

Issue #935, opened 2020-10-31
Typescript: SASLOptions interface does not have support for AWS specific fields


r115:
**Describe the bug**
When working with typescript, initializing the Kafka client for aws sasl mechanism throws a typing error.

**To Reproduce**
```
import { Kafka } from 'kafkajs';

const client = new Kafka({
   brokers: ['localhost:9092'],
   clientId: 'SampleClientId',
   ssl: true,
   sasl: {
     mechanism: 'aws',
     authorizationIdentity: 'awsidentity',
     accessKeyId: 'access_key',
     secretAccessKey: 'secret_key',
   }
})
```
REPL : https://repl.it/@r115/BrokenOptimalGoals

**Expected behavior**
Adding aws auth specific fields e.g `accessKeyId` should not throw a typing error.

**Observed behavior**
Adding aws mechanism auth fields e.g. `accessKeyId` throws an error.

```
 Type '{ mechanism: "aws"; authorizationIdentity: string; accessKeyId: string; secretAccessKey: string; }' is not assignable to type 'SASLOptions'.
  Object literal may only specify known properties, and 'authorizationIdentity' does not exist in type 'SASLOptions'.

```

**Environment:**
 - OS: `Mac OS 10.15.7`
 - KafkaJS version: `1.14.0`
 - Kafka version : `2.3.1`
 - NodeJS version: `12.16.3`

**Additional context**
N/A

PR #936, opened 2020-10-31
Update sasl options to support AWS specific fields

r115:
Adds optional fields to the `SASLOption` interface. #935

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -49,6 +49,10 @@ export interface SASLOptions {
   mechanism: SASLMechanism
   username?: string
   password?: string
+  authorizationIdentity?: string
+  accessKeyId?: string
+  secretAccessKey?: string
+  sessionToken?: string
   oauthBearerProvider?: () => Promise<OauthbearerProviderResponse>
 }
 


Nevon:
I think we can do better than just adding optional fields, since these fields are only relevant for certain authentication mechanisms. Using conditional types and discriminating unions we can make the fields dependent on the mechanism:

```typescript
type SASLMechanismOptionsMap = {
  'plain': { username: string, password: string },
  'scram-sha-256': { username: string, password: string },
  'scram-sha-512': { username: string, password: string },
  'aws': { authorizationIdentity: string, accessKeyId: string, secretAccessKey: string, sessionToken?: string },
  'oauthbearer': { oauthBearerProvider: () => Promise<OauthbearerProviderResponse> }
}

export type SASLMechanism = keyof SASLMechanismOptionsMap
type SASLMechanismOptions<T> = T extends SASLMechanism ? { mechanism: T } & SASLMechanismOptionsMap[T] : never
export type SASLOptions = SASLMechanismOptions<SASLMechanism>
```

It's a little verbose, but basically we map the mechanism to the desired options (`SASLMechanismOptionsMap`), then use that mapping to create a union type (`SASLOptions`) with all the possible keys of the map (`SASLMechanismOptions<T>`).

The result is that the fields are checked depending on the value of the `mechanism`, so you couldn't, for example, specify `authorizationIdentity` unless you are using the `aws` mechanism.

r115:
@Nevon gotcha. Let me refactor.

Issue #592, opened 2019-12-11
Async Events


sklose:
Hi,

I have a use case where I need to invalidate some state in my application when the consumer group rebalances which must complete before the rebalancing continues (in order to be compliant with the Kafka rebalance protocol and EOS semantics).

With the current API I can do

```
const { GROUP_JOIN } = consumer.events;
consumer.on(GROUP_JOIN, () => {
  // custom logic
});
```

The problem is that GROUP_JOIN is sync and my state invalidation code is inherently async.
I was thinking of submitting a PR that adds `onAsync` event callbacks for all the same events, but wanted to confirm with you beforehand if you have a better idea for this.

JaapRood:
We do exactly the same thing: setup work on every rebalance before we continue processing. The trick here is to use [`consumer.pause` and `consumer.resume`](https://kafka.js.org/docs/consuming#a-name-pause-resume-a-pause-resume), which can stop and start the flow of incoming messages. As soon as the group join is received, you pause incoming messages. Once you're ready to consume again, you resume again.

Aside from that, while async event handlers are an interesting thing, they also introduce issues. The problem in particular with them here is that the Kafka client is essentially a state machine, and to function correctly as a client to the Brokers, should never really be blocked from state transitions. Async event handling does just that, so keeping them synchronous and providing other methods to control the state of the client is the simpler and preferred option.

sklose:
@JaapRood, thanks for the quick response. Not sure if pause and resume will be sufficient for my use case. I am effectively managing my own queues of messages outside of the kafkajs runner loop and when a rebalance happens I need to wait for currently in-flight messages and clear pending items from the queue.

![image](https://user-images.githubusercontent.com/546088/70726324-ee4a3180-1ccb-11ea-947d-5a44d74d8e5d.png)
https://www.slideshare.net/ConfluentInc/everything-you-always-wanted-to-know-about-kafkas-rebalance-protocol-but-were-afraid-to-ask

The rebalance protocol in Kafka dictates that all consumers must stop processing messages and commit any outstanding offsets before crossing the synchronization barrier. It's mainly important for EOS, that's how double-processing of messages is avoided.

JaapRood:
Very interesting @sklose, thanks for shedding light on that. For my client work we're dealing with exactly the same scenario as you are, but don't support EOS yet for commits stored on the Broker. We definitely have plans for that, though. Can't say that I was aware that this synchronisation barrier played an essential role in it, so will have to dive into that!

[...]

PR #937, opened 2020-11-02
Emit GROUP_JOIN event on stale partition assignments

### LINKED ISSUE #923: Adding Partitions Didn't Trigger Group Join Event
**Describe the bug**
We increased the partitions on a topic to meet the demands of incoming traffic. The addition to partitions did not trigger a rebalance for our consumers or producers.

**To Reproduce**

If none of the above are possible to provide, please write down the exact steps to reproduce the behavior:
1. Run a producer that continuously produces messages to a topic
2. Run a consumer that subscribes to that topic and logs each topic offset
3. Send some messages
4. Increase partition count
5. GROUP_JOIN event did not trigger

**Expected behavior**
If the partition count changes, causing a rebalance, then I expect the GROUP_JOIN event would trigger the new partition map. 

**Observed behavior**
The consumer/producer continued as if no changes to partition count change.

**Environment:**
 - KafkaJS version [1.12.0]
 - Kafka version [2.3.1]
 - NodeJS version [12]

Nevon:
Doing this, it occurs to me that `GROUP_JOIN` is not exactly the right name for this event. It's more like `REBALANCED` or `GROUP_STABLE` or something, as it implies that the consumer has not only joined the group but also gone through the synchronization barrier.

Regardless, the join + sync is moved into the consumer group instead of living in the runner, and the `join` and `sync` methods become private so that we don't have the same issue again somewhere else.

Fixes #923

### DIFF
--- a/src/consumer/consumerGroup.js
+++ b/src/consumer/consumerGroup.js
@@ -309,11 +309,11 @@ module.exports = class ConsumerGroup {
         this.logger.info('Consumer has joined the group', payload)
       } catch (e) {
         if (isRebalancing(e)) {
-          // Rebalance in progress isn't a retriable error since the consumer
+          // Rebalance in progress isn't a retriable protocol error since the consumer
           // has to go through find coordinator and join again before it can
-          // actually retry. Throwing a retriable error to allow the retrier
-          // to keep going
-          throw new KafkaJSError('The group is rebalancing')
+          // actually retry the operation. We wrap the original error in a retriable error
+          // here instead in order to restart the join + sync sequence using the retrier.
+          throw new KafkaJSError(e)
         }
 
         bail(e)
--- a/src/consumer/consumerGroup.js
+++ b/src/consumer/consumerGroup.js
@@ -34,6 +34,11 @@ const STALE_METADATA_ERRORS = [
 const isRebalancing = e =>
   e.type === 'REBALANCE_IN_PROGRESS' || e.type === 'NOT_COORDINATOR_FOR_GROUP'
 
+const PRIVATE = {
+  JOIN: Symbol('private:ConsumerGroup:join'),
+  SYNC: Symbol('private:ConsumerGroup:sync'),
+}
+
 module.exports = class ConsumerGroup {
   constructor({
     retry,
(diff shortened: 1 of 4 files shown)


src/consumer/consumerGroup.js
```diff
@@ -272,6 +283,44 @@ module.exports = class ConsumerGroup {
...
+
+        this.instrumentationEmitter.emit(GROUP_JOIN, payload)
+        this.logger.info('Consumer has joined the group', payload)
+      } catch (e) {
+        if (isRebalancing(e)) {
+          // Rebalance in progress isn't a retriable error since the consumer
+          // has to go through find coordinator and join again before it can
+          // actually retry. Throwing a retriable error to allow the retrier
+          // to keep going
+          throw new KafkaJSError('The group is rebalancing')
```
ankon, line 316: I'm having trouble parsing the comment: "It isn't retriable [...] therefore throw a retriable error".

Nevon: Heh, yeah, I had the same issue. Took me a bit to figure out what it meant (I just moved it from the runner). Maybe we can come up with a better phrasing.

What it's saying is that the errors that we consider "rebalancing errors" are not retriable errors according to the protocol, but they are operations that we want to handle and retry - so we want the retrier to actually retry, not just bail out. Therefore we turn them into retriable errors even though they technically aren't.

ankon: > While "rebalance in progress" isn't a retriable error in the protocol [since the consumer ...], we can throw a retriable `KafkaJSError` here to restart that whole sequence using the retrier itself.

Something like that?

Nevon: a051f0f hopefully improved it a bit.

JaapRood: I understood that last one first try.

JaapRood (approved):
Haven't ran it, but looks good to me. 

Regarding the naming of `GROUP_JOIN`: you're right that this is more about the rebalance, but more problematic than that is it being an instrument event rather than a first class API on the consumer. If we want to enable exactly-once processing, it's required, as to be able to deal with the sync barrier correctly see [a comment on an related issue ](https://github.com/tulios/kafkajs/issues/592#issuecomment-565063376) and [slack conversation about it](https://kafkajs.slack.com/archives/CF6RFPF6K/p1585645227034200).

None of that should block merging this fix, but something to keep in mind when planning to make changes in that area 🙂.

PR #941, opened 2020-11-03
Fix acl type declarations to match implementation

CurtisEppel:
The [Javascript implementation][1] of createAcls doesn't match the Typescript type declarations

[1]: https://github.com/tulios/kafkajs/blob/f7a1664883/src/admin/index.js#L1131


Nevon (approved):

Nevon:
Thanks for the fix!

Issue #943, opened 2020-11-05
Deadlock on connection.onError handler


danielbatica:
**Describe the bug**
ETIMEDOUT error when consuming is not retried and thus, DISCONNECT consumer event is not fired.
In v1.12.0 a timeout error on the tls socket leads to retrier used and when retry is maxed out, the Consumer.onCrash is called. This behaviour no longer happens. 

It seems that connection.disconnect() awaits for inflight to get empty. This never happens since rejectRequests is called right after.

**To Reproduce**

If none of the above are possible to provide, please write down the exact steps to reproduce the behavior:
1. Run a consumer that subscribes to a topic and subscribe to the consumer DISCONNECT event. 
2. Wait for the consumer to connect.
3. Kill the connection to kafka (block port for instance).
4. The DISCONNECT event is not fired and the ETIMEDOUT error is not retried.

**Expected behavior**
Socket errors should lead to retry mechanism, which would let KafkaJSNumberOfRetriesExceeded to fire a consumer DISCONNECT event.

**Observed behavior**
This used to work on kafka 1.12.0

**Environment:**
 - OS: MacOs 10.15.4 
 - KafkaJS version 1.12.0
 - Kafka version:  2.5.0 , used confluent cloud basic cluster with ssl and local docker confluentinc/cp-kafka 
 - NodeJS version tested with 13.14.0 and 14.15.0

**Additional context**

danielbatica:
After digging a bit it seems that there is a deadlock when calling the connection disconnect in the onError handler of the socket object. https://github.com/tulios/kafkajs/blob/3bd245d31f10acf83c91cba00e8cbb983947add5/src/network/connection.js#L212

danielbatica:
Proposed fix: https://github.com/tulios/kafkajs/pull/944

PR #944, opened 2020-11-06
fix: deadlock on connection onError handler

danielbatica:
Issue described here: https://github.com/tulios/kafkajs/issues/943

### DIFF
--- a/src/network/connection.js
+++ b/src/network/connection.js
@@ -155,8 +155,8 @@ module.exports = class Connection {
         })
 
         this.logError(error.message, { stack: e.stack })
-        await this.disconnect()
         this.rejectRequests(error)
+        await this.disconnect()
 
         reject(error)
       }


tulios (approved):

danielbatica:
@tulios Do you know when is the next release due? Thank you.

tulios:
The fix was released on pre-release `1.15.0-beta.20`, so you can already use it. We don't have a plan for 1.15 yet, but I think we should ship what's in master soon

danielbatica:
Thank you!

src/network/connection.js
```diff
@@ -155,8 +155,8 @@ module.exports = class Connection {
         })
 
         this.logError(error.message, { stack: e.stack })
-        await this.disconnect()
         this.rejectRequests(error)
+        await this.disconnect()
```
smartinio, line 159: This bug seems to happen right below this, inside `onTimeout` as well. I supposed that poses the same deadlock issue? @tulios @danielbatica

smartinio: Added #956

PR #951, opened 2020-11-11
Document the possible values for the state of a consumer group

ankon:
It took me quite a while to understand that non-existing consumer groups aren't reported as non-existing (no value), but rather come as a group with the literal value 'Dead'. The Kafka protocol documentation doesn't say that clearly either, but luckily I found a unit test inside KafkaJS that explicitly described the case.

This PR then adds the actual possible values for the consumer group `state` as a type, so that anyone else trying to work with them gets a headstart :)

Note that I suspect Kafka could introduce new states, and these new states would appear then as values as well. From a typing view that "sucks", as the compiler wouldn't be able to notice the problem. 

But, any code written would need to learn about any new state as well, and could temporarily cast the `state` property back to `string` until KafkaJS would follow such a hypothetical change.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -657,12 +657,15 @@ export type MemberDescription = {
   memberMetadata: Buffer
 }
 
+// See https://github.com/apache/kafka/blob/2.4.0/clients/src/main/java/org/apache/kafka/common/ConsumerGroupState.java#L25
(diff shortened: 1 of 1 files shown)


Nevon (approved):

Nevon:
I agree that having to update the type definition when kafka adds new states sucks, but I definitely see the value in having this as a string literal type just for the purposes of documentation. I had to do the same digging when writing some healthchecks that used the consumer group state, so it'd be good for people to not have to do that anymore.

Issue #950, opened 2020-11-11
Metadata is not refreshed on connection errors


smartinio:
**Describe the bug**
⚠️  **Disclaimer:** _This is primarily a hypothesis to explain an issue I have in production (weekly maintenance)._

When a broker is killed remotely, despite the consequent connection errors, kafkajs doesn't realize that the node is currently inactive, and will keep trying to send messages until it inevitably fails with a KafkaJSNonRetriableError.

**To Reproduce**
I wrote a [test](https://github.com/smartinio/kafkajs/blob/refresh-metadata-on-error/src/producer/sendMessages.spec.js#L208) and [fix](https://github.com/smartinio/kafkajs/blob/refresh-metadata-on-error/src/producer/sendMessages.js#L128) for this in a fork (commenting out the fix fails the test).

**Expected behavior**
I expect new metadata to be fetched if there's a connection error when attempting to send messages to the broker.

**Observed behavior**
Metadata does not seem to be refreshed, since kafkajs retries the same exact broker until it fails.

Logs from my service:
```
{"level":"ERROR","timestamp":"2020-11-10T11:09:29.843Z","logger":"kafkajs","message":"[Connection] Connection timeout","broker":"k-12.production.eu1.[masked]:9094","clientId":"[masked]"}

{"level":"ERROR","timestamp":"2020-11-10T11:09:29.844Z","logger":"kafkajs","message":"[BrokerPool] Failed to connect to broker, reconnecting","retryCount":0,"retryTime":301}

{"level":"ERROR","timestamp":"2020-11-10T11:09:35.145Z","logger":"kafkajs","message":"[Connection] Connection timeout","broker":"k-12.production.eu1.[masked]:9094","clientId":"[masked]"}

{"level":"ERROR","timestamp":"2020-11-10T11:09:35.146Z","logger":"kafkajs","message":"[BrokerPool] Failed to connect to broker, reconnecting","retryCount":1,"retryTime":660}

{"level":"ERROR","timestamp":"2020-11-10T11:09:40.806Z","logger":"kafkajs","message":"[Connection] Connection timeout","broker":"k-12.production.eu1.[masked]:9094","clientId":"[masked]"}

{"level":"ERROR","timestamp":"2020-11-10T11:09:40.807Z","logger":"kafkajs","message":"[BrokerPool] Failed to connect to broker, reconnecting","retryCount":2,"retryTime":1396}

[...]

PR #953, opened 2020-11-11
Refresh metadata on connection error [#950]

smartinio:
As I'm relatively new to the KafkaJS codebase, I might be missing some context as to why the code I'm submitting isn't already the case.

See #950

### DIFF
--- a/src/producer/sendMessages.js
+++ b/src/producer/sendMessages.js
@@ -122,7 +122,11 @@ module.exports = ({ logger, cluster, partitioner, eosManager }) => {
         const responses = Array.from(responsePerBroker.values())
         return flatten(responses)
       } catch (e) {
-        if (staleMetadata(e) || e.name === 'KafkaJSMetadataNotLoaded') {
+        if (
+          staleMetadata(e) ||
+          e.name === 'KafkaJSMetadataNotLoaded' ||
+          e.name === 'KafkaJSConnectionError'
+        ) {
           await cluster.refreshMetadata()
         }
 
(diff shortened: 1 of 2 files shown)


tulios:
@smartinio this is covered in a layer above:
https://github.com/tulios/kafkajs/blob/daeec895c09828622a550134b7cf9b1c0ceebded/src/producer/messageProducer.js#L117-L127

but you haven't seen the message `Failed to send messages: ...`, which means the client is first doing network retries. I think it will eventually retry the messages once the error bubbles up.

tulios:
The client should also distinguish between a regular network error and a broker going offline, so the messages you are seeing are due to network issues while the broker is going down but is not yet fully down.

https://github.com/tulios/kafkajs/blob/daeec895c09828622a550134b7cf9b1c0ceebded/src/network/connection.js#L137-L143

Once the broker is offline, the client will propagate the error and fully refresh the metadata.

In the issue you mentioned that you are using version `1.12.0`, I recommend you to use the latest pre-release or at least the latest version first.

smartinio:
> but you haven't seen the message `Failed to send messages: ...`, which means the client is first doing network retries. I think it will eventually retry the messages once the error bubbles up.

@tulios Yeah I don't have "Failed to send messages" anywhere in my service logs. Since I'm seeing "Connection timeout", could it be related to this deadlock issue that was recently fixed, but only fixed in the `onError` handler?
> https://github.com/tulios/kafkajs/blob/3bd245d31f10acf83c91cba00e8cbb983947add5/src/network/connection.js#L164-L173

smartinio:
Closing this in favor of #956.

PR #959, opened 2020-11-15
fixed: ISocketFactory type's argument

### LINKED ISSUE #908: ISocketFactory seems to be a wrong type declaration
Right now the ISocketFactory type is mismatch with the real implementation

It is currently like this
```ts
export type ISocketFactory = (
  host: string,
  port: number,
  ssl: tls.ConnectionOptions,
  onConnect: () => void
) => net.Socket
```

while the real implementation is sending only 1 argument to it like this
https://kafka.js.org/docs/configuration#custom-socket-factory
```js
const myCustomSocketFactory = ({ host, port, ssl, onConnect }) => {
  const socket = ssl
    ? tls.connect(
        Object.assign({ host, port }, ssl),
        onConnect
      )
    : net.connect(
        { host, port },
        onConnect
      )

  socket.setKeepAlive(true, 30000)

  return socket
}
```

**Expected behavior**
Could be something similar to the below example
```ts
type ISocketFactoryArguments = {
  host: string,
  port: number,
  ssl: tls.ConnectionOptions,
  onConnect: () => void,
}

export type ISocketFactory = (args: ISocketFactoryArguments) => net.Socket
```

**Environment:**
 - KafkaJS version 1.14.0
 - NodeJS version v12.14.1
 - TypeScript version 3.7.5

Howie-IO:
Fixed #908 ISocketFactory seems to be a wrong type declaration. Created new type `ISocketFactoryArgs`

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -32,12 +32,14 @@ export interface KafkaConfig {
   logCreator?: logCreator
 }
 
-export type ISocketFactory = (
+export interface ISocketFactoryArgs {
   host: string,
   port: number,
   ssl: tls.ConnectionOptions,
   onConnect: () => void
-) => net.Socket
+}
+
+export type ISocketFactory = (args: ISocketFactoryArgs) => net.Socket
 
 export type SASLMechanism = 'plain' | 'scram-sha-256' | 'scram-sha-512' | 'aws' | 'oauthbearer'
 


Nevon (approved):

Nevon:
Thanks for the contribution!

PR #976, opened 2020-12-01
Using one container proposal

Ivasan7:


tulios:
@Ivasan7 `spotify/kafka` is not up to date and an ancient container. I wouldn't recommend its use. The reason we use separate containers is that it makes it simpler for us to update Kafka versions

Ivasan7:
@tulios good to know, then I switch to this one

Issue #961, opened 2020-11-16
Allow sharing of `Cluster`/`BrokerPool` instances with multiple consumers


ankon:
**Is your feature request related to a problem? Please describe.**

KafkaJS creates a lot of connections when one uses a lot of consumers[1], which may explain throttling issues seen with Confluent Cloud (#931) and also issues I have seen and so far ignored with smaller AWS MSK clusters.

Looking through the Kafka protocol and the code it should be possible to multiplex multiple consumers onto the same connection, at the (likely) cost of latency.

[1] or producers, or admin clients -- but this issue is specifically about the consumers :)

**Describe the solution you'd like**

For advanced users I would like to be able to create a single `Cluster`, and then hand it to the `consumer` method. The `consumer` method would check whether the cluster is "compatible" with the settings otherwise given to it that affect the connection (such as max requests in flight), and if these settings are compatible would use this cluster rather than create a new one when needed.

Methods for connecting/disconnecting may potentially be disabled on the returned consumer, in favor of having to control that explicitly on the shared cluster.

ankon:
I'm probably going to try this out over the week, but instead of doing and then filing a PR I want to see what you guys think about this -- would it be ok? Is this really "advanced", or possibly the norm?

Nevon:
Intuitively I like this idea, since as you say, there's no real reason that multiple different clients couldn't share connections. One issue I see immediately though is that currently cluster metadata lives on the cluster and is determined by `targetTopics`. This would be different for different consumers. It might be that what you actually want to share is the BrokerPool, which is where the connections live. If that's the layer that's being shared, you could share it between multiple consumers as well as producers or admin clients.

What I don't necessarily like about it is that:

1. It may complicate the code. We don't know that without trying, so I'm not discounting the idea just based on this, but the benefit needs to outweigh the complexity. As long as the complexity doesn't find its way deep inside several layers of the code, it's probably fine.
2. I would be a little worried that people will use this (complicating their code a lot, since you now have a shared lifecycle between multiple consumers) without it actually benefitting them. If your client is talking to 15 brokers, I don't think it will make any difference whether you have 30 connections or 15 from the client's perspective - you still only have <15 cores to pull data from the socket. That said, it may make a difference from the broker perspective. Do you know if any other clients do something similar?

ankon:
> It might be that what you actually want to share is the BrokerPool

I'm also suspecting more that, than the complete `Cluster` instance.

[...]

PR #977, opened 2020-12-01
Feature: Optionally allow sharing of a broker pool across consumers/producers

ankon:
See #961 for a discussion.

I left cleanup/improvement commits in this, tell me if you rather have some of them split apart into separate PRs.

### DIFF
--- a/src/cluster/brokerPool.js
+++ b/src/cluster/brokerPool.js
@@ -95,7 +95,7 @@ module.exports = class BrokerPool {
   /**
    * @public
    * @param {number} [clusterHandle]
-   * @returns {Promise<null>}
+   * @returns {Promise<void>}
    */
   async connect(clusterHandle) {
     if (clusterHandle) {
--- a/src/cluster/brokerPool.js
+++ b/src/cluster/brokerPool.js
@@ -12,12 +12,14 @@ const hasBrokerBeenReplaced = (broker, { host, port, rack }) =>
 
 module.exports = class BrokerPool {
   /**
-   * @param {ConnectionBuilder} connectionBuilder
-   * @param {Logger} logger
-   * @param {Object} retry
-   * @param {number} authenticationTimeout
-   * @param {number} reauthenticationThreshold
-   * @param {number} metadataMaxAge
+   * @param {object} options
+   * @param {import("./connectionBuilder").ConnectionBuilder} options.connectionBuilder
+   * @param {import("../../types").Logger} options.logger
+   * @param {import("../../types").RetryOptions} [options.retry]
+   * @param {boolean} [options.allowAutoTopicCreation]
+   * @param {number} [options.authenticationTimeout]
+   * @param {number} [options.reauthenticationThreshold]
+   * @param {number} [options.metadataMaxAge]
    */
   constructor({
     connectionBuilder,
@@ -43,6 +45,9 @@ module.exports = class BrokerPool {
       })
 
     this.brokers = {}
+    /** @type {Broker | undefined} */
+    this.seedBroker = undefined
+    /** @type {import("../../types").BrokerMetadata | null} */
     this.metadata = null
     this.metadataExpireAt = null
     this.versions = null
@@ -123,8 +128,9 @@ module.exports = class BrokerPool {
 
   /**
    * @public
-   * @param {String} host
-   * @param {Number} port
+   * @param {Object} destination
+   * @param {string} destination.host
+   * @param {number} destination.port
    */
   removeBroker({ host, port }) {
     const removedBroker = values(this.brokers).find(
@@ -214,7 +220,7 @@ module.exports = class BrokerPool {
   }
 
   /**
-   * Only refreshes metadata if the data is stale according to the `metadataMaxAge` param
+   * Only refreshes metadata if the data is stale according to the `metadataMaxAge` param or does not contain information about the provided topics
    *
    * @public
    * @param {Array<String>} topics
@@ -236,7 +242,8 @@ module.exports = class BrokerPool {
 
   /**
    * @public
-   * @param {string} nodeId
+   * @param {object} options
+   * @param {string} options.nodeId
    * @returns {Promise<Broker>}
    */
   async findBroker({ nodeId }) {
@@ -252,8 +259,9 @@ module.exports = class BrokerPool {
 
   /**
    * @public
-   * @param {Promise<{ nodeId<String>, broker<Broker> }>} callback
-   * @returns {Promise<null>}
+   * @param {(params: { nodeId: string, broker: Broker }) => Promise<T>} callback
+   * @returns {Promise<T>}
+   * @template T
    */
   async withBroker(callback) {
     const brokers = shuffle(keys(this.brokers))
--- a/src/cluster/brokerPool.js
+++ b/src/cluster/brokerPool.js
@@ -52,6 +52,15 @@ module.exports = class BrokerPool {
     this.metadataExpireAt = null
     this.versions = null
     this.supportAuthenticationProtocol = null
+
+    /** @type {{[topic: string]: Promise<null>}} */
+    this.pendingMetadataTopicPromises = {}
+    /** @type {{[topic: string]: () => void}} */
+    this.pendingMetadataTopicResolves = {}
+    /** @type {{[topic: string]: (err: Error) => void}} */
+    this.pendingMetadataTopicRejects = {}
+    /** @type {Promise<void>} */
+    this.nextSuccessfulRefreshMetadata = undefined
   }
 
   /**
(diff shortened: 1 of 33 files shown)


docs/SharedBrokerPool.md
```diff
@@ -0,0 +1,43 @@
...
+
+```javascript
+const consumer = kafka.consumer({ groupId: 'my-group', brokerPool })
+```
+
+```javascript
+const producer = kafka.producer({ brokerPool })
+```
+
+The consumer and producer can then be used as described in [Consuming Messages](Consuming.md) and [Producing Messages](Producing.md), with the caveat that the `connect` and `disconnect` methods will affect the complete pool, and therefore should typicaly not be used.
```
Nevon, line 27:
```suggestion
The consumer and producer can then be used as described in [Consuming Messages](Consuming.md) and [Producing Messages](Producing.md), with the caveat that the `connect` and `disconnect` methods will affect the complete pool, and therefore should typically not be used.
```

docs/SharedBrokerPool.md
```diff
@@ -0,0 +1,43 @@
...
+It is possible to influence this behavior for consumers and producers by manually configuring "broker pools", and sharing these.
+
+## <a name="consumer"></a> Example
+
+```javascript
+const brokerPool = kafka.brokerPool({ })
+await brokerPool.connect()
+```
+
+This broker pool can be provided to the `kafka.consumer` and `kafka.producer` methods, instead of the options that would typically configure the pool.
```
Nevon, line 17: Any reason not to do the same for the admin client? Granted, it is typically not used for long running processes where this would really matter, but it could be, and as long as it doesn't make the admin client more complicated, I don't see any reason not to allow it.

docs/SharedBrokerPool.md
```diff
@@ -0,0 +1,43 @@
+---
+id: shared-broker-pool
```
Nevon, line 2: What do you think about re-using the term "connection pool" for this, rather than the term "broker pool"?

It's not exactly the same thing, but it's very close, and it might make it easier for people to understand what this does and why they might want to do it. I only want to invent new terms if using existing ones misleads and causes more problems than they solve.

docs/SharedBrokerPool.md
Nevon, line 27: How will graceful shutdowns work with this? Let's say I have a typical stream processor, consuming from one topic and publishing into another.

In a normal KafkaJS application, I'd want to first shut down the consumer and then the producer, so that while my consumer is in the process of shutting down the producer is still up to be able to be used while wrapping up. How would that work with this solution?

Also, is there a way we can prevent people from shooting themselves in the foot here, since now calling `consumer.disconnect()` could have unintended consequences if you're not paying attention. I'm not sure what the best option is, to be honest.

src/cluster/brokerPool.js
```diff
@@ -143,16 +145,32 @@ module.exports = class BrokerPool {
...
    * @returns {Promise<null>}
    */
   async refreshMetadata(topics) {
+    const getTargetTopics = topicMetadata => {
+      if (!topicMetadata) {
+        return topics
+      }
+      const targetTopics = topicMetadata.map(({ topic }) => topic)
+      return topics.reduce(
+        (result, topic) => (result.includes(topic) ? result : result.concat(topic)),
```
Nevon, line 158: Shouldn't this be `result.push(topic)` since topic is a string?

src/index.js
```diff
@@ -40,33 +57,65 @@ module.exports = class Client {
...
+        metadataMaxAge,
+      })
+    }
+    this[PRIVATE.CREATE_CLUSTER] = ({
+      isolationLevel,
+      brokerPool,
+      instrumentationEmitter,
+      ...brokerPoolOptions
+    }) => {
+      if (
```
Nevon, line 96: This part could be a lot clearer. I guess this is trying to determine if we're trying to create a cluster with options that are not compatible with the existing broker pool, but I can't really tell what's going on.

src/index.js
```diff
@@ -40,33 +57,65 @@ module.exports = class Client {
...
+        Object.entries(brokerPoolOptions).filter(([, value]) => typeof value !== 'undefined')
+          .length > 0 &&
+        brokerPool
+      ) {
+        // XXX: We could compare against the actual options of the provided pool ...
+        throw new Error('Incompatible options: brokerPool and broker pool creation options')
```
Nevon, line 102: This should be a `KafkaJSNonRetriableError` and probably include some more information to understand how to resolve it.

src/index.js
```diff
@@ -40,33 +57,65 @@ module.exports = class Client {
...
+      } else if (!brokerPool) {
+        brokerPool = this[PRIVATE.CREATE_BROKERPOOL]({
+          ...brokerPoolOptions,
+          instrumentationEmitter,
+        })
+      } else if (instrumentationEmitter) {
+        brokerPool.forwardInstrumentationEvents(instrumentationEmitter)
```
Nevon, line 109: What's up with this forwarding of the instrumentation events?

Nevon: I haven't fully understood the idea here, but just one thing that we need to be mindful of is that we aren't duplicating events now. For example, let's say we are sending metrics on network requests, and we have both a consumer and a producer instrumented that share a brokerpool, would they both receive an event (and would they receive "each other's" events?

docs/SharedBrokerPool.md
ankon: Allowing is fine, the problem was mainly that the `admin()` function would always force the `allowTopicCreation` setting, and I didn't want to spend too much time thinking about the implications.
But yes, I think we could simply allow that as well, and make the caller responsible for knowing what they do -- which I think is fair for an "advanced" use case.

docs/SharedBrokerPool.md
ankon: Fair point. 

But I suspect then that it would make sense to completely rename things also in the code to stay consistent?

docs/SharedBrokerPool.md
ankon: > Also, is there a way we can prevent people from shooting themselves in the foot here, since now calling consumer.disconnect() could have unintended consequences if you're not paying attention. I'm not sure what the best option is, to be honest.

I toyed with this a bit a few minutes ago, exactly because I shot myself in the foot when the crash handling disconnected my cluster :)

I eventually decided that the crash handler definitely was wrong in attempting it, and pushed the problem away for now. I find it difficult to have names like "connect" and "disconnect" on the consumer/producers, because there's not much there except doing exactly that. Having connect/disconnect not do connect/disconnect is just simply confusing. But, at the same time disconnecting a shared pool a real issue. Still thinking as well, but I could see an approach that would return a different class depending on whether you have or have not provided a shared pool, which should be possible to model with types so that the compiler can scream at you.

src/index.js
ankon: I'll try to pull this one out, indeed it's quite a mess to follow.

src/index.js
ankon: The problem that this is supposed to solve is that consumers/producers emit instrumentation events from the underlying connections, with slightly adjusted names. I want to stay compatible with the existing API whereever possible, so that the sharing really can be completely ignored unless you need to look into it.

> For example, let's say we are sending metrics on network requests, and we have both a consumer and a producer instrumented that share a brokerpool, would they both receive an event (and would they receive "each other's" events?

Good one! Right now they would receive the events for their own requests and also other requests ... definitely needs to be touched further to prevent that, probably by adding some reasonably smart filter somewhere.

ankon:
For clarity: I set this to "draft", and will add commits that clean up/improve things. I'll rebase before moving away from draft, and then shuffle 'squash:'/'fixup:' commits together. This hopefully makes reviewing a bit easier by keeping smaller meaningful commits for longer.

Right now we're busy also testing the impact of these changes in our test environment, so that we can be reasonably sure that things work out.

src/index.js
ankon: https://github.com/tulios/kafkajs/pull/977/commits/71883cbe5638b86ecf859238df30ed9e0e517dbe

src/cluster/brokerPool.js
ankon: Well, no. The result of `result.push(topic)` would be the new length of `result`, but we need the actual result here. But, given that I quite often mix up `push` and `concat`, and this code shouldn't be particularly performance critical either, I rewrote it slightly in https://github.com/tulios/kafkajs/pull/977/commits/6aee9ac9be509690019e66729732473495bf2e54 to be more "obvious":

```js
return topics.reduce(
  (result, topic) => (result.includes(topic) ? result : [...result, topic]),
  targetTopics
)
```

docs/SharedBrokerPool.md
Nevon: > But I suspect then that it would make sense to completely rename things also in the code to stay consistent?

I instinctually think you are correct, as it rarely makes sense to call things one thing externally and another thing internally, but in this case I'm not sure. The reason I think BrokerPool makes sense internally is because it is not just holding connections, it is also responsible for holding metadata and mapping between node ids and connections (through the broker abstraction). So internally, it makes sense to call it a BrokerPool.

We initially called it that because we wanted to have a method that allowed us to execute something on any connected broker (`.withBroker((fn: { nodeId: string, broker: Broker }) => Promise<void>)`), which made it sound like it was a pool of brokers that we could borrow from.

Externally, however, the point is to communicate that this is a way to share connections between different instances, which is why Connection Pool is a good name. No one that hasn't dug into the innards of KafkaJS has ever heard of a BrokerPool, and I don't think you intuitively understand that it's a way to share connections unless you have experience developing KafkaJS itself.

I'm really not 100% sure. I am quite sure that I like the name "Connection Pool" externally - certainly more than I like exposing the name "Broker Pool". But I'm not sure whether or not I want to rename everything internally.

docs/SharedBrokerPool.md
Nevon: > Still thinking as well, but I could see an approach that would return a different class depending on whether you have or have not provided a shared pool, which should be possible to model with types so that the compiler can scream at you.

This is what I had in mind as well, for exactly the same reasons you described.

src/cluster/brokerPool.js
Nevon: [From the MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat):

> `const new_array = old_array.concat([value1[, value2[, ...[, valueN]]]])`
> 
> **ValueN**
> Arrays and/or values to concatenate into a new array [...]

I had no idea Array.concat could push elements onto the array if the value is not itself an array.

docs/SharedBrokerPool.md
ankon: I tried just touching the external API to use "connection pool", I think I agree: This reads easier and makes intuitively sense. Keeping the existing internal BrokerPool is fine, and in fact makes it more obvious that one cannot just create one of these things manually -- one must go through the exposed API on the client.

See https://github.com/tulios/kafkajs/pull/977/commits/a02f4557e8b12b896de28a9e9a79ff9422f17838

src/cluster/index.js
```diff
@@ -218,6 +218,11 @@ module.exports = class Cluster {
     }
 
     const topicMetadata = metadata.topicMetadata.find(t => t.topic === topic)
+    if (!topicMetadata) {
+      throw new KafkaJSTopicMetadataNotLoaded('Topic metadata does not include required topic', {
```
ankon, line 222: A note here: This is an API change, in the sense that a partition assigner would typically call this function. 

It is possible that members of a group see different topics, and the assigner cannot just assume that it can look up the cached metadata for all topics known to all members in the group.
Still I think this is fair, and the assigner code should actually explicitly handle this error. In our case for example I know that the issue is transient and will go away with one of the next runs, so we simply ignore it as before.

docs/SharedBrokerPool.md
Nevon: I asked for some feedback from internal customers of KafkaJS on this API, to see if they had some thoughts, and one person had a very interesting idea that I think makes a lot of sense, even if it makes life a bit more annoying for us.

What if we don't expose `connect` and `disconnect` on the connection pool at all, but rather have the connection pool connect if it's not connected, and disconnect when there are no more clients that require it to be connected?

For example:

```js
const connectionPool = kafka.connectionPool()
const consumer = kafka.consumer({ connectionPool })
const producer = kafka.producer({ connectionPool })

// ConnectionPool is not connected, so it connects
await producer.connect()

// Consumer registers that it also wants to be connected
await consumer.connect()

// Consumer doesn't want to be connected anymore.
// It stops etc, but the connection pool is still connected.
await consumer.disconnect()

// No more connections are required, connection pool disconnects.
await producer.disconnect()
```

That way the difference in the application code is almost 0 whether you have connection pooling or not. As a user, the way you describe the dependency between instances of producers and consumers is explicit and clear. The only downside I see is that the connection pool has to do a bit of work to keep track of when it should disconnect and not.

ankon: I think this is a good idea.

Essentially the pool knows about it's consumers at this point, which would produce a cycle -- but maybe a simple counter of how many consumers exist might be enough.

src/cluster/index.js
ankon: In testing it showed that while this might be "theoretically nice", the change leads to a lot of places that now need to handle the error, but sometimes the error isn't actually an error.

Dropped this one again for that reason.

src/index.js
ankon: I looked a lot here whether it would be somehow possible to know which event really is "caused by", and therefore should get reported to a particular consumer/producer.

I couldn't find a good way to do it in general, but looking at the problem a bit:
1. Most "interesting" events are coming from the cluster and the consumer/producer itself, and these aren't shared.
2. The only events not coming from there are the ones from the request queue, and at that point we have really no link to the originating consumer/producer. But also: Events such as "got a response to the metadata request" would be interesting to all consumers/producers using the same pool, because they share that metadata. Trying to pull this apart to find only those then that have interest in the specific topic (or request) seems like a huge undertaking.

I think that when you share a connection pool you should be somewhat aware of the fact that network traffic goes through that pool coming from all clients that you share the pool with, WDYT?

src/index.js
ankon: Moved it into a separate function, and tried to shorten the code a bit more to be easier to follow.

docs/SharedBrokerPool.md
ankon: I think https://github.com/tulios/kafkajs/pull/977/commits/ccfaa69cef0f8741cf5af122ab9c268402c495e8 actually does that quite nicely (to be squashed into the main commit eventually, but now pulled out to show it)

docs/SharedBrokerPool.md
ankon: I think the way forward here is probably to not just check whether you provide any settings at all, but rather expose them on the connection pool entity so that the `admin`/`consumer`/`producer` functions can easily check for "compatibility".

I'll hopefully get to do that tomorrow as well, which might help further with making the compatibility check more obvious.

ankon:
Small update here after a lot of internal testing and quite some puzzling: Functionality-wise I'm pretty happy with this, and the only thing missing right now there is to make the compatibility check work so that `admin` can be configured with a connection pool as well.

I guess documentation could be a bit fleshed out, at least I do want to add a short section on "measure your performance, latency will potentially go up depending on your traffic pattern". Incidentally that's a bit blocked on me working out how to efficiently measure that in our environment :) 

Lastly, something I wanted to point out: https://github.com/tulios/kafkajs/pull/977/commits/6cb83d686bc940ce7d0601c843f57d5524aa13b0 adds _a lot_ of fixes for JSDoc and the typescript `types/index.d.ts`. There's still a lot left over, but this alone helps quite a bit in the touched areas to make intellisense work nice ... and I'm trying to carefully watch out if there's any actual bugs getting exposed.

ankon:
> Incidentally [measuring latency is] a bit blocked on me working out how to efficiently measure that in our environment :)

Went with the most obvious approach: Actually introduce a `ConnectionPool` class that extends the broker pool with instrumentation events. This class is only returned when using the `connectionPool` function to create it.

src/index.js
Nevon: My only worry is that we would ruin metrics collection by emitting the same event twice. For example, let's say people are collecting metrics on request queue depth or request duration. If those events are dispatched to multiple instances, that could skew the metrics.

I might be wrong about this, as I don't know the details of what events are shared and which aren't, but it feels like something that's very easy to get wrong for our users. That said, I don't have a solution off the top of my head either. If the current structure of the event emitters makes it impossible to know which instance to propagate the events to, maybe we need to restructure it somehow.

I hope that I can properly review this PR this weekend, and maybe have some ideas. 🤞

src/cluster/index.js
```diff
@@ -20,131 +18,94 @@ const mergeTopics = (obj, { topic, partitions }) => ({
...
-      reauthenticationThreshold,
-      metadataMaxAge,
-    })
+    this.brokerPool = brokerPool
+    this.clusterHandle = nextClusterHandle++
     this.committedOffsetsByGroup = offsets
   }
 
   isConnected() {
+    // XXX: Would this also mean "... and consider me one of the stakeholders?"
```
Nevon, line 60: This is a tricky one. I would look into where this is called from, and if we can maybe get rid of it in favor of an idempotent connect.

src/cluster/index.js
```diff
@@ -434,7 +398,7 @@ module.exports = class Cluster {
 
       topicConfigurations[topic] = { timestamp }
 
-      keys(partitionsPerLeader).map(nodeId => {
+      keys(partitionsPerLeader).forEach(nodeId => {
```
Nevon, line 401: This is my pet-peeve, so thank you 😄  I'm not gonna `git blame` this line, but just assume that Tulio wrote it and not me. 😉

src/index.js
```diff
@@ -40,33 +61,90 @@ module.exports = class Client {
...
         sasl,
         clientId,
         connectionTimeout,
-        authenticationTimeout,
-        reauthenticationThreshold,
         requestTimeout,
         enforceRequestTimeout,
+        maxInFlightRequests,
+      })
+      return new BrokerPoolType({
```
Nevon, line 87: Do we really need to have two different implementations here? What I'm thinking is, is there any difference between a connection pool that is not shared with any other instance, and just having a regular brokerpool? From what I can see, the only difference is in instrumentation event dispatching, but is there actually a benefit to not using that same code for "non-shared" connectionpools?

src/instrumentation/emitter.js
```diff
@@ -16,19 +18,35 @@ module.exports = class InstrumentationEventEmitter {
...
-   * @returns {Function} removeListener
+   * @param {(...args: any[]) => void} listener
+   * @returns {import("../../types").RemoveInstrumentationEventListener<string>} removeListener
    */
   addListener(eventName, listener) {
     this.emitter.addListener(eventName, listener)
     return () => this.emitter.removeListener(eventName, listener)
   }
+
+  forward(anotherInstrumentationEmitter) {
```
Nevon, line 45: Is it intentional here that you can subscribe the same forwarder more than once, or should `this.forwarders` perhaps be a set?

ankon:
> [latency, cpu usage, measuring]

Quite some time since the last update here. Part of the reason is a bit of a depressing news item: After really measuring things for a while, and a lot of code reading, it became clear that there is a fundamental problem here that I missed in the Kafka protocol documentation.

> The server guarantees that on a single TCP connection, requests will be processed in the order they are sent and responses will return in that order as well. The broker's request processing allows only **a single in-flight request per connection** in order to guarantee this ordering. Note that clients can (and ideally should) use non-blocking IO to implement request pipelining and achieve higher throughput. i.e., clients can send requests even while awaiting responses for preceding requests since the outstanding requests will be buffered in the underlying OS socket buffer. All requests are initiated by the client, and result in a corresponding response message from the server except where noted.

(From https://kafka.apache.org/protocol#protocol_network, highlights by me)

Multiplexing many requests onto the same connection is good and as written there also encouraged, but these requests will block each other, and increase the perceived latency for each request. This becomes very visible with fetch requests, which KafkaJS sends as quickly as it can, and Kafka will wait the given `maxWaitTimeInMs` for each request -- before processing any other, "obviously independent!" request.

Thinking about this further: In theory one could try to come up with a goal for the fetch requests (for example: "should take at most 100ms between two fetch _responses_"), and then tune the `maxWaitTimeInMs` option taking into account the other (fetch) requests on the same connection. In some cases this could be done statically (if the number of consumers and their goals is known), but in most cases this is probably something that would need to happen dynamically by looking at the time between the responses, and then injecting pauses if needed. Depending on the situation though it may be impossible to share the connection betwen two consumers, without compromising their goals.
It is also possible that consumers may miss their heartbeats and/or join timeouts simply because another consumer has just send out a long fetch request.

So, I'm a bit on the fence whether it actually makes sense to proceed here: _Are the potential benefits really worth the complexity of the changes?_

I suppose the JSDoc changes have a fair chance of passing that bar, and I could pull those out easily into a new PR. The smaller cleanups are probably also fine. 

What do you think?

Nevon:
> I suppose the JSDoc changes have a fair chance of passing that bar, and I could pull those out easily into a new PR. The smaller cleanups are probably also fine.

Let's do this regardless. Those changes should be quick and easy to review.

> So, I'm a bit on the fence whether it actually makes sense to proceed here: Are the potential benefits really worth the complexity of the changes?

I'm glad you thought about it and are taking the time to consider what the repercussions are. While I didn't exactly connect it to the section of the protocol documentation you mentioned, I did consider that we were likely going to see a trade-off between resource utilization and latency. As you say, during high contention, connection pooling will indeed increase latency - in some cases possibly to the point of it being problematic (a heartbeat having to wait for a fetch, for example).

The upside that I see is in reducing the amount of essentially idle connections in cases where you don't have a lot of contention. This obviously doesn't matter for the client, or at least not much, but it might matter for the brokers. However, that might be a concern better solved on the broker side via enforcement of quotas.

However, I definitely see this possibly turning into a footgun where pooled consumers work just fine during normal operations, but then starting to fail when one consumer starts having long-running requests impacting the heartbeats of the other consumer, or the like. Given that we have already encountered some pretty gnarly issues with the instrumentation emitters, I'm leaning towards it not being worth the effort. :/

> After really measuring things for a while

Any measurements you can share?

ankon:
> Any measurements you can share?

What I did was to use instrumentation events from the request queue/connection and expose them to Prometheus, so that I could look at the queue time and the actual "on the wire" time grouped by request type, and from that found that the timings I got were way to "stable": Having n consumers share the connection basically got me to n*maxWaitTimeMillis in times between the fetches. After staring at the graphs for a while and excluding any kind of "me being horribly stupid" I found [this discussion on the Kafka ML](https://mail-archives.apache.org/mod_mbox/kafka-dev/201609.mbox/%3cCAJDuW=C7mttPUm_bha9QBHTEFgwmtGZgxYCd35iyxOKPSfNg7w@mail.gmail.com%3e) and then from there the sentence in the documentation (thereby establishing that indeed I was "horribly stupid", just in a different way than expected :D).

The problem appears the most pronounced on fetch requests (if you tell the Kafka server to wait for 5s, it will do exactly that -- and not process any other request on that connection until that waiting is done), and of course to a lesser degree also for complex/big produce requests that require acks from all brokers. 

I guess one could come up with all kinds of heuristics here (for example the above "dynamically calculate maxWaitTime per fetch based on historic response latencies and insert waits if needed", but also interesting ones such as "send fetch requests through different connection than other requests" -- which probably will look sane until you work out the various exceptions you'd need to keep everything else working).

> I'm leaning towards it not being worth the effort.

Agreed. And, this PR and the linked issue can still serve as documentation if people are pondering about whether this may or may not be a good idea, so it's not completely wasted time either :)

Nevon:
Alright, then let's close this PR. Feel free to open up a new PR with just the JSDoc changes and whatever other little fixes you found along the way whenever you feel up to it.

PR #980, opened 2020-12-02
Update maxInFlightRequests to 5

mrnonz:
The kafkajs require Kafka version >= 0.11, so We can use maxInFlightRequests with value 5.

Here more document for guarantee with this value. https://issues.apache.org/jira/browse/KAFKA-5494

So, It's not required to set the maxInFlightRequests to 1. (It's only use for ordering message but not relate to transaction purpose).

### DIFF
--- a/docs/Transactions.md
+++ b/docs/Transactions.md
@@ -13,18 +13,18 @@ You initialize a transaction by making an async call to `producer.transaction()`
 
 > Note: Kafka requires that the transactional producer have the following configuration to _guarantee_ EoS ("Exactly-once-semantics"):
 >
-> - The producer must have a max in flight requests of 1
+> - The producer must have a max in flight requests of 5
 > - The producer must wait for acknowledgement from all replicas (acks=-1)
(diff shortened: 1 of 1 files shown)


Nevon:
While I haven't read through the entire implementation of this, I'm afraid it's not as simple as just setting it to 5. It requires that the client implement a certain behavior to allow for it first. Specifically from [KAFKA-5494](https://github.com/apache/kafka/pull/3743):

> Sequence number management on the client when there are request failures. When a batch fails, future inflight batches will also fail with OutOfOrderSequenceException. This must be handled on the client with intelligent sequence reassignment. We must also deal with the fatal failure of some batch: the future batches must get different sequence numbers when the come back.

[This document describes the design](https://docs.google.com/document/d/1EBt5rDfsvpK6mAPOOWjxa9vY0hJ0s9Jx9Wpwciy0aVo/edit)

Granted, I haven't looked at our transaction code for literal years, but I don't think KafkaJS is currently designed to adhere to those requirements.

PR #984, opened 2020-12-10
Prevent uncaught exception on errors.js when 'bugs' field is missing from package.json

EricGeigerQ:
Addresses edge case where uncaught exception can occur in errors.js if "bugs" field is missing from package.json.

Tests for 'bugs' field, and if missing, sets issueUrl to null. Side effect is that, if null, the `helpUrl` will not point to new issue path should a KafkaJSInvariantViolation error get thrown.

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -1,4 +1,5 @@
 const pkgJson = require('../package.json')
+const { bugs } = pkgJson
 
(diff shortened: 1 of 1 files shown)


tulios:
This is already available in `1.16.0-beta.1`

PR #985, opened 2020-12-11
Use Array.push() instead of spread/concat

t-d-d:
Found this via profiling.  The change made a significant improvement for my use case - restoring multi Gigabyte state stores with 10s millions of messages (passing the key and value Buffers directly on to leveldb.)

Full data load time:
Before: 136s
After: 39s

With an empty eachBatch function:
Before: 127s
After: 28s


tulios:
Great contribution, thanks @t-d-d

PR #992, opened 2020-12-29
New admin method for fetching group offsets for multiple topics

### LINKED ISSUE #989: fetchOffsets should allow fetching offsets for all topics consumed by a consumerGroup
I'm writing a Kafka plugin to [Backstage](https://backstage.io/) and want to use kafkajs. One of the features I'm developing is the ability to view all topics (+offsets) consumed by a consumer group so I need a way to fetch that from the Kafka cluster.

Currently the only API I see that I can use for that is `admin.fetchOffsets`. However, it [requires](https://github.com/tulios/kafkajs/blob/f7a166488321216a0feec4428f8e589b116eb31f/src/admin/index.js#L376) me to set the topic I want to fetch offset from even though the [internal method](https://github.com/tulios/kafkajs/blob/f7a166488321216a0feec4428f8e589b116eb31f/src/broker/index.js#L524:9) doesn't seem to require it. I think we can remove the assertion that `topic` is not null on calls to `admin.fetchOffsets` with `resolveOffsets===false`.

**Additional context**
[Backstage Issue #3791](https://github.com/backstage/backstage/issues/3791)

nirga:
Fixes #989 

This PR introduces a new method called `fetchConsumerGroupOffsets` that mimics the behavior `fetchOffsets` (which I now deprecated), but adds support for querying multiple topics at once, or omitting the `topics` parameter altogether and getting the group offsets in all consumed topics.

I chose to re-implement this method instead of changing `fetchOffsets` because this introduces an API breaking change: the result is now a list of topics and offsets (instead of just the offsets as it was in the original function).


src/admin/index.js
```diff
@@ -426,6 +426,82 @@ module.exports = ({
...
+    let consumerOffsets
+    if (topics.length) {
+      const topicsToFetch = Promise.all(
+        topics.map(async topic => {
+          const partitions = await findTopicPartitions(cluster, topic)
+          const partitionsToFetch = partitions.map(partition => ({ partition }))
+          return { topic, partitions: partitionsToFetch }
+        })
+      )
+      const { responses } = await coordinator.offsetFetch({
```
elithe1, line 451: Instead of  `let consumerOffsets`
consider writing const `{ responses: consumerOffsets } = await coordinator.offsetFetch(....`

src/admin/index.js
```diff
@@ -426,6 +426,82 @@ module.exports = ({
...
+        })
+      )
+
+      await Promise.all(
+        consumerOffsets.map(({ topic, partitions }) => setOffsets({ groupId, topic, partitions }))
+      )
+    }
+
+    return consumerOffsets
+      .filter(response => !topics.length || response.topic in topics)
```
Nevon, line 493: Isn't `topics` an array? `in` checks for the existence of a key in an object, not whether an item is in an array. Maybe you meant `topics.includes(response.topic)`?

nirga: Yeah it was just a draft PR and I'm not super familiar with javascript :)
I've removed this line altogether since I realized why would response not always include exactly the topics we requested?

src/admin/index.js
nirga: Thanks! It doesn't work though, got an `ReferenceError: consumerOffsets is not defined`

Nevon: Yeah since the intention is to assign to `consumerOffsets` in the outer scope, they are not equivalent operations.

src/admin/index.js
```diff
@@ -428,38 +431,43 @@ module.exports = ({
...
+    if (!Array.isArray(topics)) {
+      throw new KafkaJSNonRetriableError(`Expected topics array, got ${topics}`)
+    }
+
     const coordinator = await cluster.findGroupCoordinator({ groupId })
 
     let consumerOffsets
     if (topics.length) {
-      const topicsToFetch = Promise.all(
+      const topicsToFetch = await Promise.all(
```
Nevon, line 451: Huh, how could this test have been passing before? 🤔  I suspect owls in the bog, as we say in Swedish...

src/admin/index.js
```diff
@@ -428,38 +431,43 @@ module.exports = ({
...
+
       const { responses } = await coordinator.offsetFetch({
         groupId,
-        topicsToFetch,
+        topics: topicsToFetch,
       })
       consumerOffsets = responses
     } else {
-      const { responses } = await coordinator.offsetFetch({ groupId })
+      const { responses } = await coordinator.offsetFetch({ groupId }, [])
```
Nevon: What's with the empty array here?

Nevon:
Since `fetchOffsets` is deprecated, I would suggest removing the documentation for it and leaving only a "See `fetchConsumerGroupOffsets`" link, linking to the documentation for the new method.

Nevon:
I see two options:

1. Create a new method, as is done in this PR.
2. Change the signature of the old method so that it has two optional parameters, `topic: string` and `topics: string[]`. Internally, we could normalize the arguments so that `topic: "foo"` becomes `topics: ["foo"]`, to avoid having to have a bunch of duplicated code. We could eventually deprecate the `topic` argument in favor of `topics`. If `topic` and `topics` are both defined, we could throw an error.

I think I'm in favor of option 2 more than option 1, because it limits the number of methods on the client and migrating from the old signature to the new is very simple. Do you see any big problem with going with that approach, @nirga?

@tulios / @JaapRood: Any input from you two?

src/admin/index.js
```diff
@@ -426,6 +429,85 @@ module.exports = ({
...
+          return { topic, partitions: partitionsToFetch }
+        })
+      )
+
+      const { responses } = await coordinator.offsetFetch({
+        groupId,
+        topics: topicsToFetch,
+      })
+      consumerOffsets = responses
+    } else {
```
roeezolantz, line 464: @nirga Why do you need this if-else?
Given that you’re in this part of the code, `topics` must be array, and if its `length === 0` - `Promise.all([])` will result an empty array in `topicsToFetch`, so you’ll actually send `topics: []` to `offsetFetch` either way

src/admin/index.js
```diff
@@ -426,6 +429,85 @@ module.exports = ({
...
+   * @param {string[]} topics
+   * @param {boolean} [resolveOffsets=false]
+   * @return {Promise}
+   */
+  const fetchConsumerGroupOffsets = async ({ groupId, topics = [], resolveOffsets = false }) => {
+    if (!groupId) {
+      throw new KafkaJSNonRetriableError(`Invalid groupId ${groupId}`)
+    }
+
+    if (!Array.isArray(topics)) {
```
roeezolantz, line 443: In the previous implementation if `topic` wasn't passed an exception was thrown
```
    if (!topic) {
      throw new KafkaJSNonRetriableError(`Invalid topic ${topic}`)
    }
```
but here you're setting `topics`'s default value to pass your exception throwing (`[]`), so you're actually not validating that this param was sent here

src/admin/index.js
```diff
@@ -426,6 +429,85 @@ module.exports = ({
...
+      await Promise.all(
+        consumerOffsets.map(({ topic, partitions }) => setOffsets({ groupId, topic, partitions }))
+      )
+    }
+
+    return consumerOffsets.map(({ topic, partitions }) => {
+      const completePartitions = partitions.map(({ partition, offset, metadata }) => ({
+        partition,
+        offset,
+        metadata: metadata || null,
```
roeezolantz, line 504: Why not just set it as optional instead?

src/admin/index.js
```diff
@@ -426,6 +429,85 @@ module.exports = ({
...
+          })
+
+          return {
+            topic,
+            partitions: recalculatedPartitions,
+          }
+        })
+      )
+
+      await Promise.all(
```
roeezolantz, line 495: why do you need that in a separated `Promise.all` execution? putting it inside the previous `Promise.all` will avoid starting the setOffset process just after the longest offset fetching finished (you'll set the offset foreach fetched one which might improve total performance)

src/admin/index.js
nirga: mm yeah but I think it's kinda weird to force users to set topics=[] if they want to just fetch all of topics of the consumer instead of just calling it with `fetchConsumerGroupOffsets({groupId})`.

nirga:
@Nevon I agree with you. I moved the changes back to `fetchOffset` and added logic go support passing either `topic` or `topics` (tweaking the returned value accordingly). I've also updated the docs.

src/admin/index.js
nirga: WDYM? I'm not familiar with optional on JS  🤔  (it's not TS)

src/admin/index.js
nirga: It needs to be passed to `offsetFetch` so that it will fetch all topics (it actually sets it in the protocol it sends to Kafka from what I've seen).

src/admin/index.js
nirga: Yeah that was a change from the draft PR that had no tests so ofc it didn't cause any issues 😄

docs/Admin.md
```diff
@@ -180,18 +180,32 @@ await admin.fetchTopicOffsetsByTimestamp(topic, timestamp)
...
+//     topic: 'topic2',
+//     partitions: [
+//       { partition: 0, offset: '1234' },
+//       { partition: 1, offset: '4567' },
+//     ],
+//   },
 // ]
 ```
 
+Omit `topics` altogether if you want to get the consumer group offsets for all of its consumed topics.
```
Nevon, line 207:
```suggestion
Omit `topics` altogether if you want to get the consumer group offsets for all topics with committed offsets.
```

docs/Admin.md
```diff
@@ -180,18 +180,32 @@ await admin.fetchTopicOffsetsByTimestamp(topic, timestamp)
 
 ## <a name="fetch-offsets"></a> Fetch consumer group offsets
 
-`fetchOffsets` returns the consumer group offset for a topic.
+`fetchOffsets` returns the consumer group offset for a list of topics.
 
 ```javascript
-await admin.fetchOffsets({ groupId, topic, })
+await admin.fetchOffsets({ groupId, topics, })
```
Nevon, line 186: Removing trailing comma and making the input match the example output.

```suggestion
await admin.fetchOffsets({ groupId, topics: ['topic1', 'topic2'] })
```

src/admin/__tests__/fetchOffsets.spec.js
```diff
@@ -40,11 +42,16 @@ describe('Admin', () => {
...
 
-    test('throws an error if the topic name is not a valid string', async () => {
-      await expect(admin.fetchOffsets({ groupId: 'groupId', topic: null })).rejects.toHaveProperty(
-        'message',
-        'Invalid topic null'
-      )
+    test('throws an error if the topics argument is not a valid list', async () => {
+      await expect(
+        admin.fetchOffsets({ groupId: 'groupId', topics: topicName })
+      ).rejects.toHaveProperty('message', `Expected topic or topics array to be set`)
```
Nevon, line 48: We should make sure that this is a non-retriable error, which we can do using the `toThrow` method, which takes a constructor and an optional string that it compares to the message.

```suggestion
      await expect(broker.apiVersions()).rejects.toThrow(
        KafkaJSNonRetriableError,
        `Invalid argument "topics". Expected array of strings, got: "${topicName}"`
      )
```

Nevon: Same for the test below

src/admin/__tests__/fetchOffsets.spec.js
```diff
@@ -71,6 +78,66 @@ describe('Admin', () => {
...
+      expect(new Set(offsets)).toEqual(
+        new Set([
+          {
+            topic: yetAnotherTopicName,
+            partitions: [{ partition: 0, offset: '42', metadata: null }],
+          },
+          { topic: anotherTopicName, partitions: [{ partition: 0, offset: '23', metadata: null }] },
+          { topic: topicName, partitions: [{ partition: 0, offset: '13', metadata: null }] },
+        ])
+      )
```
Nevon, line 112: That's kind of neat. We use the package [`jest-extended`](https://github.com/jest-community/jest-extended) though, which has a matcher exactly for this, called [`toIncludeSameMembers`](https://github.com/jest-community/jest-extended#toincludesamemembersmembers).

```suggestion
      expect(offsets).toIncludeSameMembers([
        {
          topic: yetAnotherTopicName,
          partitions: [{ partition: 0, offset: '42', metadata: null }],
        },
        { topic: anotherTopicName, partitions: [{ partition: 0, offset: '23', metadata: null }] },
        { topic: topicName, partitions: [{ partition: 0, offset: '13', metadata: null }] },
      ])
```

Nevon: Similarly for the test below.

src/admin/index.js
```diff
@@ -371,60 +371,88 @@ module.exports = ({
   /**
    * @param {string} groupId
    * @param {string} topic
+   * @param {string[]} topics
```
Nevon, line 374: I'm not sure how to write the JSDoc for this, now that the interplay between `topic` and `topics` is more complicated.

Both `topic` and `topics` are optional. If neither is set, the default is for `topics` to get the value `[]`. Do you know if there's any way to express that in JSDoc, @ankon? Not a big deal, but I just don't know.

src/admin/index.js
Nevon:
```suggestion
        metadata,
```

There's no optional in JS, but I guess what he means is omitting the property or setting it to undefined, rather than explicitly setting it to `null`. There may be some exceptions, as I'm sure we haven't been completely consistent, but in general if a property is considered optional we either omit it or set it to `undefined` rather than `null` - at least from a quick `grep` around the codebase.

Nevon: Ah, but that would change the typescript return type, so let's not change this.

types/index.d.ts
```diff
@@ -434,7 +434,8 @@ export type Admin = {
   fetchTopicMetadata(options?: { topics: string[] }): Promise<{ topics: Array<ITopicMetadata> }>
   fetchOffsets(options: {
     groupId: string
-    topic: string
+    topic?: string
```
Nevon, line 437: Github won't let me create a suggestion that goes outside of the lines of the diff, but there's an issue here where the return type of this function is different depending on the input, so we need to be a little more clever here. There's probably some fancy-pants thing you can do with conditional types, but we can also just use plain old method overloading:

```ts
interface FetchOffsetsBaseOptions {
  groupId: string,
  resolveOffsets?: boolean
}

export interface FetchOffsetsTopicOptions extends FetchOffsetsBaseOptions {
  topic: string
}

export interface FetchOffsetsTopicsOptions extends FetchOffsetsBaseOptions {
  topics: string[]
}

export type Admin = {
  fetchOffsets(options: FetchOffsetsTopicOptions): Promise<Array<PartitionOffset & { metadata: string | null }>>,
  fetchOffsets(options: FetchOffsetsTopicsOptions): Promise<Array<{ topic: string, partitions: Array<PartitionOffset & { metadata: string | null }>}>>,
}
```

This has a few benefits. The first is that the type system will prevent `admin.fetchOffsets({ topic: 'foo', topics: ['bar'] })`, because the `topic` and `topics` properties are mutually exclusive. The second is that it allows us to define the relationship between the input type and the return type. The return type is a bit of an abomination. If we didn't return `metadata` it could be simplified to just `Promise<TopicOffsets[]>` and `Promise<PartitionOffset[]>`, but alas, here we are.

types/tests.ts
```diff
@@ -180,6 +180,9 @@ const runAdmin = async () => {
 
   await admin.listTopics()
 
+  await admin.fetchOffsets({ groupId: 'test-group', topic: 'topic1' })
```
Nevon, line 183: Maybe toss an `await admin.fetchOffsets({ groupId })` in there as well to show that `topics` is optional.

Nevon (changes requested):
Looks good overall. Just some fairly minor things to fix.

src/admin/index.js
roeezolantz: Is this `null` necessary ? or can it just be `undefined` instead? 
I mean that you can set the `metadata` field as optional in the types file
not sure about the real syntax but maybe something like that :
```js
  fetchOffsets(options: {
...
  }): Promise<Array<SeekEntry & { metadata?: string | null }
```

and then you'll be able to remove all that
```js
    const result = consumerOffsets.map(({ topic, partitions }) => {
      const completePartitions = partitions.map(({ partition, offset, metadata }) => ({
        partition,
        offset,
        metadata: metadata || null,
      }))

      return { topic, partitions: completePartitions }
    })
```
and just work on `consumerOffsets` instead of `result`

Nevon: It _could_ definitely be undefined rather than null, but the problem is that then we are changing the return type of an existing method, which would be a backwards-incompatible change. Not worth introducing that for such a small improvement.

src/admin/index.js
ankon: Interesting, from the top of my head: I think this might be too much to express with JSDoc. There could be a way to do it in the TypeScript definition, and then you could do magic with `@type` here. 

But it also might not be needed: If these are optional, then from the callers point that's all they need to know. The specific defaulting could just be considered an implementation detail.

I guess I would just go with
```js
/**
 * Function does stuff.
 *
 * Note that `topics` will be defaulted to `[]` if neither `topic` nor `topics` is provided.
 * 
 * @param {string} groupId
 * @param {string} [topic]
 * @param {string[]} [topic]
 */
```

nirga: I added comments as @ankon suggested, inline in the parameter list as well.

types/index.d.ts
nirga: Done. I've opted to keep the options parameter inline though, but moved the `Array<PartitionOffset & { metadata: string | null }>` to a new type. LMK if you prefer the options to be extracted to a type as well.

nirga:
@Nevon Fixed everything. Thanks so much for your review and responsiveness!

types/index.d.ts
```diff
@@ -436,7 +439,12 @@ export type Admin = {
     groupId: string
     topic: string
     resolveOffsets?: boolean
-  }): Promise<Array<SeekEntry & { metadata: string | null }>>
+  }): Promise<FetchOffsetPartitions>
```
Nevon, line 442: This isn't correct, is it? This indicates that the returned promise resolves to something like:

```js
{
  partitions: [{ partition: '1', offset: '123', metadata: null }]
}
```

But unless I'm mistaken, it actually resolves to:

```js
[{ partition: '1', offset: '123', metadata: 'foo' }]
```

I guess this type should be unchanged from before:

```ts
Array<SeekEntry & { metadata: string | null }>
```

Btw, prefer to use `PartitionOffset` over `SeekEntry`. They are aliases of each other, so functionally there's no difference, but eventually `SeekEntry` will be deprecated in favor of `PartitionOffset`, so it's good if we don't spread it further.

nirga: Right! @Nevon my mistake. Fixed that.

Nevon:
Great, thanks for your contribution! Sorry you had to wait so long for a review.

tulios:
> I see two options:
> 
> 1. Create a new method, as is done in this PR.
> 2. Change the signature of the old method so that it has two optional parameters, `topic: string` and `topics: string[]`. Internally, we could normalize the arguments so that `topic: "foo"` becomes `topics: ["foo"]`, to avoid having to have a bunch of duplicated code. We could eventually deprecate the `topic` argument in favor of `topics`. If `topic` and `topics` are both defined, we could throw an error.
> 
> I think I'm in favor of option 2 more than option 1, because it limits the number of methods on the client and migrating from the old signature to the new is very simple. Do you see any big problem with going with that approach, @nirga?
> 
> @tulios / @JaapRood: Any input from you two?

I guess you were too fast @Nevon 😅 
Glad that the PR was merged.

PR #995, opened 2020-12-30
fix assigner with builtin method topics

### LINKED ISSUE #994: Consumer for >= 1.13 crashes on some topic names
**Describe the bug**
I upgraded `kafkajs` to 1.15 from 1.12 and found a consumer was crashing with

```
TypeError: assignment[assignee][topicPartition.topic].push is not a function
```

After digging a little deeper, the problem was caused by having a topic called `shift`. What happens is that [here](https://github.com/tulios/kafkajs/blob/9a4065f4f8b61cfc4ed1a3359ca0c6dc7eae2eb6/src/consumer/assigners/roundRobinAssigner/index.js#L55) an array is created (instead of an object), and the builtin `shift` method is [mistaken](https://github.com/tulios/kafkajs/blob/9a4065f4f8b61cfc4ed1a3359ca0c6dc7eae2eb6/src/consumer/assigners/roundRobinAssigner/index.js#L58) for that topic's assignment.

**To Reproduce**

Call `assign` with a topic named `shift`, or any other array builtin method.

**Expected behavior**

This should not crash, I should be able to call my topics `shift` or even `__proto__`.

**Observed behavior**

The consumer crashes at startup.

**Environment:**
 - OS: Arch Linux
 - KafkaJS version: 1.15
 - Kafka version: not sure, shouldn't be relevant
 - NodeJS version: 12.20

I'm opening a PR with a fix.

asmeikal:
This fixes #994.

I would have used `hasOwnProperty` on line 58, but the linter prevented me from doing that, so I'm creating an object without a prototype to hold the assignments info and avoid conflicts with builtin methods.

### DIFF
--- a/src/consumer/assigners/roundRobinAssigner/index.js
+++ b/src/consumer/assigners/roundRobinAssigner/index.js
@@ -52,7 +52,7 @@ module.exports = ({ cluster }) => ({
       const assignee = sortedMembers[i % membersCount]
 
       if (!assignment[assignee]) {
-        assignment[assignee] = []
+        assignment[assignee] = Object.create(null)
       }
 
       if (!assignment[assignee][topicPartition.topic]) {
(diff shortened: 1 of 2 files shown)


Nevon:
Oh wow, that is quite the catch! Thanks for the contribution!

PR #998, opened 2021-01-07
Fix fetch offset type

Nevon:
One millisecond after merging #992 I noticed a few small things to improve in the typings. This removes an exported type, but given that the beta package containing it isn't even out yet, I highly doubt anyone is using it. 😅


ankon (approved):

JaapRood:
Really glad @ankon is picking up my maintainer slack, thanks :)

PR #1002, opened 2021-01-17
types: mark ConsumerConfig required

Amorites:


Nevon:
Thanks for the contribution. I can see how this type was marked as optional since we default to `{}` in the factory function, but given that we then throw if `groupId` isn't set, it doesn't make much sense.

Nevon (approved):

Issue #945, opened 2020-11-06
Throwing error out of eachBatch/eachMessage causes all concurrent partitions to restart processing their offset


hughlivingstone:
**Describe the bug**
We are using` partitionsConsumedConcurrently: 3 `so we concurrently consume messages.  However we are finding that when we throw an error out of our eachBatch/eachMessage handler it causes all partitions to start processing the messages again even though the partitions that did not fail are still being processed.

It leads to us processing the same message multiple times in parallel.  The log output of a test we ran illustrates it below. We are also seeing this same error in our production environment.
```
2020-11-06 13:32:23 info BEGIN processing message0 on partition 0
2020-11-06 13:32:23 info BEGIN processing message1 on partition 1
2020-11-06 13:32:23 info BEGIN processing message2 on partition 2
2020-11-06 13:32:26 info END Sleeping message0 on partition 0
  console.error node_modules/kafkajs/src/loggers/console.js:15
    {"level":"ERROR","timestamp":"2020-11-06T13:32:26.802Z","logger":"kafkajs","message":"[Runner] Error when calling eachMessage","topic":"hugh","partition":0,"offset":"0","stack":"Error: TEST SHOULD RETRY\n    at Runner.eachMessage (/Users/hughlivingstone/Dev/code/packages/testing-tools/integration-tests/kafka.integration.test.ts:82:17)\n    at Runner.processEachMessage (/Users/hughlivingstone/Dev/code/packages/testing-tools/node_modules/kafkajs/src/consumer/runner.js:185:9)\n    at onBatch (/Users/hughlivingstone/Dev/code/packages/testing-tools/node_modules/kafkajs/src/consumer/runner.js:321:9)\n    at /Users/hughlivingstone/Dev/code/packages/testing-tools/node_modules/kafkajs/src/consumer/runner.js:373:15","error":{}}

2020-11-06 13:32:26 info BEGIN processing message2 on partition 2
2020-11-06 13:32:26 info BEGIN processing message1 on partition 1
2020-11-06 13:32:28 info END processing message1 on partition 1
2020-11-06 13:32:29 info END processing message2 on partition 2
2020-11-06 13:32:31 info END processing message1 on partition 1
2020-11-06 13:32:32 info END processing message2 on partition 2
```
I have attached a file below that we used to produce a similar output to above.
[kafka.concurrency.test.ts.zip](https://github.com/tulios/kafkajs/files/5501539/kafka.concurrency.test.ts.zip)

[...]

PR #1015, opened 2021-01-31
On batch error, wait for in-progress processing and ignore remaining batches

t-d-d:
Proposed fix for https://github.com/tulios/kafkajs/issues/945

On a batch processing error:
- wait for any in-progress (concurrent) batch processing to complete
- skip all remaining batches
Before re-throwing the error to the fetch retrier.

2 test cases added that demonstrate the issue.

An alternative would be to not skip the remaining batches but let them all be processed.

I think the code could be made more readable but I've intentionally minimised the changes in this PR.


src/consumer/runner.js
```diff
@@ -322,31 +323,33 @@ module.exports = class Runner extends EventEmitter {
       }
 
       enqueuedTasks.push(async () => {
-        const batches = await result.value
+        let batches
+        try {
+          batches = await result.value
```
Nevon, line 328: Not sure if my brain is just bad or if this code is awful (not your changes specifically, but rather the fetch loop in general). It's like playing 8 dimensional chess at times.

So I get the general idea is to basically try to collect any error that happens, and if there are any, abort any future batch processing before it happens. What I don't understand is that if `result.value` throws, then the processing would never get scheduled (i.e. we'd never reach `batches.map(batch => concurrently(...))`), so wouldn't this achieve the same thing?

```js
let batches
try {
  batches = await result.value
} catch (e) {
  unlockWithError(e)
}
```

If we throw the error here, the lock would never be unlocked, since the `Promise.all` would reject.

t-d-d: > if result.value throws, then the processing would never get scheduled (i.e. we'd never reach batches.map(batch => concurrently(...)))

Yes, for that broker response.  But processing work has already been added to the concurrency queue from previous broker responses.  And future broker responses could also add work to the queue.

My understanding of how it works:  There are 2 things happening concurrently
1) broker fetch requests are resolving and as they do, processing tasks are added to the concurrency queue
2) concurrency queue processing
An error can occur in either of these.  In either case, we don't want to do any more processing.

You comment has made me realise that the current proposal is still not complete.  If a broker fetch fails and we throw, and the Promise.all() throws, we are still throwing/rejecting too soon - there could be some processing tasks executing (although no more will happen after those.)  We should still wait for the barrier lock to resolve.  I'll push another change to fix that.

I'll also see if a couple of comments can help here.  But if you think the code is too far gone to be readable/ maintainable, I could have a go at re-working it.

src/consumer/runner.js
```diff
@@ -322,43 +323,47 @@ module.exports = class Runner extends EventEmitter {
...
             try {
-              if (!this.running) {
+              // If stopping or any error has occurred, don't process (or resolve) any more batches
+              if (!this.running || error || batch.isEmpty()) {
                 return
               }
 
-              if (batch.isEmpty()) {
-                return
+              if (!batch.isEmpty()) {
```
Nevon, line 341: This is redundant.

src/consumer/runner.js
```diff
@@ -295,13 +295,14 @@ module.exports = class Runner extends EventEmitter {
...
 
-    const { lock, unlock, unlockWithError } = barrier()
+    const { lock, unlock } = barrier()
     const concurrently = limitConcurrency({ limit: this.partitionsConsumedConcurrently })
 
     let requestsCompleted = false
     let numberOfExecutions = 0
     let expectedNumberOfExecutions = 0
     const enqueuedTasks = []
+    let error = null // the first fetch request, processing or heartbeat error encountered
```
Nevon, line 305: Instead of keeping this error reference and making sure that we catch everywhere that can throw, we can query the state of the `lock`. If it is unlocked, it means that we should skip any further processing.

We'd just need to expose the lock state from the barrier:

```js
module.exports = () => {
  let unlock
  let unlockWithError
  let locked = true
  const lock = new Promise(resolve => {
    unlock = () => {
      locked = false
      resolve()
    }
    unlockWithError = error => {
      locked = false
      resolve(error)
    }
  })

  return { lock, locked, unlock, unlockWithError }
}
```

Then when we want to check if we should skip the batch, instead of checking for `error` we check for `locked`.

There are two benefits from this:
1. We don't need this additional reference to the error, since this is already kept in the lock. Less moving parts to keep track of.
2. If we miss updating the error reference somewhere that could throw, this will still be handled by the `.catch` on each `enqueuedTask`

src/consumer/__tests__/consumeMessages.spec.js
```diff
@@ -205,6 +206,99 @@ describe('Consumer', () => {
     }
   })
 
+  it('after an error, aborts pending batch processing before retrying', async () => {
```
Nevon, line 209: Let's make sure that these tests are verifying what we think they're verifying. For this one, it means we expect eachBatch to have been called *once* for any partition above the concurrency limit, and *twice* for any partition within the concurrency limit.

Meaning if we have 3 partitions and a concurrency limit of 2, we expect eachBatch to have been called twice for partitions 1 and 2, and just once for partition 3 (because we skip it since there was a previous error).

src/consumer/__tests__/consumeMessages.spec.js
```diff
@@ -205,6 +206,99 @@ describe('Consumer', () => {
...
+
+    let inProgress = 0
+    let eachBatchCallCount = 0
+    let exceededConcurrencyLimit = false
+
+    const messagesConsumed = []
+    consumer.run({
+      partitionsConsumedConcurrently,
+      eachBatch: async ({ batch: { messages } }) => {
+        if (eachBatchCallCount++ === partitionsConsumedConcurrently - 1)
```
Nevon, line 274: So if I understand things correctly, this is because we want to make sure that we have already started processing some partitions before we throw, but we want to make sure we throw before starting the next partition after the concurrency limit.

src/consumer/__tests__/consumeMessages.spec.js
```diff
@@ -205,6 +206,99 @@ describe('Consumer', () => {
...
+        return { key: `key-${value}`, value: `value-${value}` }
+      })
+
+    await producer.send({ acks: 1, topic: topicName, messages })
+    await waitForMessages(messagesConsumed, { number: messages.length })
+    expect(messagesConsumed.length).toEqual(messages.length)
+    expect(exceededConcurrencyLimit).toBeFalse()
+  })
+
+  it('after an error, completes in-progress batch processing before retrying', async () => {
```
Nevon, line 255: This one is interesting. Given 3 partitions and a concurrency limit of 2. If partition 2 throws and we allow partition 1 to finish before throwing and retrying, does it mean that partition 1 gets re-processed, or do the offsets for partition 1 get resolved so that we skip those and only actually do any re-work for partition 2?

Again, this test needs some more explicit assertions to check that it's actually verifying what it says it's verifying. If the above is true, then there should be some way of asserting that partition 1 was only processed once.

Nevon (changes requested):

src/consumer/runner.js
t-d-d: But the point is that we don't want to unlock the barrier as soon as we get an error.  If we do that then the barrier promise and thus the fetch() promise settle before any already started batch processing has completed.
We could move the error state to the barrier.  I'll try that - let's see if its clearer.

t-d-d: @Nevon I have re-worked it to something I think shows more clearly what is going on.  Let me know if you agree.  I'm aware the tests still need improvements.

src/consumer/barrier.js
```diff
@@ -1,14 +1,20 @@
 /**
  * @template T
- * @return {{lock: Promise<T>, unlock: (v?: T) => void, unlockWithError: (e: Error) => void}}
+ * @return {{lock: Promise<T>, unlock: (v?: T) => void, setError: (e: Error) => void, error: Error}}
```
Nevon, line 3:
```suggestion
 * @return {{lock: Promise<T>, unlock: (v?: T) => void, setError: (e: Error) => void, error: Error | null}}
```

src/consumer/runner.js
Nevon: I understand your point now. So you want to keep the lock so that we wait for all processing to complete, but you need to keep a reference to the error so that it can be thrown at the end. I think it makes sense.

The one thing I'm wondering about is whether or not we can have multiple errors - since currently we're only keeping a single reference. Perhaps `setError` should be wrapping all the errors [in an AggregateError](https://github.com/tulios/kafkajs/blob/master/src/errors.js#L235-L241) so that we don't lose information.

t-d-d: @Nevon sorry, seems I didn't push.  Anyway, I think we both understanding it the same now.
The changes I just pushed are in line with what you suggest - we collect all the errors.  I'll add to it to throw the aggregateError (instead of just the first) asap.  And update test cases.

t-d-d: Re-worked to always throw an AggregateError from fetch().

src/consumer/__tests__/consumeMessages.spec.js
t-d-d: Re-worked the tests to be more focussed

src/consumer/__tests__/consumeMessages.spec.js
t-d-d: Re-worked the tests to be more focused and explicit

t-d-d:
@Nevon I re-worked this.  The change is bigger but I think it makes the code more readable.  As @Nevon suggested, we explicitly use a AggregateError for fetch.
1 change since before: I decided not to abandon already received batched when we get a fetch request error but to continue.  Errors while processing the batches still cause queued batches to be abandoned.

src/consumer/__tests__/consumeMessages.spec.js
t-d-d: I re-worked the tests cases.  please take another look.

matray:
@Nevon any chance this can be reviewed? Would help resolve experienced issues of #945

Issue #1025, opened 2021-02-11
HeartbeatInterval not respected for parallel heartbeat calls


Nevon:
If you make multiple calls to `heartbeat` in parallel, the `heartbeatInterval` will not be respected. For example:

```js
await Promise.all([
  heartbeat(),
  heartbeat(),
  heartbeat(),
])
```

This will cause 3 hearbeat requests to be made, even though after the first one succeeds, the rest should be skipped since the last heartbeat was < `heartbeatInterval` ms ago. This could be an issue if you for example heartbeat after each message in a batch with parallel message processing, and you end up making 50 heartbeats at the same time.

https://github.com/tulios/kafkajs/blob/e1a9d9eecd4cfd29e719f1c02cae6e544df0894a/src/consumer/consumerGroup.js#L386-L401

We'll need to add a lock before aquiring the value of `lastRequest`.

XBeg9:
@Nevon looks like debouncing should be implemented here, what do you think?

Nevon:
Kind of, but it's not just a matter of a regular debounce, as there are some additional concerns to keep in mind:

1. We should still not resolve the promises until one has resolved.
2. If the promise rejects, should we reject all waiting promises as well, or let the next one try (I'm leaning towards the latter).
3. We should only update the `lastRequest` if the heartbeat actually succeedeed.

I think it's as simple as just adding a lock on line 389 and then releasing it after you either succeed or fail to heartbeat. We already have the lock functionality built, so it's just a matter of dropping that in.

t-d-d:
I have an implementation for this as part of a POC I'm working on for independent broker fetches.

>or let the next one try

I didn't go with this - I let them all fail.  (They actually all share the same pending promise.)
I'll raise a draft PR for discussion.

tulios:
> We should only update the lastRequest if the heartbeat actually succeeded.
This is already the case @Nevon 

> If the promise rejects, should we reject all waiting promises as well, or let the next one try (I'm leaning towards the latter).
In this case, it doesn't matter. One failed heartbeat will not break the consumer. In the next event, a new heartbeat will happen, so I would say that the best case is to reject all and wait for the next attempt.

I think @t-d-d can work, or a lock as you mentioned previously.

t-d-d:
Yes, in this case (for the heartbeat) a lock would work as there is an interval and subsequent calls will short-circuit and not heartbeat.
[...]

PR #1026, opened 2021-02-12
Shared concurrent heartbeat() promise

t-d-d:
All concurrent heartbeat() calls will return the same promise object.  Proposed solution to https://github.com/tulios/kafkajs/issues/1025

Added a test that reproduces https://github.com/tulios/kafkajs/issues/1025 but more tests needed.

### DIFF
--- a/src/consumer/consumerGroup.js
+++ b/src/consumer/consumerGroup.js
@@ -4,6 +4,7 @@ const BufferedAsyncIterator = require('../utils/bufferedAsyncIterator')
 const websiteUrl = require('../utils/websiteUrl')
 const arrayDiff = require('../utils/arrayDiff')
 const createRetry = require('../retry')
+const sharedPromiseTo = require('../utils/sharedPromiseTo')
 
 const OffsetManager = require('./offsetManager')
 const Batch = require('./batch')
@@ -107,6 +108,8 @@ module.exports = class ConsumerGroup {
     this.subscriptionState = new SubscriptionState()
 
     this.lastRequest = Date.now()
+
+    this.sharedPromiseToToHeartbeat = sharedPromiseTo()
   }
 
   isLeader() {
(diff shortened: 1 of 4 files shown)


src/utils/sharedPromiseTo.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * @template T
+ * @param { () => Promise<T> } [defaultFunc]
+ * @returns { (func?: () => Promise<T>) => Promise<T> }
+ */
+module.exports = defaultFunc => {
```
tulios, line 6: Do we need `defaultFunc`? It's currently not used anywhere. I would also use curly braces and make this a little bit less dense, wdyt?

src/consumer/consumerGroup.js
```diff
@@ -384,20 +387,22 @@ module.exports = class ConsumerGroup {
   }
 
   async heartbeat({ interval }) {
-    const { groupId, generationId, memberId } = this
-    const now = Date.now()
+    return await this.sharedPromiseToToHeartbeat(async () => {
```
tulios, line 390: We can just return `this.sharedPromiseToToHeartbeat`, no need to `await`.

src/consumer/consumerGroup.js
```diff
@@ -107,6 +108,8 @@ module.exports = class ConsumerGroup {
     this.subscriptionState = new SubscriptionState()
 
     this.lastRequest = Date.now()
+
+    this.sharedPromiseToToHeartbeat = sharedPromiseTo()
```
tulios, line 112: Maybe we should make this private, using symbols, wdyt?

tulios (changes requested):
@t-d-d thanks a lot for the PR. I left some comments. They are basically suggestions; the only one that I would like to see changed is making the utility a little less dense, so it's easier to grasp.

src/utils/sharedPromiseTo.js
t-d-d: I think you are right - it is trying to be too clever.  I had both options because it's slightly easier to use the 'invocation time' method from a class (well, less code changes are needed to adopt it.)

I have changed to only provide option at initialisation time.  Let me know what you think.

t-d-d:
Actually this shouldn't be merged until we conclude the issue of what to do with failed heartbeats that @Nevon raises in https://github.com/tulios/kafkajs/issues/1025 .
My thinking is that there's no point in queuing up heartbeat calls that will get issued one-by-one when they fail.  I presume that in most cases, if one fails they are all going to fail.

tulios (approved):

Issue #1046, opened 2021-03-17
Commits to unassigned partitions and unnecessary duplicate processing after rebalance


feydan:
**Describe the bug**
While testing 1.15.0, I noticed a consumer successfully commit to an unassigned partition after a rebalance.

Additionally, I experienced decent amount of "duplicate" message delivery when running multiple consumers with periodic rebalances that can potentially be avoided.  I did not see any "duplicate" message delivery when using 1.12 after [this patch](https://github.com/tulios/kafkajs/commit/a5c3662)

**To Reproduce**
https://github.com/feydan/kjs-rebalance-test
(Setup/repro steps in README) -- I used two instances of `runner.js`

**Expected behavior**
Offset Commits should not be allowed for unassigned partitions.

Additionally, it would be great to have a mechanism to prevent processing or skipping messages from batches in older group generations and/or partitions no longer assigned to the consumer to reduce the potential for duplicate processing.

**Observed behavior**
I see offset commit requests and successful responses for unassigned partitions

My handler is processing messages from unassigned partitions after a rebalance group join with new partition assignments.

Here are some of the logs (the full payload would be too much).  This happens reliably with the test repository above:
```bash
{"level":"INFO","timestamp":"2021-03-17T20:52:25.031Z","logger":"kafkajs","message":"[ConsumerGroup] Consumer has joined the group","groupId":"group.46","memberId":"test-36c55fd0-d27f-4d48-873a-50ee83a15c74","leaderId":"test-36c55fd0-d27f-4d48-873a-50ee83a15c74","isLeader":true,"memberAssignment":{"rebalance.test.46":[0,2,4]},"groupProtocol":"RoundRobinAssigner","duration":15}
...
{"level":"DEBUG","timestamp":"2021-03-17T20:52:28.320Z","logger":"kafkajs","message":"[Connection] Response Heartbeat(key: 12, version: 3)","broker":"127.0.0.1:9092","clientId":"test","error":"The group is rebalancing, so a rejoin is needed","correlationId":18,"payload":{"type":"Buffer","data":"[filtered]"}}
...
[...]

PR #1047, opened 2021-03-18
Adding generation id check to isStale

feydan:
Adding a check to isStale to see if the current generationId matches the generationId when the batch was created.

Intended as a partial solution for unnecessary duplicate processing as described in #1046.  Does not address the issue of commits to unassigned partitions.


Nevon:
Is this still relevant after #1015?

feydan:
I believe #1015 will solve the issues I described in #1046.  I'm happy to close this and potentially revisit once #1015 is merged and I start testing again.

Issue #598, opened 2019-12-13
KafkaJSProtocolError: The broker received an out of order sequence number


Delapouite:
Hi

I have a single producer:

```
const producer = kafka.producer({
  allowAutoTopicCreation: false,
  idempotent: true,
  maxInFlightRequests: 1
});
```

As you can see, I'm using the `idempotent` feature introduced in https://github.com/tulios/kafkajs/pull/203 I'm also forcing the `maxInFlightRequests` to `1` 

It works fine when my producer sends messages to the kafka server at a slow pace. By that I mean no requests are stored in the `RequestQueue`.

But during spike in activity, sometimes messages are automatically sent in batch.
Kafka then returns the following error: 

```
ERROR [ReplicaManager broker=0] Error processing append operation on partition TOPIC (kafka.server.ReplicaManager)
kafka_1                 | org.apache.kafka.common.errors.OutOfOrderSequenceException: Out of order sequence number for producerId 74: 158 (incoming seq. number), 156 (current end sequence number)
```

In this bug case, we have `156` → `158` jump, because 2 messages where sent in the batch.

According the kafka java code, it expects only a 1 unit jump so `156` → `157`:

https://github.com/apache/kafka/blob/fecb977b257888e2022c1b1e04dd7bf03e18720c/core/src/main/scala/kafka/log/ProducerStateManager.scala#L241-L251

My current understanding/intuition, is that the `firstSequence` number should always be incremented by `1` and not by the number of messages present in the batch request.

Which roughly means that the `updateSequence` method of the `EOSManager` should not take an `increment` param but always increment by 1 only.
https://github.com/tulios/kafkajs/blob/eae5ea42fc8ac035b0591a705b81829e3fedf8ad/src/producer/eosManager/index.js#L171-L197

I'm still quite noob on this codebase so my reasoning is probably flawed.

@ianwsperber , as you're the main implementer of the eos code, what's your take on this? Thanks a lot.

Delapouite:
Here's a minimal reproducible case:

```js
const { Kafka, logLevel } = require('kafkajs')

const delay = Number(process.argv[2]) || 100
console.log({ delay })

const kafka = new Kafka({
	clientId: 'test-app',
	brokers: ['localhost:9092'],
})

const createProducer = async () => {
	const producer = kafka.producer({
		allowAutoTopicCreation: false,
		idempotent: true,
		maxInFlightRequests: 1,
	})
	await producer.connect()
[...]

PR #1050, opened 2021-03-21
Concurrent produce() sequence number fix

### LINKED ISSUE #1005: The idempotent producer doesn't work properly when sending R request, R > # of partitions
**Describe the bug**
Said the # of the partitions = N.
R > N

Put R requests, but only the first N requests are caught by consumer , and the left R - N requests are gone. 
And I don't get any rejections. (the `await Promise.all(promises)` doesn't throw any errors)
If I set idempotent to false then I can exactly R requests in consumer.

**To Reproduce**
https://github.com/hiroya8649/kafka-report

```
node consumer.js
node producer.js
```

**Expected behavior**
If set idempotent to true, the consumer should catch the same amount of requests send by the producer.

**Observed behavior**
I loop and put 5 requests, here is the result of producer and consumer:

producer
```
InstrumentationEvent {
  id: 0,
  type: 'producer.network.request',
  timestamp: 1611581796507,
  payload: {
    broker: 'localhost:9092',
    clientId: 'my-app',
    correlationId: 0,
    size: 338,
    createdAt: 1611581796506,
    sentAt: 1611581796507,
    pendingDuration: 1,
    duration: 0,
    apiName: 'ApiVersions',
    apiKey: 18,
    apiVersion: 2
  }
}
InstrumentationEvent {
  id: 1,
  type: 'producer.network.request',
  timestamp: 1611581796513,
  payload: {
    broker: 'localhost:9092',
    clientId: 'my-app',
    correlationId: 1,
    size: 1813,
    createdAt: 1611581796512,
    sentAt: 1611581796513,
    pendingDuration: 1,
    duration: 0,
    apiName: 'Metadata',
    apiKey: 3,
    apiVersion: 6
  }
}
InstrumentationEvent {
  id: 2,
  type: 'producer.network.request',
  timestamp: 1611581796515,
  payload: {
    broker: 'hao-MS-7B85:9092',
    clientId: 'my-app',
    correlationId: 0,
    size: 20,
    createdAt: 1611581796515,
    sentAt: 1611581796515,
    pendingDuration: 0,
    duration: 0,
    apiName: 'InitProducerId',
    apiKey: 22,
    apiVersion: 1
  }
}
InstrumentationEvent {
  id: 3,
  type: 'producer.network.request',
  timestamp: 1611581796516,
  payload: {
    broker: 'hao-MS-7B85:9092',
    clientId: 'my-app',
    correlationId: 1,
    size: 177,
    createdAt: 1611581796516,
    sentAt: 1611581796516,
    pendingDuration: 0,
    duration: 0,
    apiName: 'Metadata',
    apiKey: 3,
    apiVersion: 6
  }
}
InstrumentationEvent {
  id: 4,
  type: 'producer.network.request',
  timestamp: 1611581796523,
  payload: {
    broker: 'hao-MS-7B85:9092',
    clientId: 'my-app',
    correlationId: 2,
    size: 59,
    createdAt: 1611581796522,
    sentAt: 1611581796522,
    pendingDuration: 0,
    duration: 1,
    apiName: 'Produce',
    apiKey: 0,
    apiVersion: 7
  }
}
InstrumentationEvent {
  id: 5,
  type: 'producer.network.request',
  timestamp: 1611581796523,
  payload: {
    broker: 'hao-MS-7B85:9092',
    clientId: 'my-app',
    correlationId: 3,
    size: 59,
    createdAt: 1611581796522,
    sentAt: 1611581796523,
    pendingDuration: 1,
    duration: 0,
    apiName: 'Produce',
    apiKey: 0,
    apiVersion: 7
  }
}
InstrumentationEvent {
  id: 6,
  type: 'producer.network.request',
  timestamp: 1611581796524,
  payload: {
    broker: 'hao-MS-7B85:9092',
    clientId: 'my-app',
    correlationId: 4,
    size: 59,
    createdAt: 1611581796523,
    sentAt: 1611581796523,
    pendingDuration: 0,
    duration: 1,
    apiName: 'Produce',
    apiKey: 0,
    apiVersion: 7
  }
}
InstrumentationEvent {
  id: 7,
  type: 'producer.network.request',
  timestamp: 1611581796524,
  payload: {
    broker: 'hao-MS-7B85:9092',
    clientId: 'my-app',
    correlationId: 5,
    size: 59,
    createdAt: 1611581796523,
    sentAt: 1611581796523,
    pendingDuration: 0,
    duration: 1,
    apiName: 'Produce',
    apiKey: 0,
    apiVersion: 7
  }
}
InstrumentationEvent {
  id: 8,
  type: 'producer.network.request',
  timestamp: 1611581796524,
  payload: {
    broker: 'hao-MS-7B85:9092',
    clientId: 'my-app',
    correlationId: 6,
    size: 59,
    createdAt: 1611581796523,
    sentAt: 1611581796523,
    pendingDuration: 0,
    duration: 1,
    apiName: 'Produce',
    apiKey: 0,
    apiVersion: 7
  }
}
[
  [
    {
      topicName: 'first_topic',
      partition: 2,
      errorCode: 0,
      baseOffset: '14',
      logAppendTime: '-1',
      logStartOffset: '0'
    }
  ],
  [
    {
      topicName: 'first_topic',
      partition: 0,
      errorCode: 0,
      baseOffset: '14',
      logAppendTime: '-1',
      logStartOffset: '0'
    }
  ],
  [
    {
      topicName: 'first_topic',
      partition: 1,
      errorCode: 0,
      baseOffset: '16',
      logAppendTime: '-1',
      logStartOffset: '0'
    }
  ],
  [
    {
      topicName: 'first_topic',
      partition: 0,
      errorCode: 0,
      baseOffset: '14',
      logAppendTime: '1611581796522',
      logStartOffset: '0'
    }
  ],
  [
    {
      topicName: 'first_topic',
      partition: 2,
      errorCode: 0,
      baseOffset: '14',
      logAppendTime: '1611581796521',
      logStartOffset: '0'
    }
  ]
]
Cost: 8ms
```

consumer
```
{ partition: 2, offset: '14', value: 'Hello! 1611581796516-0' }
{ partition: 0, offset: '14', value: 'Hello! 1611581796516-1' }
{ partition: 1, offset: '16', value: 'Hello! 1611581796516-2' }
```

**Environment:**
 - OS: Ubuntu 18.04
 - KafkaJS version 1.15.0
 - Kafka version kafka_2.13-2.7.0
 - NodeJS version v12.14.1

**Additional context**

t-d-d:
Read and update sequence numbers immediately before the produce request.
Fixes #1005 Maybe #598 

Putting this up for review and input but I'm still unsure as to the failure modes.  I'm wondering whether we should decrement the sequence numbers if the broker.produce() call fails.


t-d-d:
More [info](https://issues.apache.org/jira/browse/KAFKA-5494) on dealing with failures with inflight request > 1.  Although I don't think we need this full implementation, as I don't think we need to guarantee ordering for concurrent requests (or we at least state that we don't.)

I think this PR as-is is an improvement to the current behaviour.  But I think there are questions around concurrent invocations of produce() and inflight requests in the idempotent mode, especially dealing with errors.

src/producer/__tests__/concurrentIdempotent.spec.js
```diff
@@ -0,0 +1,64 @@
...
+        .fill()
+        .map(() => {
+          const value = secureRandom()
+          return { key: `key-${value}`, value: `value-${value}` }
+        })
+
+      await Promise.all(
+        messages.map(m => producer.send({ acks: -1, topic: topicName, messages: [m] }))
+      )
+      await Promise.all(
```
Nevon, line 55: I guess this second one doesn't really matter. It should already have failed on the previous line if the bug wasn't fixed.

t-d-d: See re-worked test cases

t-d-d:
@Nevon I have added code that reverts the sequence numbers in case of a failure. Now at least it will recover in the normal case of sequential produce() calls (for a topic-partition.)

For errors when there are concurrent produce() calls ongoing (for a topic-partition) the idempotent producer will die with a  UNKNOWN_PRODUCER_ID or OUT_OF_ORDER_SEQUENCE_NUMBER error on any error.  This could be made to work if the sequence number stuff was moved inside the mutex lock that limits maxInFlightRequests.  It would be pretty simple to add another lock (per broker, or maybe even per partition?) in sendMessages.js 

Test cases added that capture the current behaviour.

I just noticed https://github.com/tulios/kafkajs/blob/9cf3dfbe4c234920db171fadb113fc5f8699c14e/src/producer/sendMessages.js#L127 should probably be an allSettled().  Would cause problems if not all the promises are settled before the retry.

src/producer/__tests__/idempotentProduceMessages.spec.js
```diff
@@ -0,0 +1,222 @@
...
+    await consumer.run({ eachMessage: async message => messagesConsumed.push(message) })
+
+    await waitForMessages(messagesConsumed, { number: messages.length })
+
+    messagesConsumed.forEach(({ message: { value } }, i) =>
+      expect(value.toString()).toEqual(`${i}`)
+    )
+  })
+
+  it('concurrent produce() calls > all messages are written to the partition once', async () => {
```
Nevon, line 130: Don't worry about changing this, but just for future reference, you can just wrap these test cases in a describe block to group the tests together, instead of repeating the prefix in each test case:

```js
describe('concurrent produce() calls', () => {
  test('all messages are written to the partition at once')
})
```

src/producer/__tests__/idempotentProduceMessages.spec.js
```diff
@@ -0,0 +1,222 @@
...
+    )
+  })
+
+  it('concurrent produce() calls > where produce() throws a retriable error on the first call, all subsequent calls throw UNKNOWN_PRODUCER_ID', async () => {
+    for (const nodeId of [0, 1, 2]) {
+      const broker = await cluster.findBroker({ nodeId })
+
+      const brokerProduce = jest.spyOn(broker, 'produce')
+      brokerProduce.mockImplementationOnce(async () => {
+        await sleep(100)
```
Nevon, line 152: Do these sleeps make any meaningful difference? Our tests are already taking a bit, so if we don't need sleeps to test some timing thing, let's not add them.

src/utils/promiseAllSettled.js
```diff
@@ -0,0 +1,18 @@
+/**
+ *
+ * @param { [Promise<T>] } promises
+ * @returns {Promise<[{ status: "fulfilled", value: T} | { status: "rejected", reason: Error}]> }
+ * @template T
+ */
+
+function allSettled(promises) {
```
Nevon, line 8: Nice. We should actually bump the requirement on Node to 12 at this point, but that's outside the scope of this PR.

Nevon (approved):
Just one minor note about whether or not we can remove the sleeps from the tests, but overall it looks good to me. Thanks for your patience!

src/producer/__tests__/idempotentProduceMessages.spec.js
t-d-d: Yes, they were meaningful.  The intent was to make sure the other concurrent produce requests got initiated before the mocked one threw the error. But agree that's far from ideal way to do this.

I have now changed it to actually await on the number of times the produce request is called before throwing.

pi2sqr:
Hi,

I have a question. 
I'm trying to do performance test for default topic: -topic kafka-test --replication-factor 2 --partitions 1.
No other settings different from default.

I've Producer and Consumer written in Java. 
Producer tries to do async send 10 messages in loop one by one, each with 10kB volume record:

   for (int i=0; i<numberOfRecords; i++) {
            start_time = System.currentTimeMillis();
            ProducerRecord<String, String> record = new ProducerRecord<>(kafkaProp.getProperty("kafka.topic"), "key " + i, start_time + message);
            Future<RecordMetadata> sendFuture = producer.send(record);
            if (sendFuture.isDone()) {
                try {
                    sendFuture.get();
                } catch (InterruptedException var7) {
                    Thread.currentThread().interrupt();
                    throw new KafkaException("Interrupted", var7);
                } catch (ExecutionException var8) {
                    throw new KafkaException("Send failed", var8.getCause());
                }
            }
        }
        producer.flush();
        producer.close();


But even for one Producer I'm getting: 
"_...org.apache.kafka.common.errors.OutOfOrderSequenceException: The broker received an out of order sequence number. for topic-partition kafka-test-0 with producerId 24053, epoch 851, and sequence number 1
13:19:36:569 [kafka-producer-network-thread | producer-1] INFO org.apache.kafka.clients.producer.internals.TransactionManager - [Producer clientId=producer-1] ProducerId set to 24053 with epoch 852_"

What is strange for me, that the first 2050 messages are sent in less than a second.
Then it take some time to send next messages, 20ms up, but still 2050 x 10270B is not a size of topic's buffer.
The more Producers I have for one partition the more errors I'm getting.
From Consumer site I'm counting how many messages I'm not receiving and it's from 200 over 2000 with 1 Producer. For and more Producers its over a 60% of messages that are not received (maybe send correctly).

Can you advise?

PR #1054, opened 2021-03-24
Docusaurus v2 migration

j-a-h-i-r:
docusaurus v1 website move to v2

Almost all functionalities are working except,
1. Version dropdown in navbar does not navigate correctly.
2. The logo in footer should be smaller.

I will work on fixing these issues.

I tried to keep the styling as close to the v1 website as possible.


tulios:
Great work @j-a-h-i-r 🥇

j-a-h-i-r:
In v2, docusaurus has [home page docs](https://docusaurus.io/docs/docs-introduction#home-page-docs) where a specific doc page can be made available on `/docs`. The docusaurus website uses this. For example, the `Installation` doc is available at https://docusaurus.io/docs/ instead of https://docusaurus.io/docs/installation. 

In KafkaJS, the default `Getting Started` page is available https://kafka.js.org/docs/getting-started. Should we make it available at https://kafka.js.org/docs following the v2 conventions? It would also fix the broken versioned doc links in nav bar too.

tulios:
> Should we make it available at https://kafka.js.org/docs following the v2 conventions?

j-a-h-i-r:
I couldn't get the GitHub star button to render properly. But I think it is due to docusaurus v2 moving to SPA. [This issue discusses github start not rendering](https://github.com/ntkme/github-buttons/issues/66#issuecomment-622192370). I could not find a workaround to this problem.

As an alternative, I used shields.io (https://img.shields.io/github/stars/tulios/kafkajs) and it works without any issues.

Aside from this, other issues mentioned in the first comment are resolved.

PR #1071, opened 2021-04-03
Changed rebalance event to fire when the rebalance has finished

benbirt97:
**Description**
In a previous PR I added in a REBALANCE event which triggered whenever a rebalance started however in hindsight it would actually be more useful for applications to have it trigger when a rebalance finishes. I have also added the memberAssignment into the event now that we know what the new assignment is.

**Reason**
When we reset metrics on the previous event, it is still possible that we get messages from the old partitions straight after the event but before the rebalancing has occurred. This means we would need to put in an arbitrary timeout before resetting our metrics so that we can be sure that our metrics won't have any previous partitions. In hindsight it would be more useful to applications to know when the rebalancing has ended rather than when the rebalancing begins.

**Changes**
- Changed the Event to be fire after we have rejoined the consumer group
- Changed the REBALANCING Event to instead be a REBALANCE Event
- Updated documentation
- Updated rebalance test
- Fixed Husky depreciation warning


Nevon:
With this implementation, how is this different from `GROUP_JOIN`? That event already signifies that we have join and synced - i.e. completed a rebalance. That event also already contains the member assignments.

benbirt97:
I guess theres a few things...

1. It will only trigger on a rebalance. It won't trigger on any group joins which weren't caused by a rebalance (Initial Joins, Stopping and Starting the consumer or any UNKNOWN_MEMBER_ID errors). Allows you to put specific rebalance logic in (logging about rebalance, rebalance metrics etc.) without worrying about working out what is a join or what is a rebalance. 
2. It separates them out within the underlying KafkaJS code. In the future there might be some specific scenarios where it is easier to rejoin to the group rather than considering it a rebalance. For example the NOT_COORDINATOR_FOR_GROUP error you mentioned in the previous PR. It means you can treat this as a rejoin without triggering any application side rebalance logic.
3. It is easier for anybody using KafkaJS to see there is a rebalance event that contains the memberAssignment and use that, rather than having to work out that GROUP_JOIN is the one they should be using by looking at the underlying KafkaJS code and then adding additional logic to ignore certain events.

PR #1079, opened 2021-04-17
Fix REBALANCING event test

t-d-d:
Test case code assumed the consumer was a node eventEmitter.  In fact, the call to consumer.removeAllListeners() is an error as no such function exists.  Error was swallowed, but the intent of the call not being met means that the test case is flakey.

Also, consumers weren't getting disconnected.


Nevon:
Nice catch!

PR #1082, opened 2021-04-19
Fix typings for LoggerEntryContent timestamp

bdchauvette:
The previous typings had `LoggerEntryContent['timestamp']` as a `Date`, but it's actually a `string`.

See: https://github.com/tulios/kafkajs/blob/a00d664fb2a25330c5b5e5e1ffe6d80c26d5c671/src/loggers/index.js#L22


Nevon:
Nice catch, and thanks for the contribution!

PR #1084, opened 2021-04-22
Adding TypeScript examples for Producing and Consuming

### LINKED ISSUE #1080: Document Typescript usage
Our search analytics show that "Typescript" is a common search query in our documentation that yields no results. Since we do support Typescript, we should make sure that at the very least there is some search result that tells you that KafkaJS works with Typescript and shows a minimal example.

Under the "Examples" section, I'd like to have a page called "Typescript" that shows a small example of initializing the client in Typescript, just to assure people that we provide types.

panzerama:
Closes #1080 
My team at Nordstrom uses kafkajs extensively. I've adapted samples from one our apps to serve as examples of TypeScript usage. This is my first contribution to kafkajs, so please let me know if I've missed or need to add anything to my PR.

It seems like adding [multi-language example support](https://docusaurus.io/docs/markdown-features/code-blocks#multi-language-support-code-blocks) would be more work than is merited in this case, but it's worth calling out that it is possible to compress the display of these examples.

### DIFF
--- a/docs/ConsumerExample.md
+++ b/docs/ConsumerExample.md
@@ -64,6 +64,83 @@ signalTraps.map(type => {
 })
 ```
 
+## <a name="typescript-consumer-example"></a> TypeScript Example
+
+A similar example in TypeScript
+
+```typescript
+import { Consumer, ConsumerSubscribeTopic, EachBatchPayload, Kafka, EachMessagePayload } from 'kafkajs'
+
+// ExampleMessageProcessor is a placeholder for the logic applied to each 
+// message or message batch and is not provided by kafkajs
+import ExampleMessageProcessor from './ExampleMessageProcessor'
+
+export default class ExampleConsumer {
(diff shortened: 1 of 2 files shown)


docs/ConsumerExample.md
```diff
@@ -64,6 +64,80 @@ signalTraps.map(type => {
...
 ```
 
+## <a name="typescript-consumer-example"></a> TypeScript Example
+
+A similar example in TypeScript
+
+```typescript
+import { Consumer, ConsumerSubscribeTopic, EachBatchPayload, Kafka, EachMessagePayload } from 'kafkajs'
+
+// ExampleMessageProcessor is a placeholder for the logic applied to each
```
Nevon, line 74: I think it's better if we remove this, to reduce the risk of confusion. Let's just have the eachMessage handler console.log the message value, like we do in the JS example.

panzerama: Updated.

PR #1099, opened 2021-05-13
Re-enable skipped tests

Nevon:
I guess I left an `.only` in there at some point. Re-enabling the tests showed me I had one failing test, but I would say it was failing for good reason. The `toInt` method is meant to return a int32, but it was expecting it to return a number larger than the range of a signed int32. Maybe at some point I thought it should return a uint32 or something. Fixed the tests to match what the code is actually supposed to do.


JaapRood (approved):
LGTM!

PR #1101, opened 2021-05-13
Allow injecting authentication mechanisms

### LINKED ISSUE #840: Allow for injecting authentication mechanisms
Currently there is no way for a developer to add support for SASL GSSAPI (Kerberos) without bundling that into KafkaJS itself. We would likely not want to do that, as that particular case requires a native module.

I should note that while I think this change makes sense, and [I have definitely seen](https://stackoverflow.com/questions/63315956/how-to-connect-to-kafka-cluster-using-kerberos-sasl-in-node-js) people asking for Kerberos authentication (#609 & #213), I don't have this use-case myself, so unless someone is willing to sponsor the work, consider this more of a roadmap ticket.

Similar to how we allow users to add compression codecs, we should expose the list of supported mechanisms and allow users to add their implementations:

```js
const {  SaslMechanisms, SaslAuthenticators } = require('kafkajs')

// Adding support for new mechanism
const GssapiAuthenticator = require('kafkajs-gssapi')
SaslAuthenticators["GSSAPI"] = GssapiAuthenticator

// Overriding built-in authenticator for supported mechanism
const OauthbearerAuthenticator = require('custom-kafkajs-oauthbearer')
SaslAuthenticators[SaslMechanisms.OAUTHBEARER] = OauthbearerAuthenticator
```

One challenge I see is that the interface for the authenticator is not entirely fit for external usage at the moment. Essentially, an authenticator looks something like this today:

```ts
interface AuthenticatorInterface {
    authenticate(): Promise<void>;
}

interface AuthenticatorConstructor {
  new (connection: Connection, logger: Logger, saslAuthenticate: SaslAuthenticationFunction): AuthenticatorInterface;
}

interface Encoder {
    buffer: Buffer
}

interface Request {
    encode(): Promise<Encoder>
}

interface Response {
    decode(data: Buffer): Promise<Buffer>
    parse(data: Buffer): Promise<any> 
}

interface SaslAuthenticationFunctionParams {
    request: Request,
    response: Response,
    authExpectResponse?: boolean
}

type SaslAuthenticationFunction = (params: SaslAuthenticationFunctionParams) => Promise<void>;
```

`AuthenticatorConstructor` is where we expose KafkaJS internals. Specifically:

* `Logger` - This one is fine. This is already available and has a public interface.
* `Connection` - This is worse. This is very much not a public interface.

From what I can see from existing authenticators, we are really only providing the Connection to them because they need a subset of the following information:

* Broker `host` and `port`. From what I can see this is only for logging.
* `sasl` username and password. This is obviously needed for the authenticator.

The authenticator does not actually need the connection itself, because the only request it has to make to the broker is done via `SaslAuthenticationFunction` where it's safely encapsulated. We could just provide the `sasl` and `broker` information.

I'm also not entirely sure about the parameter and return types of `Request` and `Response`. I think I saw different authenticators return `Promise<Encoder|Buffer>` for `Request.encode()` and `Promise<Encoder|Buffer|boolean>` for `Response.decode` and `Response.parse`, so there's probably some cleaning up to do there. If we make it public, it should only accept and return buffers, rather than `Encoder`s.

Nevon:
This allows users to inject their own custom authentication mechanisms, similar to what's done for compression codecs.

```js
const { AuthenticationMechanisms } = require('kafkajs')
const { Mechanism, Type } = require('customauth')

AuthenticationMechanisms[Type] = () => Mechanism
```

The existing authentication mechanisms use the same interface, which meant I had to rework them some. Primarily, I moved them from being classes to functions. This isn't strictly necessary, but I find it easier to have people implement a function than to have them extend a class or similar - especially if they are using different compile-to-js languages.

The other thing was that I changed the interface to not expose any internal classes. That means moving from Encoders/Decoders to just buffers, and to not pass in the Connection anymore, and instead provide the `sasl` options and the broker host + port.

One thing that is missing is documentation on Typescript usage. After getting some advice from the internet, I opted for a closed type for the `AuthenticationMechanisms`, so someone implementing their own authentication mechanism in Typescript should use declaration merging to extend that type with their own type. But I haven't tried this out yet myself, which is why I haven't documented it yet.

Fixes #840

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -171,6 +171,21 @@ You can also programmatically retrieve the `aws:userid` for currently available 
 A complete breakdown can be found in the IAM User Guide's
 [Reference on Policy Variables](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html#policy-vars-infotouse).
 
+### Custom Authentication Mechanisms
+
+If an authentication mechanism is not supported out of the box in KafkaJS, a custom authentication
+mechanism can be introduced as a plugin:
+
+```js
+const CustomAuthenticationMechanism = require('custom-authentication-mechanism')
+const { AuthenticationMechanisms } = require('kafkajs')
+
+AuthenticationMechanisms['CUSTOM_AUTHENTICATION'] = CustomAuthenticationMechanism
+```
+
+See [Custom Authentication Mechanisms](CustomAuthenticationMechanism.md) for more information on how to implement your own
+authentication mechanism.
+
 ### Use Encrypted Protocols
 
 It is **highly recommended** that you use SSL for encryption when using `PLAIN` or `AWS`,
--- a/docs/CustomAuthenticationMechanism.md
+++ b/docs/CustomAuthenticationMechanism.md
@@ -1,0 +1,150 @@
+---
+id: custom-authentication-mechanism
+title: Custom Authentication Mechanisms
+---
+
+To use an authentication mechanism that is not supported out of the box by KafkaJS,
+custom authentication mechanisms can be introduced:
+
+```js
+const CustomAuthenticationMechanism = require('custom-authentication-mechanism')
+const { AuthenticationMechanisms } = require('kafkajs')
+
+AuthenticationMechanisms['CUSTOM_AUTHENTICATION'] = CustomAuthenticationMechanism
+```
+
+`CUSTOM_AUTHENTICATION` needs to match the SASL mechanism configured in the `sasl.enabled.mechanisms`
+property in `server.properties`. See the Kafka documentation for information on how to
+configure your brokers.
+
+## Writing a custom authentication mechanism
+
+A custom authentication mechanism needs to fulfill the following interface:
+
+```ts
+type AuthenticationMechanismCreator = (params: {
+  sasl: Record<string, any>,
+  connection: { host: string, port: number },
+  logger: Logger,
+  saslAuthenticate: (handlers: { request: SaslAuthenticationRequest, response?: SaslAuthenticationResponse }) => Promise<void>
+}) => AuthenticationMechanism
+
+type AuthenticationMechanism = {
+  authenticate(): Promise<void>
+}
+
+type SaslAuthenticationRequest = (sasl: Record<string, any>) => {
+  encode: () => Buffer
+}
+
+type SaslAuthenticationResponse = () => {
+  decode: (rawResponse: Buffer) => Buffer
+  parse: (data: Buffer) => any
+}
+```
+
+* `sasl` - Whatever is passed into the client constructor as `sasl`
+* `connection` - The host and port that the client will authenticate against
+* `logger` - A logger instance namespaced to the authentication mechanism
+* `saslAuthenticate` - an async function to make [`SaslAuthenticate](https://kafka.apache.org/protocol.html#The_Messages_SaslAuthenticate)
+requests towards the broker.
+
+### Example
+
+In this example we will create a custom authentication mechanism called `SIMON`. The general
+flow will be:
+
+1. Read `sasl.simonSays` from the client configuration
+2. Send a [`SaslAuthenticate`](https://kafka.apache.org/protocol.html#The_Messages_SaslAuthenticate)
+request with the value of `sasl.simonSays` as `auth_bytes`.
+3. Read the response from the broker and validate that it returned the same value.
+
+```js
+function SimonAuthenticator({ sasl, connection, logger, saslAuthenticate }) {
+  const INT32_SIZE = 4
+
+  const request = ({
+    /**
+     * Encodes the value for `auth_bytes` in SaslAuthenticate request
+     * @see https://kafka.apache.org/protocol.html#The_Messages_SaslAuthenticate
+     * 
+     * In this example, we are just sending `sasl.simonSays` as a string,
+     * with the length of the string in bytes prepended as an int32
+     **/
+    encode: async () => {
+      const byteLength = Buffer.byteLength(sasl.simonSays, 'utf8')
+      const buf = Buffer.alloc(INT32_SIZE + byteLength)
+      buf.writeUInt32BE(byteLength, 0)
+      buf.write(sasl.simonSays, INT32_SIZE, byteLength, 'utf8')
+      return buf
+    },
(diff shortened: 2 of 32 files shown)


Nevon:
Tested out and documented the typescript usage. The documentation has been updated accordingly.

jvincilione:
It seems this change has not had any updates and has not been merged in a while. @Nevon will this end up getting added? I'm going to be using Kerberos, so I could definitely use this...

Nevon:
I think it should definitely end up getting added, as there are already several use-cases we know of where this is necessary (aws-iam-msk & kerberos), but since this is a public API that people will be building on top of, we have to make sure that the interface works and makes sense.

I tried implementing aws-iam-msk authentication on top of this interface to test it out, and ran into some rough edges when using Typescript. It was a few weeks ago now, so the details a fussy, but from what I recall the issues were around making sure that if a user adds a new authentication mechanism, the options they pass in to `sasl` should still be typechecked according to the types of the custom authentication mechanism. I expected this to be doable using module augmentation of the `AuthenticationMechanisms` type, but I ran into some issues there, as seen by this example usage:

```ts
import { Kafka, AuthenticationMechanisms, AuthenticationMechanism } from 'kafkajs'
import { MskIamMechanism } from './MskIamMechanism'

// Augmenting the SASLMechanismsOptionsMap with the new mechanism
declare module 'kafkajs' {
  interface SASLMechanismOptionsMap {
    AWS_MSK_IAM: {
      credentialProvider: CredentialProvider
    }
  }
}

/**
 * Here's the ugly part where in order to be able to actually add the custom authentication mechanism
 * to the mechanisms, I have to cast AuthenticationMechanisms to my own custom type, which first
 * requires casting to unknown. This has to be done by the user, which sucks, but it's the only way I've found
 * that allows `sasl` to use the type provided by the custom mechanism.
 */
type ExtraAuthenticationMechanisms = AuthenticationMechanism & {
  AWS_MSK_IAM: () => typeof MskIamMechanism
}
;(AuthenticationMechanisms as unknown as ExtraAuthenticationMechanisms)['AWS_MSK_IAM'] = () =>
  MskIamMechanism

const kafka = new Kafka({
  brokers: [
    'b-1.fake-test-broker.abc123.kafka.eu-north-1.amazonaws.com:9098',
  ],
  ssl: true,
  sasl: {
    mechanism: 'AWS_MSK_IAM',
    credentialProvider: defaultProvider(),
  },
  connectionTimeout: 5000,
})
```

My suggestion would be to attempt to implement Kerberos authentication on top of this interface, and then come back with concrete feedback on what worked well, what didn't, and what the user experience will be like. Maybe we can make some improvements to make it less of a hassle for the user.

The current implementation should work just fine for the built-in authentication mechanisms, so as a pure refactor of the existing functionality, I consider it good enough. But before we cut a stable release containing this, I'd like to improve the interface. Maybe @tulios could help out with a review so that we can get this merged in the current state, and then refactor the plugin interface once we have some feedback to work from.

blacha:
I am wanting to run a few nodejs lambda's against a MSK Kafka with IAM auth enabled, so I am very interested in this pull request.

I am wondering if a slightly different approach would make the new API slightly easier to type.

Would it be possible to instead of using a JSON config, passing a instantiated `Mechanisim` class into the client

```typescript
const kafka = new Kafka({
  brokers: [
    'b-1.fake-test-broker.abc123.kafka.eu-north-1.amazonaws.com:9098',
  ],
  ssl: true,
  sasl: new AwsMSKIamMechanisim({ credentials } ),
  connectionTimeout: 5000,
})
```

This would simplify the type definitions as you could do:

```
export type sasl = SASLOptions | Mechanisim
```

This concept is similar to how the aws elstaic search connector works, where you are allowed to override the `Transport` and  `Connection` classes in the constructor of the client ref: https://github.com/compwright/aws-elasticsearch-connector

ultrarunner:
Hello, I just have a question about this PR regarding the ability to implement a custom auth mechanism to set the username / password based on an async call. Our kafka servers have ssl in place but we need to retrieve the username / password from a vault since they cannot be stored in the stash repo, very similar to the way the "oauthBearer" option is implemented. Below would be a very crude idea (interface and type taken from the index.d.ts file). Would the above PR allow for this? Thank you.

`export interface PlainProviderResponse {
   username: string,
   password: string
}`

`type SASLMechanismOptionsMap = {'plainprovider': { plainProvider: () => Promise<PlainbearerProviderResponse> }`

damaral-Fuze:
Hi, any chance of this PR being merged?

s4bdufabricio:
@Nevon I would like to use this feature ! :)

ludov04:
This would be immensely useful

amitruls1:
Hi, @Nevon When can we expect this to be released? Thanks.

leoddias:
@tulios @JaapRood @Nevon  Does anybody know when it will be released?
Any alternatives so far?
Thanks

Nevon:
It'll happen when I have time to work on it. These things don't happen by magic, but by someone spending the hours of work to make it happen, and right now my priority is #1258. 🙂

I like @blacha's suggestion and will likely refactor this in that direction once I have time to come back to this.

Nevon:
I'm going to close this PR since it's not particularly useful at the moment. The branch will remain, in case someone wants to take a crack at reworking it.

I spent some time last week trying to refactor it with a better interface, and also make the necessary adjustments from the changes to the Connection that are in master, but didn't finish. One thing to note about blacha's suggestion is that the authentication mechanism needs to expose a name that we can use in the protocol negotiation, so the `sasl` interface would have to be something like `{ mechanism: string; authenticate: (stuff: { saslAuthenticate, aFewOtherThingsICantRemember: ...unknown }) => Promise<void> }` (as a union with the existing interface, for backwards compatibility).

ro-ka:
Please keep the branch available, we’re using that branch already successfully in our project.  😬  Hope there will be a final solution someday in the main branch. 🤞  Thanks for your work here!

Nevon:
That seems incredibly dangerous. I would suggest that you fork the repo so that you have your own copy of the branch, rather than rely on me not clicking that "Delete branch" button.

antonwintergerst:
FYI, @markgaylard has completed the work to allow custom authentication mechanisms and has raised a PR here: https://github.com/tulios/kafkajs/pull/1372

PR #1105, opened 2021-05-19
Add function to override metadata brokers

woodcockjosh:
If we are connecting to a kafka cluster via ssh tunnel. We cannot connect to the brokers that the cluster meta data provided. So this provides a way for us to override the brokers provided by the metadata so we can connect to the ip and port we are tunneling through.

This can be useful in any case where the client cannot connect directly to the host name and port that is set in the metadata. Another use case would be when using port forwarding via the kubernates cli.

### DIFF
--- a/src/cluster/brokerPool.js
+++ b/src/cluster/brokerPool.js
@@ -20,6 +20,7 @@ module.exports = class BrokerPool {
    * @param {number} [options.authenticationTimeout]
    * @param {number} [options.reauthenticationThreshold]
    * @param {number} [options.metadataMaxAge]
+   * @param {function} [options.getMetadataBrokers]
    */
   constructor({
     connectionBuilder,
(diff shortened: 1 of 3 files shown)


Nevon:
I don't think this is the way to do this, because it assumes that the user has the ability to generate valid metadata, meaning they know all the node ids and what host/port corresponds to them. If a broker goes down, metadata won't accurately reflect this.

A better solution might be to do what we did when going through a websocket proxy, which is to provide a [custom socket factory](https://kafka.js.org/docs/configuration#custom-socket-factory) that maps from the host/port returned in the metadata to the proxy host/port.

For example:

```js
const net = require('net')
const tls = require('tls')

const ProxiedSocketFactory = (mapping) => ({ host, port, ssl, onConnect }) => {
  const { host: proxyHost, port: proxyPort } = mapping[`${host}:${port}`]

  const socket = ssl
    ? tls.connect(
        Object.assign({ host: proxyHost, port: proxyPort }, ssl),
        onConnect
      )
    : net.connect(
        { host: proxyHost, port: proxyPort },
        onConnect
      )

  return socket
}

const kafka = new Kafka({
  clientId: 'my-app',
  brokers: ['kafka1:9092', 'kafka2:9092'],
  socketFactory: ProxiedSocketFactory({
    'broker1:9092': { host: 'proxy', port: 9092 },
    'broker2:9092': { host: 'proxy', port: 9093 },
    'broker3:9092': { host: 'proxy', port: 9094 }
  }),
})
```

woodcockjosh:
Interesting. I think I'll stick with my fork. It would be cool if kafkajs
could enable connection over ssh tunnel and kubernetes port forwarding out
of the box.

Nevon:
I'm gonna close this PR as per my previous suggestion.

Issue #1106, opened 2021-05-21
sen't


woodcockjosh:
**Describe the bug**
      "This most likely occurs because of a request being malformed by the client library or the message was sen't to an incompatible broker. See the broker logs for more details",

sen't doesn't need a "'"

https://github.com/tulios/kafkajs/blob/d8fd93e7ce8e4675e3bb9b13d7a1e55a1e0f6bbf/src/protocol/error.js#L269

PR #1111, opened 2021-05-24
spell check

prithvijit-dasgupta:
Fixes https://github.com/tulios/kafkajs/issues/1106


Nevon:
Thank you for the fix!

PR #1118, opened 2021-06-09
Rename acl -> acls in DescribeAclResource

zachary-povey:
At the moment the type for `DescribeAclResource` has the property `acl` , however as detailed [in the docs for describeAcls](https://kafka.js.org/docs/admin#describe-acl) the data structure that comes back from calling `describeAcls` has the key `acls`.

This PR updates the type to reflect this.


Nevon:
Nice catch! Thanks!

PR #1132, opened 2021-06-23
Feature/duplicate header keys

### LINKED ISSUE #1121: Handling multiple headers with the same key
Kafka allows multiple headers with the same key.
>Duplicate headers with the same key must be supported. [(KIP-82)](https://cwiki.apache.org/confluence/display/KAFKA/KIP-82+-+Add+Record+Headers#KIP82AddRecordHeaders-Keyduplicationandordering)

It seems that KafkaJS does not handle messages with such headers. The consumer ignores headers after the first appearance with the same key. Sending such messages is also impossible because the `Message.headers` is defined as a JavaScript object, which does not allow duplicate keys.

`Message.headers` should be able to hold arrays of `Buffer | string`
```ts
export interface IHeaders {
  [key: string]: Buffer | string | (Buffer | string)[] | undefined
}
```

**Additional context**
Applications may encounters an exception if they don't expect arrays.

jeongki-kim:
This enables KafkaJS to handle multiple headers with the same key. `IHeader` will hold an array of `String | Buffer` in order to store values of such headers.

```json
{
  "a": "b",
  "c": ["d", "e"]
}
```

Fixes #1121


Nevon:
Thank you! This is an excellent contribution, and the implementation looks great.

In terms of backwards compatibility, this does now mean that anyone using headers will need to update their code to potentially handle getting an array value. Technically this is a breaking change, which we try to avoid whenever possible. The alternative would be to expose an additional headers field with this new signature, and then deprecate the old one.

However, I just noticed that Node 10 is no longer supported since a few weeks back, so we could drop support for it and move to Node 12, which would be a breaking change regardless and warrant a major version bump (it would also solve the issue of your breaking tests, since you're using Array.prototype.flatMap, which only became supported in Node 11).

Would you mind updating the engine in the package.json and changing the pipeline to use Node 12 instead?

PR #1133, opened 2021-06-28
Don't set default offset if autoCommit is disabled

chosh31:
`this.autoCommit` is disabled but committed when the offset in consumer.seek goes out of range.
(e.g. offset expired)

### DIFF
--- a/src/consumer/consumerGroup.js
+++ b/src/consumer/consumerGroup.js
@@ -633,6 +633,13 @@ module.exports = class ConsumerGroup {
         memberId: this.memberId,
       })
       delete preferredReadReplicas[e.partition]
+    } else if (!this.autoCommit) {
+      this.logger.error('Offset out of range', {
+        topic: e.topic,
+        partition: e.partition,
+        groupId: this.groupId,
+        memberId: this.memberId,
+      })
     } else {
       this.logger.error('Offset out of range, resetting to default offset', {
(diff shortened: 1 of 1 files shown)


Nevon:
I was thinking about how to write a test for this to ensure that we don't introduce a regression in the future, but I can't think of what the expected behavior should be. I imagine a test similar to this one: https://github.com/tulios/kafkajs/blob/f8366f30d965214701393b31e1cc6eddae1fe9db/src/consumer/__tests__/seek.spec.js#L144-L166

But if the consumer doesn't reset the offset and it swallows the offset out of range error, the only thing that will happen is that it will retry fetching the same offset - which is still going to be out of range.

I see two alternatives, but I don't know which one is preferable. It would be a good idea to look into how the Java client behaves in this scenario:

1. Re-throw the error so that we get a crash that the userland code can act on.
2. The offset manager already knows if autoCommit is enabled or not, so if it isn't, `setDefaultOffset` could just set the resolved offset to be the default offset, instead of committing it. That way the high-level behavior is the same, whether autocommit is enabled or not, but we don't end up committing offsets.

chosh31:
> But if the consumer doesn't reset the offset and it swallows the offset out of range error, the only thing that will happen is that it will retry fetching the same offset - which is still going to be out of range.

@Nevon I agree with you. This pr is closed

PR #1136, opened 2021-06-29
Offset manager setDefaultOffset check if autoCommit is enabled

chosh31:
> 2. The offset manager already knows if autoCommit is enabled or not, so if it isn't, setDefaultOffset could just set the resolved offset to be the default offset, instead of committing it. That way the high-level behavior is the same, whether autocommit is enabled or not, but we don't end up committing offsets.

I agree with the second alternative.
So I modified the code. Is this similar to the fix you were thinking of?
@Nevon 

https://github.com/tulios/kafkajs/pull/1133#issuecomment-870280517

### DIFF
--- a/src/consumer/__tests__/seek.spec.js
+++ b/src/consumer/__tests__/seek.spec.js
@@ -42,6 +42,16 @@ describe('Consumer', () => {
   })
 
   describe('when seek offset', () => {
+    let admin
+
+    beforeEach(() => {
+      admin = createAdmin({ logger: newLogger(), cluster })
+    })
+
+    afterEach(async () => {
+      admin && (await admin.disconnect())
+    })
+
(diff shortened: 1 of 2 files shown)


Nevon:
Could you write a test case that proves that this actually works? I would expect it to be identical to this test: https://github.com/tulios/kafkajs/blob/f8366f30d965214701393b31e1cc6eddae1fe9db/src/consumer/__tests__/seek.spec.js#L144-L166 except that the consumer is configured to not autoCommit, and if we [fetch consumer group offsets](https://kafka.js.org/docs/admin#a-name-fetch-offsets-a-fetch-consumer-group-offsets) afterwards, there should be no committed offsets for the group.

Nevon:
Awesome!

PR #1172, opened 2021-08-14
Concurrent idempotent producer error handling fixes

t-d-d:
This is on top of https://github.com/tulios/kafkajs/pull/1050 .  Added:

- a lock per broker so that the broker request and sequence number handling (including reverting on an error) is serialised
- used sharedPromise on a few cluster functions to stop multiple concurrent calls


src/cluster/index.js
```diff
@@ -106,7 +144,7 @@ module.exports = class Cluster {
    * @returns {Promise<void>}
    */
   async connect() {
-    await this.brokerPool.connect()
+    return await this[PRIVATE.REFRESHMETADATA]()
```
aikoven, line 147:
```suggestion
    return await this[PRIVATE.CONNECT]()
```

PR #1173, opened 2021-08-24
Minor grammar update in docs - Consuming.md

byrne-greg:
Grammar error around the use of 'an'

_"for building an processing reset tool"_ should use `a` rather than `an`


Nevon:
Thanks for the fix!

PR #1188, opened 2021-09-15
doc: use forEach instead map

javigaralva:
In the examples, the .map function should be a .forEach function instead.


Nevon:
Thanks!

PR #1193, opened 2021-09-23
Fixed connections to brokers that share the same host and port

### LINKED ISSUE #1190: Send message to a specific partition
Hi!
I'm trying to send a message in to a specific partition, just like the example code in the documentation, but for example, if my topic has 2 partition (0 and 1) im only able to send Message to partition 1, and if i have 5 partition (0,1,2,3,4) I only can send to the partition 5. Can anyone help me?

dcrn:
This fixes #1190

The problem was caused by brokers getting overridden by the seed broker, because all nodes share the same host and port in Oracle Cloud's Streaming service. I changed it to update the nodeId and rack (if necessary) before making the list of brokers, which allows us to use a more specific check when returning the seed broker itself.

### DIFF
--- a/src/cluster/brokerPool.js
+++ b/src/cluster/brokerPool.js
@@ -41,6 +41,7 @@ module.exports = class BrokerPool {
         allowAutoTopicCreation,
         authenticationTimeout,
         reauthenticationThreshold,
+        onDisconnect: broker => this.removeBroker({ broker }),
         ...options,
       })
 
@@ -75,7 +76,6 @@ module.exports = class BrokerPool {
       connection: await this.connectionBuilder.build(),
       logger: this.rootLogger,
     })
-    this.seedBroker.connection.onDisconnect = () => this.removeBroker({ broker: this.seedBroker })
   }
 
   /**
@@ -138,6 +138,7 @@ module.exports = class BrokerPool {
       this.metadataExpireAt = null
     }
 
+    // The seed broker does not always have a nodeId
     if (broker === this.seedBroker) {
       const nextSeedBroker = shuffle(values(this.brokers))[0]
       if (nextSeedBroker) {
@@ -193,15 +194,15 @@ module.exports = class BrokerPool {
               })
             }
 
-            const broker = this.createBroker({
-              logger: this.rootLogger,
-              versions: this.versions,
-              supportAuthenticationProtocol: this.supportAuthenticationProtocol,
-              connection: await this.connectionBuilder.build({ host, port, rack }),
-              nodeId,
+            return assign(result, {
+              [nodeId]: this.createBroker({
+                logger: this.rootLogger,
+                versions: this.versions,
+                supportAuthenticationProtocol: this.supportAuthenticationProtocol,
+                connection: await this.connectionBuilder.build({ host, port, rack }),
+                nodeId,
+              }),
             })
-            broker.connection.onDisconnect = () => this.removeBroker({ broker })
-            return assign(result, { [nodeId]: broker })
           },
           this.brokers
         )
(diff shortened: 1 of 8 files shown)


src/cluster/brokerPool.js
```diff
@@ -163,6 +163,23 @@ module.exports = class BrokerPool {
 
         const replacedBrokers = []
 
+        // Update the seed broker if necessary
```
Nevon, line 166: Could you write a unit test that explains what this is doing and why? I understand this is a workaround for something in Oracle's streaming service, but I have no idea when it's necessary to update the seed broker or why.

Nevon (changes requested):

src/cluster/brokerPool.js
dcrn: Thanks for checking this @Nevon. The bug fix here is specifically on line 195, which checks if the nodeId matches the seed broker's nodeId before returning it in the reduce call. Updating the seed broker is existing logic that I had to move around, but still works the same.

The current code does the seed broker nodeId and rack updates on lines 179-180: https://github.com/tulios/kafkajs/pull/1193/files#diff-017114e919136dd5728c2aa40fbab06e921c35272aa7d220f6f5e11afd29ee93L179-L180

Having checked the git history, it looks like this was done to resolve some connection issues, e.g. if the seed broker is recreated and has a new nodeId or rack. These commits were a few years ago, so it's hard to say if it's still needed or not. Do you think it can be removed?

dcrn: I tested removing this code and discovered several tests that failed. Turns out this code is required for the seed broker, since it starts off with just a host and port (no nodeId / rack): https://github.com/tulios/kafkajs/blob/master/src/cluster/brokerPool.js#L74-L77

I clarified this in the code a bit and added tests that target this.

dcrn:
@Nevon I found one more place that makes the assumption about one node per host/port: `brokerPool.removeBroker({ host, port })`

It seems that this is only called on KafkaJSConnectionClosedError. Do you have any recommendations here? Would it be correct to remove all brokers with the specified host and port?

Nevon:
> It seems that this is only called on KafkaJSConnectionClosedError. Do you have any recommendations here? Would it be correct to remove all brokers with the specified host and port?

That would probably "work", since all requests would be rejected and on subsequent retries we would reconnect, but I'm not sure if it makes sense since in the case of Oracle Streaming service that would apparently mean rejecting all requests and disconnecting all brokers.

One option is to change `brokerPool.removeBroker` to take a `Broker` instance instead of a host+port. I don't like this solution because it means that the `Connection` has to have a reference to the broker, despite being on a lower layer and not being supposed to know anything about brokers (it does have a `broker` property, but it's just a convenience over `host:port` - it's not an actual broker). The reason the `Connection` would need it is to be able to include it as a property in the `KafkaJSConnectionClosedError`.

Another option is to change how the removal of the `Broker` from the `BrokerPool` works. Currently, the `Cluster` is responsible for removing the broker from the brokerpool, but what if we change it so that the broker itself can deregister itself from the brokerpool? When the broker is instantiated, the brokerpool can pass in a function that allows the broker to remove itself from the brokerpool, which it can call when it gets a `KafkaJSConnectionClosedError` (currently it just disconnects). We can then remove the corresponding code from the `Consumer`, `Producer` and `ConsumerGroup`.

```js
this.createBroker = options =>
  new Broker({
    allowAutoTopicCreation,
    authenticationTimeout,
    reauthenticationThreshold,
    onDisconnect: broker => {
      this.removeBroker(broker)
    },
    ...options,
  })
```

I don't know if there's a better solution, but at least it avoids a circular dependency between the broker and the connection, and probably allows us to remove some duplicated code.

> all nodes share the same host and port in Oracle Cloud's Streaming service

On another note, do you have an account manager or something you can contact in Oracle Cloud about this kind of issue? While I'm happy and thankful that you're trying to solve your own issue, it grinds my gears a bit that you and I are paying the cost of Oracle doing weird shit (you can put AWS MSK in the same category) instead of them supporting their own product. Especially since now we have to pay the maintenance cost of this forever, without even being able to reproduce the issue since their service is closed-source.

PR #1202, opened 2021-10-07
Add overload typings for all event types

brianphillips:
This allows for event handlers to get the appropriately typed data given
the event they are subscribed to.


Nevon:
This is a great change! Thanks for the contribution.

PR #1204, opened 2021-10-08
Move to bitnami/kafka image for kafka

kbariotis:
Using the original Kafka image, I wasn't able to connect through KafkaJS on an Apple M1 MBP. Bitnami supports it so I thought I would update the docs.

Thanks


Nevon:
I think we should probably move to the Confluent images rather than wurstmeister, but that's for another day. Closing this for now.

PR #1207, opened 2021-10-12
add event instrumentation to readme features

arthur-mts:


Nevon:
Thanks for taking the time, but I think if anything we'll re-work the feature list to something more useful. Currently it's kind of a random selection of features, whereas the purpose was initially to highlight that the client supports the most commonly needed functionality - not to be an exhaustive list of all the things it provides.

PR #1215, opened 2021-10-21
Update Producing.md

gaganshera:
The documentation doesn't mention that the `kafka.producer` function can take the JSON options parameter, currently it just shows a list of options after the function call. I've added an extra function call to specify the usage of the  `kafka.producer` function with parameters.


Nevon:
Thanks for the contribution

Issue #1039, opened 2021-03-04
KafkaJSLockTimeout when producing large amount of messages at once


j-a-h-i-r:
This issue is extremely close to the https://github.com/tulios/kafkajs/issues/554. The difference is that I am using `SASL/OAUTHBEARER` for authentication.

**Some background**
I am watching a directory using chokidar. When a new file arrives in the directory, it is immediately sent to Kafka.
I am currently stress testing my Kafka setup. For this, I am copying some (~1000) files to the directory using a bash script. Unfortunately, this is causing errors and only a fraction of files (200-400) are ending up in Kafka.

**Errors faced**
Below is the list of errors being thrown from KafkaJS. I have tried to add as many errors as possible. I have formatted the errors for brevity.

```
{"level":"ERROR",
"message": "[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (55 waiting locks): "connect to broker localhost:9093"",
"retryCount": 0, "retryTime": 311, 
"stack": "KafkaJSLockTimeout: Timeout while acquiring lock (55 waiting locks):  "connect to broker localhost:9093"
at Timeout._onTimeout (node_modules\\kafkajs\\src\\utils\\lock.js:48:23)}
```

```
{"level": "ERROR", "message": "[Connection] Connection timeout", "broker": "localhost:9093", "clientId":"test-producer"}
```

```
{"level":"ERROR",
"message":"[SASLOAuthBearerAuthenticator] SASL OAUTHBEARER authentication failed: Not connected", "broker": "localhost:9093"}

{"level":"ERROR",
"message":"[BrokerPool] KafkaJSSASLAuthenticationError: SASL OAUTHBEARER authentication failed: Not connected",
"retryCount":0, "retryTime":306,
"stack": "KafkaJSSASLAuthenticationError: SASL OAUTHBEARER authentication failed: Not connected
at OAuthBearerAuthenticator.authenticate (node_modules\\kafkajs\\src\\broker\\saslAuthenticator\\oauthBearer.js:49:21)
---- snipped -----
at async sendBatch (node_modules\\kafkajs\\src\\producer\\messageProducer.js:95:12)"}
```

```
error:  KafkaJSNonRetriableError
  Caused by: KafkaJSLockTimeout: Timeout while acquiring 
lock (107 waiting locks): "connect to broker localhost:9093"
at Timeout._onTimeout (node_modules\kafkajs\src\utils\lock.js:48:23)
[...]

PR #1241, opened 2021-11-29
authenticate on every request to broker when needed

### LINKED ISSUE #1240: KafkaJS is not re-auhtenticating to all brokers
**Describe the bug**
When I run kafkajs with multiple brokers I see in logs that it authenticated to all of them successfuly
```
[KafkaJS] [SASLOAuthBearerAuthenticator][DEBUG] - Authenticate with SASL OAUTHBEARER kafka-1.flucivja.local:9094
[KafkaJS] [SASLOAuthBearerAuthenticator][DEBUG] - SASL OAUTHBEARER authentication successful kafka-1.flucivja.local:9094

[KafkaJS] [SASLOAuthBearerAuthenticator][DEBUG] - Authenticate with SASL OAUTHBEARER kafka-2.flucivja.local:9094
[KafkaJS] [SASLOAuthBearerAuthenticator][DEBUG] - SASL OAUTHBEARER authentication successful kafka-2.flucivja.local:9094
```

then I see it succesfully receving topics and when `reauthenticationThreshold ` is hit then re-authentication happes but only to first broker
```
[KafkaJS] [SASLOAuthBearerAuthenticator][DEBUG] - Authenticate with SASL OAUTHBEARER kafka-1.flucivja.local:9094
[KafkaJS] [SASLOAuthBearerAuthenticator][DEBUG] - SASL OAUTHBEARER authentication successful kafka-1.flucivja.local:9094
```

then it is working for a while and suddenly I got this error where it is clearly visible that connection was closed by second broker because client didn't re-authenticate to it.
```
[KafkaJS] [Connection][DEBUG] - Kafka server has closed connection kafka-2.flucivja.local:9094
[KafkaJS] [Connection][DEBUG] - disconnecting...
[KafkaJS] [Connection][DEBUG] - disconnected
[KafkaJS] [Runner][DEBUG] - Error while fetching data, trying again...
[KafkaJS] [ConsumerGroup][DEBUG] - Fetching from 1 partitions for 1 out of 1 topics
 [KafkaJS] [Connection][DEBUG] - Request Fetch(key: 1, version: 11)
 [KafkaJS] [Connection][DEBUG] - Response Fetch(key: 1, version: 11)
 [KafkaJS] [Runner][DEBUG] - Error while fetching data, trying again...
... repeated several times
[KafkaJS] [Consumer][ERROR] - Crash: KafkaJSNumberOfRetriesExceeded: Not connected
[KafkaJS] [Runner][DEBUG] - stop consumer group
[KafkaJS] [Consumer][INFO] - Stopped
[KafkaJS] [Consumer][DEBUG] - consumer has stopped, disconnecting
[KafkaJS] [Connection][DEBUG] - disconnecting...
[KafkaJS] [Connection][DEBUG] - disconnected
[KafkaJS] [Consumer][ERROR] - Restarting the consumer
...then consumer is restarted and same happens every 5 minutes
```

**To Reproduce**
You need to run KafkaJS with multiple brokers with SASL OAUTHBEARER auth and server brokers must be configured that client needs to reauthenticate every e.g. 5 minutes.

**Expected behavior**
KafkaJS re-authenticates to all connected brokers

**Environment:**
 - Mac OS and also Linux
 - KafkaJS version 1.15.0
 - NodeJS version 14

flucivja:
Fixes #1240

### DIFF
--- a/src/broker/__tests__/reauthenticate.spec.js
+++ b/src/broker/__tests__/reauthenticate.spec.js
@@ -1,0 +1,118 @@
+const {
+  saslOAuthBearerConnectionOpts,
+  newLogger,
+  describeIfOauthbearerEnabled,
+  createConnection,
+} = require('testHelpers')
+
+const { SaslAuthenticate: apiSaslAuthenticateKey } = require('../../protocol/requests/apiKeys')
+const Broker = require('../index')
+
+/**
+ * Session lifetime on broker is configured in docker-compose.2_4_oauthbearer.yml with 15000ms in env KAFKA_CONNECTIONS_MAX_REAUTH_MS
+ * see https://docs.confluent.io/platform/current/installation/configuration/broker-configs.html#brokerconfigs_connections.max.reauth.ms
+ */
+describeIfOauthbearerEnabled('Brokers re-authentication with SASL OAUTHBEARER', () => {
+  let broker
+
+  const wait = (time = 1000) =>
+    new Promise(resolve => {
+      setTimeout(resolve, time)
+    })
+
+  const getAuthCalls = spy =>
+    spy.mock.calls.filter(mock => mock[0].request.apiKey === apiSaslAuthenticateKey)
+
+  test('Does not re-authenticate when still within session validity threshold', async () => {
+    const connection = createConnection(saslOAuthBearerConnectionOpts())
+    const spy = jest.spyOn(connection, 'send')
+    broker = new Broker({
+      connection,
+      logger: newLogger(),
+      reauthenticationThreshold: 2000,
+    })
+    await broker.connect()
+    await wait(1000)
+    await broker.listGroups()
+    await wait(1000)
+    await broker.listGroups()
+    await wait(1000)
+    await broker.listGroups()
+
+    expect(getAuthCalls(spy).length).toBe(1)
+  })
+
+  test('Re-authenticate if needed before making a request', async () => {
+    const connection = createConnection(saslOAuthBearerConnectionOpts())
+    const spy = jest.spyOn(connection, 'send')
+    broker = new Broker({
+      connection,
+      logger: newLogger(),
+      reauthenticationThreshold: 2000,
+    })
+    await broker.connect()
+    await wait(13000)
+    await broker.listGroups()
+
+    expect(getAuthCalls(spy).length).toBe(2)
+  })
+
+  test('Re-authenticate only once with multiple parallel requests', async () => {
+    const connection = createConnection(saslOAuthBearerConnectionOpts())
+    const spy = jest.spyOn(connection, 'send')
+    broker = new Broker({
+      connection,
+      logger: newLogger(),
+      reauthenticationThreshold: 2000,
+    })
+    await broker.connect()
+    await wait(13000)
+    await Promise.all([
+      broker.listGroups(),
+      broker.listGroups(),
+      broker.listGroups(),
+      broker.listGroups(),
+    ])
+
+    expect(getAuthCalls(spy).length).toBe(2)
+  })
+
+  test('Re-authenticate every request sent when re-authentication threshold is same as session lifetime', async () => {
+    const connection = createConnection(saslOAuthBearerConnectionOpts())
+    const spy = jest.spyOn(connection, 'send')
+    broker = new Broker({
+      connection,
+      logger: newLogger(),
+      reauthenticationThreshold: 15000,
+    })
+    await broker.connect()
+    await wait(1000)
+    await broker.listGroups()
+    await wait(1000)
+    await broker.listGroups()
+    await wait(1000)
+    await broker.listGroups()
+
+    expect(getAuthCalls(spy).length).toBe(4)
+  })
+
+  test('Re-authenticates only once with multiple requests sent in parallel when re-authentication threshold is same as session lifetime', async () => {
+    const connection = createConnection(saslOAuthBearerConnectionOpts())
+    const spy = jest.spyOn(connection, 'send')
+    broker = new Broker({
+      connection,
+      logger: newLogger(),
+      reauthenticationThreshold: 15000,
+    })
+    await broker.connect()
+    await wait(1000)
+    await Promise.all([
+      broker.listGroups(),
+      broker.listGroups(),
+      broker.listGroups(),
+      broker.listGroups(),
+    ])
+
+    expect(getAuthCalls(spy).length).toBe(2)
+  })
+})
--- a/src/broker/__tests__/reauthenticate.spec.js
+++ b/src/broker/__tests__/reauthenticate.spec.js
@@ -48,10 +48,10 @@ describeIfOauthbearerEnabled('Brokers re-authentication with SASL OAUTHBEARER', 
     broker = new Broker({
       connection,
       logger: newLogger(),
-      reauthenticationThreshold: 2000,
+      reauthenticationThreshold: 14900,
     })
     await broker.connect()
-    await wait(13000)
+    await wait(1000)
     await broker.listGroups()
 
     expect(getAuthCalls(spy).length).toBe(2)
@@ -63,10 +63,10 @@ describeIfOauthbearerEnabled('Brokers re-authentication with SASL OAUTHBEARER', 
     broker = new Broker({
       connection,
       logger: newLogger(),
-      reauthenticationThreshold: 2000,
+      reauthenticationThreshold: 14900,
     })
     await broker.connect()
-    await wait(13000)
+    await wait(1000)
     await Promise.all([
       broker.listGroups(),
       broker.listGroups(),
--- a/src/broker/index.js
+++ b/src/broker/index.js
@@ -951,10 +951,10 @@ module.exports = class Broker {
    * @private
    */
   async [PRIVATE.SEND_REQUEST](protocolRequest) {
+    if (!this.isAuthenticated() && isAuthenticatedRequest(protocolRequest.request)) {
+      await this[PRIVATE.AUTHENTICATE]()
+    }
     try {
-      if (!this.isAuthenticated() && isAuthenticatedRequest(protocolRequest.request)) {
-        await this[PRIVATE.AUTHENTICATE]()
-      }
       return await this.connection.send(protocolRequest)
     } catch (e) {
       if (e.name === 'KafkaJSConnectionClosedError') {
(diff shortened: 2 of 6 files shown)


Nevon:
Is this issue really limited to heartbeats? From reading the issue description, it sounds like the problem should occur on any request if the broker needs to reauthenticate.

If that's the case, then we could do the same thing in [sendRequest](https://github.com/tulios/kafkajs/blob/f443fc5f61fa09110bde7e9cee55c96d358a52ad/src/broker/index.js#L916), except when the request is `apiVersions`, since that happens before authentication.

src/broker/index.js
```diff
@@ -103,35 +110,45 @@ module.exports = class Broker {
...
-
-        this.logger.debug(`Verified support for SaslAuthenticate`, {
-          broker: this.brokerAddress,
-          supportAuthenticationProtocol: this.supportAuthenticationProtocol,
-        })
+  /**
+   * @private
+   * @returns {Promise}
+   */
+  async authenticate() {
```
Nevon, line 123: If this can be called from other places than the connect method, then this also needs to also be within a lock so that you don't get multiple authentication processes going on simultaneously towards the same broker.

flucivja:
> Is this issue really limited to heartbeats? From reading the issue description, it sounds like the problem should occur on any request if the broker needs to reauthenticate.
> 
> If that's the case, then we could do the same thing in [sendRequest](https://github.com/tulios/kafkajs/blob/f443fc5f61fa09110bde7e9cee55c96d358a52ad/src/broker/index.js#L916), except when the request is `apiVersions`, since that happens before authentication.

yes, it can happend at any point. I will add it into [sendRequest]. 
btw. how to check if request is not `apiVersions`? By simply comparing apiName with string? Or better to get version protocol by `requests.ApiVersions.protocol` and compare names?

src/broker/index.js
flucivja: good point, I will add it

Nevon:
> btw. how to check if request is not apiVersions? By simply comparing apiName with string? Or better to get version protocol by requests.ApiVersions.protocol and compare names?

Not sure it matters much, but you could grab the api key of ApiVersions from `src/protocol/requests/apiKeys.js` and compare it with `request.apiKey`. The `apiName` shouldn't change, but `apiKey` definitely won't.

src/broker/index.js
```diff
@@ -63,25 +73,37 @@ module.exports = class Broker {
 
     // The lock timeout has twice the connectionTimeout because the same timeout is used
     // for the first apiVersions call
-    const lockTimeout = 2 * this.connection.connectionTimeout + this.authenticationTimeout
+    const lockTimeout = 2 * this.connection.connectionTimeout
```
Nevon, line 76: This still needs to have the `authenticationTimeout` added, since we're still authenticating within the same flow.

flucivja:
@Nevon everything should be ok now. Can you please check? :) Thanks.

src/broker/index.js
```diff
@@ -71,17 +81,29 @@ module.exports = class Broker {
       description: `connect to broker ${this.brokerAddress}`,
     })
 
+    this.authLock = new Lock({
+      timeout: this.authenticationTimeout,
```
Nevon, line 85: Let's just add a little 50ms buffer to the lock timeout. The reason being that if there is an authentication timeout, we want the error to come from the authentication timeout and not a generic lock timeout. The lock timeout is just there to ensure that we don't hold a lock forever.

flucivja: good thinking, I'll add it

flucivja: done

Nevon:
I've been trying to figure out how to test this, and I wrote a test that looks something like this:

```js
test('Reauthenticates if needed before making a request', async () => {
  const saslEntry = saslEntries.pop()
  const connection = createConnection(saslEntry.opts())
  broker = new Broker({
    connection,
    logger: newLogger(),
  })
  await broker.connect()
  const spy = jest.spyOn(connection, 'send').mockImplementation()

  broker.authenticatedAt = null
  await broker.listGroups()

  // Assert that sasl authenticate requests happen
  console.log(spy.mock.calls)
})
```

But I ran into the problem that the server rejects the second authentication request with an ILLEGAL_SASL_STATE error. This got me thinking and reading the [KIP about reauthentication](https://cwiki.apache.org/confluence/display/KAFKA/KIP-368%3A+Allow+SASL+Connections+to+Periodically+Re-Authenticate).

In the current implementation, if you have multiple requests going on when you realize that you have to reauthenticate, each one is going to try to reauthenticate, which is gonna cause a problem. This is relatively easily solved by having the `isAuthenticated` method wait for the `authLock` to be released, so that you wait to check if you need to authenticate while there's an authentication going on. Unfortunately it makes the method async, which ripples through and affects a lot of things. A pain in the ass to fix, but doable.

flucivja:
What do you think about putting `isAuthenticated` check into `[PRIVATE.AUTHENTICATE]()` after auth lock? And if it will be authenticated, then it will skip auth and release lock.

t-d-d:
You might want to look at the sharedPromise in utils.  Heartbeat() is
wrapped with this so that when multiple are issues concurrently, only 1
actual request is sent.

flucivja:
`sharedPromise` looks handy for such case. But I am thinking if this is really an issue. I tried that test without my changes and it is failing too. @Nevon is that test passing for you without my changes (without check for re-authentication)?

Nevon:
> @Nevon is that test passing for you without my changes (without check for re-authentication)?

No, but I think it's an issue with the test. Maybe you can't authenticate again right after authenticating with sasl/scram like I'm doing. I'm guessing if you swap it out for oauthbearer with a short reauthentication timeout maybe it'll work. I just didn't have time to test it out.

> But I am thinking if this is really an issue.

I would be very surprised if it isn't. Tossing `broker.listGroups` or something into a `Promise.all` should trigger it immediately.

flucivja:
Ok then, are you ok to handle it with `sharedPromise`? It will look like this in the broker constructor:

```
this[PRIVATE.AUTHENTICATE] = sharedPromiseTo(async () => {
  ... authentication logic
})
```

Nevon:
Sounds like a good solution! Let's give it a try. I would also really like to test these scenarios. We do have a test suite that uses oauthbearer, which should certainly allow for reauthentication. See the `test:group:oauthbearer` yarn script, which runs a subset of test files. There's a function called `describeIfOauthbearerEnabled` in `testHelpers` that will run tests nested in it only if oauthbearer is enabled.

flucivja:
Hi @Nevon, I implemented tests and sorry it took longer time. I had issues with kafka SASL OAUTHBEARER authentication. Within test, the kafka returned me all the time that provided token created in test utils is invalid and I didn't know why. After debugging and looking to how other kafka clients are implemeneted, I found out that [expiration time](https://github.com/tulios/kafkajs/pull/1241/files#diff-889932f4cee3b60450073a0c903c4e885fb045142b8343a6ade91c4020b5419a) was missing in token.

Btw. you were right with multiple pararel requests causing error: when I removed fix with `sharedPromise` my tests which runs multuple requests in pararel requiring authentication started failing with invalid SASL state.

Can you please check PR again when you will have time? If all will be good, can you merge it and release?

Thank you very much.

flucivja:
@Nevon just a gentle reminder :)

src/broker/__tests__/reauthenticate.spec.js
```diff
@@ -0,0 +1,118 @@
...
+  newLogger,
+  describeIfOauthbearerEnabled,
+  createConnection,
+} = require('testHelpers')
+
+const { SaslAuthenticate: apiSaslAuthenticateKey } = require('../../protocol/requests/apiKeys')
+const Broker = require('../index')
+
+/**
+ * Session lifetime on broker is configured in docker-compose.2_4_oauthbearer.yml with 15000ms in env KAFKA_CONNECTIONS_MAX_REAUTH_MS
```
Nevon, line 12: Is it possible to reauthenticate much sooner than this? I'm thinking maybe we can avoid the excessive sleeps in the test by either setting the `reauthenticationThreshold` to like 14900ms or something so we only need to sleep a little bit, or override `broker.sessionLifetime` to be 0 whenever we want to trigger a reauthentication. Something like:

```ts
    const connection = createConnection(saslOAuthBearerConnectionOpts())
    const spy = jest.spyOn(connection, 'send')
    broker = new Broker({
      connection,
      logger: newLogger(),
    })
    await broker.connect()
    await broker.listGroups()
    await broker.listGroups()

    broker.sessionLifetime = 0n;
    await broker.listGroups()

    expect(getAuthCalls(spy).length).toBe(2)
```

With all the sleeps in the current test, we're adding about half a minute or more to the execution time that we can maybe avoid.

Nevon:
Haven't forgotten about you. I left the notification sitting on my phone so it wouldn't get lost. 😅  The change looks good overall - I'd just like to get your thoughts on whether we can get rid of those sleeps in the test (or at least reduce them).

src/broker/index.js
```diff
@@ -62,16 +73,55 @@ module.exports = class Broker {
...
+
     this.lookupRequest = notInitializedLookup
+
+    /**
+     * @private
+     * @returns {Promise}
+     */
+    this[PRIVATE.AUTHENTICATE] = sharedPromiseTo(async () => {
+      try {
+        await this.authLock.acquire()
```
Nevon, line 99: We don't need this lock anymore now right? If there's more callers, the first one will create a promise and start the authentication, the second one will just receive the promise from the first one. Since the promise creation itself is synchronous there shouldn't be any race condition there, I think.

src/broker/__tests__/reauthenticate.spec.js
flucivja: yes sure, I will reduce it by using higher `reauthenticationThreshold `.

src/broker/index.js
flucivja: yes you are right. I will remove auth lock.

flucivja:
@Nevon done, now my test is faster by 30 seconds (14s vs 44s).

Nevon (approved):

Nevon:
Thanks for your patience and for taking the time to contribute!

j-a-h-i-r:
Sorry for reviving a merged thread but I'm curious about one thing @Nevon. Does this fix https://github.com/tulios/kafkajs/issues/1039? As far as I can remember, that issue occurred due to re-authentication. I saw that you mentioned testing SASL OauthBearer flows. So I was wondering if the changes potentially fix that issue too.

PR #1254, opened 2021-12-17
Ensure error classes are exported

### LINKED ISSUE #982: Error class are not exported
**Describe the bug**
This bug report follows the previous issue #636. The errors are correctly described in the typescript definition but the error classes are not exported in the index of javascript file.

**To Reproduce**
When I test kafka error with an `instanceof` : 
```typescript
import { IResourceConfig, Admin, ConfigResourceTypes, KafkaJSProtocolError } from 'kafkajs';
...
try {
  await admin.alterConfigs({
    validateOnly: false,
    resources: [{
        type: ConfigResourceTypes.TOPIC,
        name: 'topic-name',
        configEntries: [{ name: 'cleanup.policy', value: 'compact' }]
    }]);
} catch (e) {
  if (e instanceof KafkaJSProtocolError && e.type === 'UNKNOWN_TOPIC_OR_PARTITION') {
    // do something
  } else {
    throw e;
  }
}
```
this throws the error : `TypeError: Right-hand side of 'instanceof' is not an object ` at `e instanceof KafkaJSProtocolError` in catch block. This is not a compile error because the typescript definition file is good but it is an execution error. When I log `KafkaJSProtocolError`, the output console is `undefined`.

**Expected behavior**
I think the root `index.js` file must export the errors from `src/errors.js` file.

julienvincent:
Fixes #982 

The KafkaJS error classes currently aren't exported which makes it hard to distinguish between application errors and kafka errors, especially while developing utils/libraries on top of KafkaJS.

This is especially confusing as the TS types included _do_ export the error classes meaning these errors are only discovered at runtime.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,7 @@ const AclResourceTypes = require('./src/protocol/aclResourceTypes')
 const AclOperationTypes = require('./src/protocol/aclOperationTypes')
 const AclPermissionTypes = require('./src/protocol/aclPermissionTypes')
 const ResourcePatternTypes = require('./src/protocol/resourcePatternTypes')
+const Errors = require('./src/errors')
 const { LEVELS } = require('./src/loggers')
 
 module.exports = {
@@ -33,4 +34,5 @@ module.exports = {
   AclPermissionTypes,
   ResourcePatternTypes,
   ConfigSource,
+  ...Errors,
 }


Nevon:
Damn. I think it was a mistake to export the error classes from TS to begin with. We should probably have exported error interfaces instead, because there's no situation in which you as a consumer of KafkaJS should be instantiating those errors yourself. You just need to know what fields are publicly available on the error instances that we throw, and do figure out which error you received in case of a union, which we could do by exposing a `type` field on the error interface instead.

But unfortunately this was done in 1.15.0, so it's a public API now :(

julienvincent:
@Nevon 

> because there's no situation in which you as a consumer of KafkaJS should be instantiating those errors yourself

In our case It's not that we want to instantiate the error, but more that we want to do `instanceof` checks on it. As an example: We have built an eachBatch util which adds some additional logic like retries, tracing, metrics, and error reporting - but we want to fall through when a KafkaJS error is encountered (which might be encountered when calling methods on a batch like `heartbeat`)

This looks something like:

```ts
const eachBatchUtil = (handler: kafkajs.EachBatchHandler): kafkajs.EachBatchHandler => {
  return async (payload) => {
    try {
      await handler(payload)
    } catch (err) {
      if (err instanceof kafkajs.KafkaJSError) {
        throw err;
      }

      // .. attempt retries
    }
  }
}

consumer.run({
  eachBatch: eachBatchUtil(async (payload) => { ... })
})
```

Nevon:
Yeah, that's why I mentioned:

> figure out which error you received in case of a union, which we could do by exposing a type field on the error interface instead.

We don't even do `instanceof` comparisons inside KafkaJS itself. Instead, each error class has a static `type` property that we use to determine which type of error we've received.

julienvincent:
Oh, gotcha. :)

PR #1255, opened 2021-12-17
Allow calling `heartbeat` from `eachMessage` handler

julienvincent:
This is a small improvement for a pain-point I have had for a while where I have often had to 'eject' out to an `eachBatch` handler just to gain access to the heartbeat function.

Sometimes consumers have to do a lot of work for individual messages in a topic and in these scenarios it makes sense to have access to the heartbeat function to avoid the session timeout and keep the consumer alive.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -858,6 +858,7 @@ export interface EachMessagePayload {
   topic: string
   partition: number
   message: KafkaMessage
+  heartbeat(): Promise<void>
 }
 
 export interface EachBatchPayload {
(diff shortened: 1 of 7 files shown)


types/index.d.ts
```diff
@@ -882,14 +883,17 @@ export type ConsumerEachMessagePayload = EachMessagePayload
  */
 export type ConsumerEachBatchPayload = EachBatchPayload
 
+export type EachBatchHandler = (payload: EachBatchPayload) => Promise<void>
+export type EachMessageHandler = (payload: EachMessagePayload) => Promise<void>
```
julienvincent, line 887: I've explicitly exported these two types to allow using them in utils and libraries (another small pain-point of mine :)

Nevon:
Could you please write a test that makes use of this? It's pretty simple, but it's more to prevent regression. It would also be good if you could update the docs with the new property.

julienvincent:
Will do. Also need to get the existing tests to pass

Nevon (approved):

PR #1256, opened 2021-12-19
Better handle empty control batches

julienvincent:
This is a two part fix for transactionally aware consumers that are processing aborted transactions.

## Instrumentation Events

The problem with the current implementation is that if a batch containing only control records or only aborted messages is received then there is no way for monitors to be aware that the consumer progressed. This can happen, for example, when reading aborted transactions.

This is fixed by emitting start and end batch instrumentation events if the batch is an empty control batch or only contains aborted messages.

## Consume Lockout

Secondarily there exists an issue wherein if a batch containing only aborted messages or only control messages is received then the consumer never resolves + commits the batch offset, resulting in the consumer processing the same batch repeatedly and getting blocked.

Similarly to the previous scenario, this is fixed by checking for this scenario and resolving + auto-committing the batches latest offset.


julienvincent:
@Nevon happy to merge?

julienvincent:
@Nevon Looks like the pipeline had an ephemeral network failure. Can you re-trigger the jobs?

Nevon:
Triggered another run now. Not sure if the Azure DevOps Github integration is smart enough for the build status notifications to work. If not, you can always push an empty commit to trigger it.

julienvincent:
@Nevon Ok, I re-pushed to force the jobs. All is passing 👍

julienvincent:
Hey @Nevon, any chance I could get some eyes on this?

Nevon:
Next week I will have some time to spend on KafkaJS, and will make sure that this is high up on the todo list.

julienvincent:
Thanks! 🙏

Issue #683, opened 2020-03-30
Fetch partitions in parallel


hmagarotto:
When the processing of one partition is slow, all partitions are affected.
It's because the "fetch" for other partitions only is performed after the process completion for all partitions.
The "fetch" for all topics and partitions are synchronous.
So, it's not possible to use the partitions as a unit of parallelism.

Can we deal with this using some setting or using any different strategy?

Test with a slow process (exaggerated) on partition 0:
https://github.com/hmagarotto/kafkajs-test-partition-parallelism

ankon:
This sounds a bit related to #370 (#570) to me. If so, maybe you could try the beta releases with the test?

hmagarotto:
Hi, thanks for your response.

What I understood about this refactoring is that:

The consumer will process batches immediately after a response for each fetch request.
But, a new fetch for this consumer group will only be started after all batches has already been processed. For a new fetch start, we need to wait for "enqueuedTasks" on "Runner::fetch".

So, if I have understood correctly, a slow task for some partition will block new fetches for all other partitions.

I tried the test with the new beta version and all partitions stay blocked by slower partition task.

Test code updated with beta KafkaJS: https://github.com/hmagarotto/kafkajs-test-partition-parallelism

Similar needs for another library/language: https://github.com/akka/alpakka-kafka/issues/110

Nevon:
Your understanding is correct, @hmagarotto. Currently the fetch loop will wait until all partitions have been processed before issuing new fetch requests. I've been having [the same discussion on our Slack today](https://kafkajs.slack.com/archives/CF6RFPF6K/p1585661917057500), and I think it makes a lot of sense to start the next fetch for a broker as soon as we have processed the last response from it, to avoid having to wait for the slowest one.

tristan-redox:
@Nevon, we recently tried a BETA version of KafkaJS and noticed that another fetch started before the current processing fetch completed.  The problem we encountered was that the second fetch received the same message/offset from the same topic/partition that was currently being processed in the yet to finish fetch.  This resulted in the same message being processed twice "concurrently" within <50 ms of each other.  This seems like a bug, should I raise a bug report against the BETA version?

I think starting the next fetch before the previous finishes, but only if there are guarantees that prevent what I described above (duplicate concurrent processing of the same messages).

Nevon:
[...]

PR #1258, opened 2021-12-20
#683 Improve concurrency

priitkaard:
Notable changes:
* Removed current concurrency + barrier logic for batch processing
* Created RunnerPool, which spawns `partitionsConsumedConcurrently` number of independent runners
* Created FetchManager, which fetches different brokers independently and assigns partitions to different runners during the consumer group sync
* Each connection now opens two sockets - one for fetch requests, the other one for the rest (TCP sockets preserve the order of messages sent to a broker. That means that other concurrent requests, eg. heartbeats and offset fetch/commit, from other runners would wait behind the fetch polls and drastically slows down the consumption in the concurrent scenario)
* Removed memory leaks from hanging consumers in tests  

Manually compared consumption times with the scripts below:

Producer:
```js

const kafka = new Kafka({ brokers: ['localhost:9092'] });

const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));

(async () => {
    // const admin = kafka.admin();
    // await admin.connect()
    // await admin.deleteTopics({ topics: ['loadtest'] })
    // await admin.createTopics({ topics: [{ topic: 'loadtest', numPartitions: 3, replicationFactor: 1 }]})
    // await admin.disconnect()

    await producer.connect();

    const batchSize = 100;
    for (let i = 0; i < 5; i++) {
        const messages = Array(batchSize)
            .fill()
            .map((_, index) => ({
                key: `key-${batchSize * i + index}`,
                value: `value-${batchSize * i + index}`,
            }));
        await producer.send({ topic: 'loadtest', messages });
        await sleep(1000);
    }
    await producer.disconnect();

    console.log('Published!');
})();
``` 

Consumer:
```js

const kafka = new Kafka({ brokers: ['localhost:9092'] });

const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));

(async () => {
    const metrics = {};

    const consumer = kafka.consumer({ groupId: 'tulios-consumer', maxWaitTimeInMs: 5000 });
    await consumer.connect();

    process.on('SIGINT', () => {
        Object.entries(metrics).forEach(([partition, { startTime, endTime }]) => {
            console.log(`Partition ${partition} processed in ${endTime - startTime} ms`)
        })
        return consumer.disconnect()
    });

    await consumer.subscribe({ topic: 'loadtest' });
    await consumer.run({
        partitionsConsumedConcurrently: 3,
        eachMessage: async ({ partition, message: { value } }) => {
            if (!metrics[partition]) metrics[partition] = { startTime: Date.now() }

            if (partition === 0) await sleep(100)
            console.log(partition, value.toString());

            metrics[partition].endTime = Date.now()
        },
    });
})();
```

Slower runners now consume their batches at their own pace, other fetch requests will not be blocked.
Partitions 2 and 3 were consumed in ~70ms (- 4s from delays) compared to the current master version, which took ~9s (-4s delays):

image-20211217-152719


KristjanTammekivi:
Solves https://github.com/tulios/kafkajs/issues/683 IIRC

julienvincent:
This is awesome. Been wanting this for a long time!

julienvincent:
FYI I tried this patch out locally and ran into quite a lot of issues + crashes that aren't happening on current master

priitkaard:
> FYI I tried this patch out locally and ran into quite a lot of issues + crashes that aren't happening on current master

Thanks for the feedback. 
Could you specify, which issues did you have? I’ve been running it for the last week on 3 nodes, ~20 consumers, 50 partitions each, on production load. 
No issues so far after the last commit.

priitkaard:
> FYI I tried this patch out locally and ran into quite a lot of issues + crashes that aren't happening on current master.
> 
> Type of failure:
> 
> ```shell
> {"level":"INFO","timestamp":"2021-12-24T09:49:06.098Z","logger":"kafkajs","message":"[Consumer] Starting","groupId":"test-latency"}
> {"level":"INFO","timestamp":"2021-12-24T09:49:06.896Z","logger":"kafkajs","message":"[ConsumerGroup] Consumer has joined the group","groupId":"test-latency","memberId":"kafkajs-ed8e7b9f-58b1-45a3-abff-0e683a9e3f6f","leaderId":"kafkajs-ed8e7b9f-58b1-45a3-abff-0e683a9e3f6f","isLeader":true,"memberAssignment":{"test-latency":[0,1,2,3,4,5,6,7,8,9]},"groupProtocol":"RoundRobinAssigner","duration":794}
> {"level":"WARN","timestamp":"2021-12-24T09:49:12.404Z","logger":"kafkajs","message":"[RequestQueue] Response without match","clientId":"kafkajs","broker":"...","correlationId":16}
> {"level":"WARN","timestamp":"2021-12-24T09:49:12.449Z","logger":"kafkajs","message":"[RequestQueue] Response without match","clientId":"kafkajs","broker":"...","correlationId":17}
> {"level":"WARN","timestamp":"2021-12-24T09:49:12.560Z","logger":"kafkajs","message":"[RequestQueue] Response without match","clientId":"kafkajs","broker":"...","correlationId":15}
> {"level":"ERROR","timestamp":"2021-12-24T09:49:20.052Z","logger":"kafkajs","message":"[Consumer] Crash: KafkaJSNumberOfRetriesExceeded: Closed connection","groupId":"test-latency","retryCount":5,"stack":"KafkaJSNonRetriableError\n  Caused by: KafkaJSConnectionClosedError: Closed connection\n    at TLSSocket.onEnd (/Users/julienvincent/code/kafkajs/src/network/connection.js:148:13)\n    at TLSSocket.emit (node:events:402:35)\n    at TLSSocket.emit (node:domain:475:12)\n    at endReadableNT (node:internal/streams/readable:1343:12)\n    at processTicksAndRejections (node:internal/process/task_queues:83:21)"}
> {"level":"INFO","timestamp":"2021-12-24T09:49:20.162Z","logger":"kafkajs","message":"[Consumer] Stopped","groupId":"test-latency"}
> {"level":"ERROR","timestamp":"2021-12-24T09:49:20.163Z","logger":"kafkajs","message":"[Consumer] Restarting the consumer in 6838ms","retryCount":5,"retryTime":6838,"groupId":"test-latency"}
> 
> ...
> ```

Did you test the latest commit? 
On the first look, eacb34a solved such crashes.

I’ve been running this fork for the last week on 3 nodes, ~20 consumers, 50 partitions each, on production load. 

If you tested the latest version, I will take a look at it in a few days.

julienvincent:
@priitkaard Yes, this is with the latest commit. I tried this patch on 2c37d39 originally and had the same result ^ and then now again on e622ed7.

julienvincent:
FYI this is connecting to a confluent cluster with 10 brokers. Might have something to do with it

priitkaard:
@julienvincent 
Thanks for the feedback! 
Indeed, the separate socket I opened for the fetch polls was not authenticated. 
We don't use SASL on our setup, but I tested the latest commit on confluent with 10 brokers and plain SASL as well.

julienvincent:
New set of errors while trying commit 0619844

```bash
{"level":"INFO","timestamp":"2021-12-30T11:40:22.182Z","logger":"kafkajs","message":"[Consumer] Starting","groupId":"1234"}
{"level":"INFO","timestamp":"2021-12-30T11:40:35.621Z","logger":"kafkajs","message":"[ConsumerGroup] Consumer has joined the group","groupId":"1234","memberId":"kafkajs-efe3c998-1894-41cc-9420-c37ad71eabba","leaderId":"kafkajs-efe3c998-1894-41cc-9420-c37ad71eabba","isLeader":true,"memberAssignment":{"test-latency":[0,1,2,3,4,5,6,7,8,9]},"groupProtocol":"RoundRobinAssigner","duration":13438}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.649Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":274,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.650Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (2 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":279,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (2 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.650Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":287,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.650Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":273,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.651Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":268,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.651Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":316,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.651Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (2 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":268,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (2 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.652Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (2 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":257,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (2 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.652Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":263,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.653Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (3 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":300,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (3 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.655Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":263,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.655Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":293,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.656Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":314,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.656Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":357,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
{"level":"ERROR","timestamp":"2021-12-30T11:40:35.658Z","logger":"kafkajs","message":"[BrokerPool] KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"","retryCount":0,"retryTime":328,"stack":"KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): \"connect to broker xxx.gcp.confluent.cloud:9092\"\n    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)"}
/Users/julienvincent/code/kafkajs/src/utils/lock.js:48
        const error = new KafkaJSLockTimeout(this[PRIVATE.TIMEOUT_ERROR_MESSAGE]())
                      ^
KafkaJSLockTimeout: Timeout while acquiring lock (1 waiting locks): "connect to broker xxx.gcp.confluent.cloud:9092"
    at Timeout._onTimeout (/Users/julienvincent/code/kafkajs/src/utils/lock.js:48:23)
    at listOnTimeout (node:internal/timers:557:17)
    at processTimers (node:internal/timers:500:7) {
  retriable: false,
  helpUrl: undefined
}
```

shyimo:
Hi @priitkaard .
This is an awesome improvement. thanks for working on that !
i would love to help you if you need so we can make sure this feature will be on the next release

priitkaard:
@shyimo thanks! But I think there’s not much to do anymore. I’m already using this fork on decent loads. 

@julienvincent is it still an issue after the last commit? I couldn’t replicate it, even on confluent cluster, but that seemed an obvious issue that could have caused it.

gogoout:
Huge thanks! We've been using it on production for several days now without any problem. We are using confluent kafka with 5 nodes. Around 30+ consumers, 300 + topics. However we are not using SASL as well.

julienvincent:
> @julienvincent is it still an issue after the last commit? I couldn’t replicate it, even on confluent cluster, but that seemed an obvious issue that could have caused it.

@priitkaard Unfortunately I am still able to reproduce the same issue. I've included below the test script I have been using for lag.

<details>
  <summary>Reproduce Script</summary>

```ts
import * as crypto from "crypto"
import * as kafka from "kafkajs"

void (async function () {
  const client = new kafka.Kafka({
    brokers: [""],
    ssl: true,
    sasl: {
      mechanism: "plain",
      username: "",
      password: "",
    },
  })

  const admin = await client.admin()
  await admin.connect()

  const cache = new Map<string, number>()

  await admin.createTopics({
    topics: [
      {
        topic: "test-latency",
        replicationFactor: 3,
        numPartitions: 10,
      },
    ],
  })

  const res = await admin.describeCluster()
  console.log(JSON.stringify(res, null, 2))

  const consumer = client.consumer({
    groupId: "test-latency",
    allowAutoTopicCreation: false,
    maxWaitTimeInMs: 10000,
    sessionTimeout: 20000,
    heartbeatInterval: 3000,
  })

  await consumer.connect()
  await consumer.subscribe({
    topic: "test-latency",
  })

  consumer.run({
    autoCommit: false,
    eachMessage: async (payload) => {
      const id = payload.message.value?.toString()!
      const val = cache.get(id)
      cache.delete(id)
      if (!val) {
        console.log("missing")
        return
      }
      console.log(`latency: ${Date.now() - val}`)
    },
  })

  const producer = client.producer()
  await producer.connect()

  setInterval(async () => {
    const id = crypto.randomBytes(10).toString("hex")
    producer.send({
      topic: "test-latency",

      messages: [
        {
          partition: 0,
          value: id,
        },
      ],
    })
    cache.set(id, Date.now())
  }, 100)

  process.on("SIGINT", async () => {
    await consumer.disconnect()
    process.exit()
  })
})()
```
</details>

julienvincent:
I retested now to confirm and I am actually no longer able to reproduce. Not sure if temperamental (i.e triggered as a result of some state the confluent brokers were in or something similar) or if I was originally testing against the wrong commit.

ramank775:
Thank you @priitkaard. I've been using it on production without any problem with `sasl_ssl`. One of my worker which subscribed to multiple topics was previously taking more than 1 seconds to fetch, which now bring down in `ms `.

shyimo:
Hi @Nevon !
Is there any plan to do a release soon with this improvement ? we test it in our pre production environment and we saw some very good results.
Thanks :)

priitkaard:
Thanks, @Nevon! 
I will need to find some time to merge these recent changes.
Will let you know once it's done.

src/consumer/fetchManager.js
```diff
@@ -0,0 +1,107 @@
...
+    .reduce((acc, _, runnerId) => ({ ...acc, [runnerId]: [] }), {})
+  let assignments = {}
+  let error
+
+  const fetchNode = async nodeId => {
+    if (fetchers[nodeId]) return fetchers[nodeId]
+
+    fetchers[nodeId] = (async () => {
+      const startFetch = Date.now()
+      instrumentationEmitter.emit(FETCH_START, { nodeId })
```
Nevon, line 23: If we're gonna include the `nodeId` in the FETCH_START and `FETCH` events, we gotta update the Typescript types and documentation to reflect that: 

https://github.com/tulios/kafkajs/blob/1876abcb5effd6183c1fbaa50dee9cef269b67a5/types/index.d.ts#L916-L917

https://github.com/tulios/kafkajs/blob/1876abcb5effd6183c1fbaa50dee9cef269b67a5/types/index.d.ts#L817-L820

src/consumer/__tests__/runner.spec.js
```diff
@@ -40,17 +56,27 @@ describe('Consumer > Runner', () => {
...
     }
-    const instrumentationEmitter = new InstrumentationEventEmitter()
+    instrumentationEmitter = new InstrumentationEventEmitter()
     runner = new Runner({
+      consumerGroup,
+      instrumentationEmitter,
+      logger: newLogger(),
+      eachBatch,
+    })
+    runnerPool = createRunnerPool({
```
Nevon, line 66: This looks a bit jacked up now. Shouldn't this test suite be testing the RunnerPool now rather than the Runner, since the runner is encapsulated within the RunnerPool? Do we need to instantiate the runner manually here or can we do all the testing via the runnerpool?

Nevon (changes requested):
I put together a diagram to help me understand how these entities fit together, what their purpose is and how they relate to each other. Have a look and see if I'm misunderstanding something gravely 

consumer-architecture

Here's an Exaclidraw link that might be easier to read: https://excalidraw.com/#json=AOfQa7KcOE2ucxnrhEtAf,Z6Y11eCI6Kp5wWpt4ae-lg

One thing that I'm not sure about is how a runner relates to a particular partition. I saw that there's an assigner that assigns a partition to a particular runner, but I don't really understand why or what that means, since a runner is a generic parallel executor, so I guess it shouldn't matter what partitions it receives batches for.

src/errors.js
```diff
@@ -31,7 +31,7 @@ class KafkaJSProtocolError extends KafkaJSError {
 
 class KafkaJSOffsetOutOfRange extends KafkaJSProtocolError {
   constructor(e, { topic, partition }) {
-    super(e)
+    super(e, { retriable: true })
```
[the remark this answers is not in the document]
Nevon: https://github.com/tulios/kafkajs/pull/1258#discussion_r804422797

Nevon:
Overall this change looks pretty good, it's just very substantial so it's gonna take me a bit to wrap my head around everything.

One thing I want to do is to do something similar to what I did for [the last refactor of the runner](https://github.com/tulios/kafkajs/pull/671#issuecomment-661249724), to see that the parallelism works the way I think it does. If my understanding of the code is correct, now every partition is processed entirely in parallel, such that we are constantly processing (or waiting for data back) from `partitionsConsumedConcurrently` partitions. Before, we'd fetch for all partitions and then process `partitionsConsumedConcurrently` partitions in parallel, but still wait for the slowest one before fetching the next round of batches. This would be an amazing improvement, but I want to actually get some data on it and view it on a timeline to make sure it works the way I think it does.

Nevon:
Finally, just to set some expectations - I spent the last 5 hours reviewing this, and I probably won't have the time to do that kind of review again until March 7th. I work on KafkaJS 25% of my time currently, and so for the next 3 weeks I gotta focus on my other work. I can try to find some time for discussion and feedback before that, but just be aware that I might not be able to spend this much time again for a little bit. I really want to land this change though.

Nevon:
I collected some traces from a consumer and visualized them using [vis-timeline](https://github.com/visjs/vis-timeline). Here's the result:

Screenshot 2022-02-11 at 16 39 27

[Interactive view](https://jsfiddle.net/cLhbprf1/3)

In this example, I just throw in a random sleep in the `eachBatch`, between 0 and 5 seconds. The topic has 3 partitions and I produced about half a million messages to be consumed from a single consumer with `partitionsConsumedConcurrently: 3`.

At the top is a batch being processed. Below that are all the API requests. The requests are stacked based on which broker they are going to.

What we can see is what I expected, which is that each partition is now basically independent from the others. As soon as a batch finishes processing, there's a new fetch request being made even though the other partitions are still being processed. Awesome! With the previous implementation, what this would have looked like would be that we make 3 fetch requests, process all partitions and only once all of them are processed do we start the next fetch cycle.

priitkaard:
@Nevon, Huge thanks for your time!
The diagrams are both very useful.

Initially, I assigned each runner particular partitions and each runner fetched only their own partitions. 
That seemed more obvious, but it quickly turned out that we would need to keep an additional `partitionsConsumedConcurrently` amount of sockets open for each broker in such case as well. 
But I agree, for now, the partition assignment is obsolete. I will remove it today sometime.

An idea for the future - the pool size could be configurable and the runners separable the same way. This could increase performance even more if the amount of open TCP connections isn't a problem.

I think I've solved most of your mentioned comments, but I will spend some more time refactoring the general architecture.
I will let you know once it's ready for another view.

julienvincent:
Over the next week or two I'll be running this through some tests on our systems and will see if I can run this on some of our non-critical production services.

Nevon:
I haven't had a chance to look through the changes yet, so apologies if this is obvious from reading the code, but as I'm heading out for a hike and won't be able to read everything through yet, I thought I'd shoot of this initial question:

> It's recommended to use at least concurrency = the number of nodes to avoid blocking requests.

I'm wondering if we're using the `partitionsConsumedConcurrently` for something it's not meant for. The intent of it was a way of controlling mostly the parallelism of user-space code. As the unit of parallelism in Kafka is the partition, it made sense to basically just say: "How many partitions do you want your consumer to be processing at any given time". Since fetching was done outside of this parallelism, we always had one connection towards each broker that were independent of each other (+ the limit on the number of in flight requests).

My mental model of your changes was that this property was basically preserved. As a user I would only need to know about the number of partitions, which put an upper bound on the `partitionsConsumedConcurrently`. How many brokers there are wouldn't really factor in, since for a given partition I would connect to a single broker, and each unit of parallelism had its own connections towards the broker.

Maybe I'm misunderstanding something, but could you expand a bit on how the number of brokers affects this, and what it is I'm not getting? Is there a shared bottleneck between what's now known as fetchers? Like I said, I haven't yet read the code, so apologies if the answer is obvious once I understand what the workerqueue and workers do.

priitkaard:
‘partitionsConsumedConcurrently’ still controls the amount of batches (partitions) passed to the user-land code concurrently. 

With the current solution, if concurrency = 2 and number of brokers = 3, one fetcher polls from two brokers at the same time. That means if one partition from the first broker receives a message, it would still have to wait up to maxWaitTime until the second broker returns an empty response. 

But I see your point. It’s probably better to have one fetcher per broker and share the worker queue instead. 
I will go over the workerQueue and fetchManager distribution logic on Monday.

src/consumer/fetcher.spec.js
```diff
@@ -0,0 +1,48 @@
...
+
+    nodeIds = [1, 2, 3]
+
+    workers = seq(5, workerId => createWorker({ workerId, handler }))
+    workerQueue = createWorkerQueue({ workers })
+    fetcher = createFetcher({ nodeIds, fetch, workerQueue })
+  })
+
+  it('should fetch all nodes', async () => {
+    fetcher.start()
```
Nevon, line 29:
```suggestion
    await fetcher.start()
```

src/consumer/fetcher.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    await fetcher.stop()
+
+    expect(fetch).toHaveBeenCalledTimes(nodeIds.length)
+    nodeIds.forEach(nodeId => {
+      expect(fetch).toHaveBeenCalledWith(nodeId)
+    })
+  })
+
+  it('should utilize all workers', async () => {
+    fetcher.start()
```
Nevon, line 39:
```suggestion
    await fetcher.start()
```

Nevon: The reason I'm suggesting this is:
1. I think it highlights that the API is flawed if it returns a promise you can't await
2. If the promise is rejected, we'll have an unhandled promise rejection, which isn't very fun to debug.

src/consumer/index.js
```diff
@@ -153,7 +109,7 @@ module.exports = ({
   }
 
   /** @type {import("../../types").Consumer["stop"]} */
-  const stop = async () => {
+  const stop = sharedPromiseTo(async () => {
```
Nevon, line 112: Not that it hurts to make this a shared promise, but is there a particular reason for it to be?

src/consumer/worker.js
```diff
@@ -0,0 +1,34 @@
...
+   * @returns {Promise<void>}
+   */
+  const run = async ({ next }) => {
+    const batch = next()
+    if (!batch) {
+      return
+    }
+
+    await handler(batch, { workerId })
+    return run({ next })
```
Nevon, line 28: If I understand this correctly, the worker is eating off of the workerQueue, and the workerQueue will contain just the batches from this fetch iteration and then resolve once the queue is empty, after which the fetcher will fetch the next round of batches, making this not infinite (which is a good thing, or I'd be concerned about memory leaks).

src/network/connection.js
```diff
@@ -94,10 +118,45 @@ module.exports = class Connection {
...
+  }
+
+  setVersions(versions) {
+    this.versions = versions
   }
 
-  get connected() {
-    return CONNECTED_STATUS.includes(this.connectionStatus)
+  isConnected() {
+    return this.connectionStatus === CONNECTION_STATUS.CONNECTED
```
Nevon, line 159: I think the reason we has `isConnected() === true` when the connection status is `disconnecting` is to allow inflight requests to drain while not accepting new requests. Is this handled differently somehow now to make this not necessary?

src/consumer/fetcher.js
```diff
@@ -0,0 +1,58 @@
...
+        await workerQueue.push(...batches)
+      }
+    } catch (error) {
+      isRunning = false
+      emitter.emit('end')
+      throw error
+    }
+
+    if (isRunning) {
+      return start()
```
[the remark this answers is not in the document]
priitkaard: Embarrassing, TIL node.js doesn't support TCO (anymore).

I personally don't like the onCrash solution for Runner, because the callback can be called from multiple places and it's not clear how the errors are merged. But I guess it's the only way to go in that case.

src/consumer/fetcher.spec.js
priitkaard: this isn't a public API. 
fetchManager.start() is only used in Runner, which awaits it until an error occurs or fetchManager.stop() is called.
but as I mentioned already, TCO is not supported, so I will have to refactor it to a loop.

src/consumer/index.js
priitkaard: `process.on('SIGTERM', () => consumer.disconnect())`

Send SIGTERM signal twice and consumerGroup.leave() will reopen the connection, which will increase the graceful shutdown time.

priitkaard:
Updated diagram, how concurrency is managed:
kafkajs - Page 1

src/consumer/fetcher.js
```diff
@@ -0,0 +1,44 @@
...
+
+  const getWorkerQueue = () => workerQueue
+
+  const start = async () => {
+    if (isRunning) return
+    isRunning = true
+
+    while (isRunning) {
+      try {
+        const batches = await fetch(nodeId)
```
Nevon, line 24: Should we interrupt here if `!isRunning`, so that we don't push the batches to the workerQueue?

Nevon:
Cool! Hopefully I get a chance to run another test with tracing to see that everything is working as expected, but it's looking good.

@julienvincent Looking forward to seeing the results!

src/consumer/fetcher.js
priitkaard: Runner will return [] anyway when it's stopping, but I will add it here as well for clarity.

Nevon:
Luckily I still had my instrumentation collection sitting around in the working tree, so I was able to do another run. Same parameters as before - 6 partitions, a single consumer, `partitionsConsumedConcurrently: 3`, 3 brokers.

Screenshot 2022-02-22 at 08 59 57

[Interactive view](https://jsfiddle.net/j6p0o18f/)

I think it looks good. The processing looks great. We can see that it's constantly processing 3 partitions in parallel. Since there are 6 partitions, the gaps between fetches is because if a worker queue is containing data for 2 partitions we need to drain both before executing the next fetch.

Nevon:
I did a highly unscientific performance test.

3 brokers, 1 consumer, 6 partitions, `partitionsConsumedConcurrently: 3`, 1 million messages. `eachMessage` doing nothing except incrementing a counter:
- `master` -> 9658ms
- `priitkaard:master` -> 8637ms

Same as above, but with a 100ms sleep in `eachBatch`:
- `master` -> 8856ms
- `priitkaard:master` -> 7919ms

The results seemed pretty much the same on multiple runs. I'd be curious to see some results from real-world workloads.

priitkaard:
@Nevon Try to unbalance the sleep between partitions. That's where the delay comes.
In my initial PR description, I also added two scripts demonstrating the issue.

Nevon:
Changed the sleep to be `100 * batch.partition` and re-ran with the same configuration.

* `master` -> 10824ms
* `priitkaard:master` -> 7798ms

mark-boichuk-allianceblock:
Guys, you are doing an amazing job!
Can't wait to see this in prod.

mark-boichuk-allianceblock:
@priitkaard  By the way, does your changes fix the problem with double processing messages on group rebalance?

priitkaard:
> @priitkaard By the way, does your changes fix the problem with double processing messages on group rebalance?

@mark-boichuk-allianceblock 
Good question. I think rebalancing starts only with the heartbeat.
Offsets are now committed after processing each batch, before the heartbeat, so that should reduce it.
But if you use heartbeat inside the eachBatch, it might still be an issue.

If you have time, you can test it out and let me know as well.

mark-boichuk-allianceblock:
@priitkaard Take a look on [these logs](https://github.com/tulios/kafkajs/issues/1300#issuecomment-1047871990)

Seems even after offset commit it double processes same message

priitkaard:
@mark-boichuk-allianceblock 
FYI, I'm able to reproduce it and working on a fix.

priitkaard:
@mark-boichuk-allianceblock 

I now autoCommit offsets on rebalance before rejoining the group.
Created a regressions test for the described issue as well: https://github.com/tulios/kafkajs/blob/2b4e62be4738ace804c81c3acc44d44d4182e36e/src/consumer/__tests__/index.spec.js#L72

Let me know if it solved your issue.

mark-boichuk-allianceblock:
Hmmm, seems that half of the messages are not processed (1 of 2 partitions)

mark-boichuk-allianceblock:
Now it works fine, will try to test it a bit more. Thanks!

mark-boichuk-allianceblock:
We need to check how it will behave on broker outage.

julienvincent:
@Nevon @priitkaard Regarding performance testing:

The biggest difference can be seen when measuring roundtrip latency on slow or idle topics. This is the primary metric which I'm interested in as we have quite a few slow or mostly idle topics where, when a message is produced, we need it to be picked up and processed as soon as possible.

On master, if using a high `maxWaitTimeInMs` (lets say 5s), there can be up to `maxWaitTimeInMs` latency between a message being put onto the topic and it being processed. The only way to get around this is to reduce `maxWaitTimeInMs`, but this incurs quite a large cost on resource usage and network ingress/egress - especially when there are a lot of processes running under this condition.

The script I posted higher up (https://github.com/tulios/kafkajs/pull/1258#issuecomment-1010079687) is a really hacky test that demonstrates this property quite well. The difference between this branch and master is very stark.

Running this locally against a Kafka cluster with 12 brokers and a topic with 10 partitions:

<details>
  <summary>Example run from master</summary>

```
RoundRobinAssigner","duration":453}
latency: 376
latency: 275
latency: 174
latency: 10365latency: 10265
latency: 10162
latency: 10062
latency: 9962
latency: 9861
latency: 9759
latency: 9658
latency: 9556
latency: 9457
latency: 9356
latency: 9255
latency: 9155
latency: 9053
latency: 8952
latency: 8851
latency: 8749
latency: 8648
latency: 8549
latency: 8448
latency: 8348
latency: 8246
latency: 8145
latency: 8045
latency: 7944
latency: 7842
latency: 7740
latency: 7640
latency: 7538
latency: 7437
latency: 7336
latency: 7235
latency: 7135
latency: 7034
latency: 6933
latency: 6832
latency: 6732
latency: 6632
latency: 6530
latency: 6428
latency: 6327
latency: 6226
latency: 6124
latency: 6022
latency: 5921
latency: 5820
latency: 5721
latency: 5619
latency: 5518
latency: 5417
latency: 5316
latency: 5216
latency: 5115
latency: 5014
latency: 4914
latency: 4813
latency: 4711
latency: 4610
latency: 4509
latency: 4408
latency: 4307
latency: 4207
latency: 4108
latency: 4007
latency: 3905
latency: 3805
latency: 3704
latency: 3603
latency: 3503
latency: 3401
latency: 3300
latency: 3200
latency: 3100
latency: 2999
latency: 2899
latency: 2799
latency: 2698
latency: 2598
latency: 2496
latency: 2395
latency: 2294
latency: 2193
latency: 2092
latency: 1992
latency: 1891
latency: 1789
latency: 1688
latency: 1586
latency: 1485
latency: 1383
latency: 1283
latency: 1182
latency: 1081
latency: 980
latency: 879
latency: 778
latency: 676
latency: 575
latency: 474
latency: 372
latency: 271
latency: 170
```
</details>

<details>
  <summary>Example run from priitkaard:master</summary>

```
latency: 93
latency: 51
latency: 62
latency: 101
latency: 43
latency: 59
latency: 21
latency: 16
latency: 26
latency: 189
latency: 147
latency: 94
latency: 28
latency: 42
latency: 54
latency: 27
latency: 14
latency: 25
latency: 21
latency: 19
latency: 103
latency: 27
latency: 36
latency: 26
latency: 85
latency: 49
latency: 20
latency: 23
latency: 15
latency: 39
latency: 143
latency: 63
latency: 18
latency: 79
latency: 116
latency: 55
latency: 18
latency: 23
latency: 20
latency: 28
latency: 164
latency: 82
latency: 27
latency: 53
latency: 18
latency: 63
latency: 16
latency: 34
latency: 36
latency: 19
```
</details>

Nevon:
Interesting, so it makes sense why it would work great with this PR, since each broker essentially has its own little fetch loop, so as soon as the consumer is done with the batches that are currently in the queue, it'll fire off new fetch requests and start processing the results as soon as one of them returns some batches (which, assuming you have a low `minBytes` would be very quick).

But why would there have been so much latency previously? Let's simplify things and say we have 3 partitions and 2 brokers. The fetch loop would have made requests to both brokers. As soon as `minBytes` was satisfied, the broker would return. As soon as either broker returned batches, the responses would start to be processed, so no latency there. Is it because we wouldn't start the _next_ fetches until we're done processing the current fetch loop? So say we immediately got back some messages on partition 1, but there's nothing on partition 2 so it's still waiting up until `maxWaitTimeInMs`, before we can make another request to get more data on partition 1?

julienvincent:
> Is it because we wouldn't start the next fetches until we're done processing the current fetch loop?

Yes, correct. This latency is mostly observed when there is an imbalance in rate of production between partitions. In my test script I am explicitly producing to a single partition to exaggerate the effect:

```ts
producer.send({
  topic: "test-latency",

  messages: [
    {
      partition: 0,
      value: id,
    },
  ],
})
```

In the real world, this effect is caused when using the message key to pick the partition (as apposed to round-robin) - resulting in quite a lot of imbalance between the partitions. As this is the most common way to pick the partition in our systems, we see this effect all over the place.

priitkaard:
Yep, exactly. 
Some partitions can have a much higher input rate when they are partitioned by a certain key, while some can be idle.
A workaround so far was to reduce maxWaitTimeInMs to a relatively small period, 10-100ms. 
However, with a higher number of consumers, the request rate goes up a lot and it's more efficient to execute long-polls instead of constantly bombarding all of the brokers.

Nevon:
Yeah that makes perfect sense.

@julienvincent: Are you planning to test this branch in your systems? I haven't yet finished setting up a test system at work, so I can't really try it out in the real world until I've got that up and running. I have enough confidence in our tests to trust that this works fine during normal operation, but I'd be a lot more confident if I knew it had been running fine with a 10 broker cluster for a few days.

priitkaard:
I’m running it over the weekend on a POC environment under production-like load. 
Next week will deploy the latest version on some production services as well.

julienvincent:
My preliminary testing is done and it seems to function as expected. I will be setting up a few services on our production environment today and tomorrow to run this branch and will let you know if I run into any problems.

src/consumer/consumerGroup.js
```diff
@@ -406,162 +435,110 @@ module.exports = class ConsumerGroup {
...
+              if (currentPreferredReadReplica !== preferredReadReplica) {
+                this.logger.info(`Preferred read replica is now ${preferredReadReplica}`, {
+                  groupId: this.groupId,
+                  memberId: this.memberId,
+                  topic: topicName,
+                  partition,
+                })
               }
+              preferredReadReplicas[partition] = {
+                nodeId: preferredReadReplica,
```
t-d-d, line 526: Once we have assigned the partition to a different broker, isn't it the case that a fetch for that broker could now execute before the batch from this broker has been processed and resolved?

t-d-d:
Is the worker/ worker queue mechanism here equivalent to the old utils/concurrency ?  The new code looks ok but I was just wondering if there is a good reason to replace it.

priitkaard:
> Is the worker/ worker queue mechanism here equivalent to the old utils/concurrency ? The new code looks ok but I was just wondering if there is a good reason to replace it.

With the previous concurrency manager, it was not possible to pause the workers or in any way handle the lifecycle of a fetch response. 
Maybe it could be adjusted now, seeing where this PR came after multiple stages of refactoring. But I find the new solution a bit more readable and prefer not to go back to the concurrency + barrier logic.

src/consumer/consumerGroup.js
priitkaard: I will focus on this at the end of the week. 
On the first look, you might be right, though I'm not sure how often this could trigger, if at all, without rebalance.

Nevon: Just to see that I'm on the same page - the scenario you're describing is one where we have partition 1 assigned to broker 1. We fetch a batch from broker 1 and start to process it. While that's being processed, a Kafka admin goes and reassigns partition 1 to broker 2. We fetch from broker 2, and now we're basically processing the same messages in parallel, potentially causing some really strange behavior. Correct?

This does sound plausible. I haven't tried it to see what would happen, but I also can't find any information about how partition reassignment would affect consumer groups (other than if the leader changes), so I'm guessing it's something that needs to be handled client-side.

If so, then it seems a reasonable assertion to make that only one worker can process batches for a particular partition at any given time. Not sure where to enforce that though - if it would be in the worker by making the workers have to "check out" the partition they want to process messages for (then how do we deal with the already processed messages when the second worker picks them up from the second batch?), or if it can be enforced higher up somehow.

Nevon: Took a stab at improving this in https://github.com/tulios/kafkajs/commit/1c4b02e1dc477c380e5a2e4705f05491f14096ad I can't figure out how to create a PR towards your branch (I guess it's unusual to have the upstream create a PR towards a fork...), but you can see the diff and do a manual merge if you think it makes sense.

I still don't know if this is the right approach, or if we should do something more like the rebalancing when we detect this happening. Even with this, there's still no guarantee that worker 1 won't finish processing a batch for partition 1 just before worker 2 starts to process the same batch from partition 1, in which case they'll still be processing the same messages. The only way I see to avoid that is to look at the resolved offset and skip offsets before the currently resolved one.

Nevon: Btw, I asked the data streaming operations team at work if they knew what would happen in this scenario. They tested it out and verified that reassigning a partition to a different broker does not cause a consumer group rebalance, so it is something we have to handle on the client side.

Nevon: And for those reading along at home, the solution we went with in https://github.com/priitkaard/kafkajs/pull/2 was to have the FetchManager keep track of which fetcher is assigned to a particular topic-partition at any given time. When a fetcher receives a response, it filters out batches for any topic-partitions it is not currently assigned to. In the scenario we've discussed, this would mean that the second fetcher (which is now fetching from the newly assigned broker) would filter out any topic-partitions that the fetcher processing batches from the previously assigned broker is currently processing or has queued up.

A more elegant solution might be to avoid fetching the topic-partition at all, if it's already assigned to a different fetcher, instead of filtering out the results. But this scenario should be quite rare, so maybe the juice is not worth the squeeze.

Nevon:
Would love to get this merged unless there are any more issues that appear. @priitkaard / @julienvincent: Let me know your experiences running the latest version of this branch whenever you've had a chance to try it out.

priitkaard:
We've been running this in production for around a week now and haven't discovered any additional issues so far.

Nevon:
> haven't discovered any additional issues so far.
> ...
>  EDIT: Your latest change is running without problems in a POC environment at the moment.

I was just about to say that that was before I pushed some commits into this PR. Now all bets are off 😁

Nevon:
Fixes #1300

Nevon (approved):

julienvincent:
We have not seen any out-of-the-ordinary issues with our deployments running this code, and performance + latency has been very good - as expected!

Nevon:
Alright, this has been marinating for long enough. Let's merge it so we can get a beta release out there and hopefully get some more people testing it out.

Nevon:
kafkajs@1.17.0-beta.5 is now out with this change.

Huge thanks to @priitkaard for his patience and dilligence! 🎉  While this change took a while to land, and was quite a bit of work for everyone involved, it was one of the more enjoyable PRs I've been part of in quite a while.

Thanks also to @julienvincent for his testing efforts, and @t-d-d for helping out with the review! 👏

PR #1261, opened 2022-01-05
exposed `assigned` method from consumer.

anakshiant:
**ISSUE**
The issue was as of now `consumer` can not know which partitions has been assigned to it.

**SOLUTION**
`ConsumerGroup` class already had `assigned` method which was not exposed from `consumer` exposed methods. 
- Created a method `assigned` in consumer
- Used `ConsumerGroup.assigned` method
- exported `assigned` method from consumer

### DIFF
--- a/src/consumer/index.js
+++ b/src/consumer/index.js
@@ -494,11 +494,24 @@ module.exports = ({
   }
 
   /**
+   * Returns the list assignment for the consumer
+   *
+   * @type {import("../../types").Consumer["assigned"]}
+   */
+  const assigned = () => {
+    if (!consumerGroup) {
+      return []
+    }
+    return consumerGroup.assigned()
+  }
+
+  /**
    * @return {Object} logger
    */
   const getLogger = () => logger
 
   return {
+    assigned,
     connect,
     disconnect,
     subscribe,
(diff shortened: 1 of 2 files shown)


Nevon:
Do you have a use-case in mind for this method?

I have a hard time seeing a situation in which I would synchronously want to know what partitions are assigned at that moment. If I wanted to keep track of assignment changes, for example to record metrics, I would rather subscribe to the `GROUP_JOIN` event so that I'm only notified when the assignments have actually changed (or may have changed at least). The event already contains the member assignment:

```js
consumer.on(consumer.events.GROUP_JOIN, event => {
  console.log(event.payload.memberAssignment) // { [topicName: string]: string[] }
})
```

Nevon:
Closing this as inactive.

Zacaria:
Hi I'm actually needing this to synchronize topics state with an external database.

I'll have a `setInterval` which will poll the state of all the assigned topics of the `consumer` instance and check the desired state in the db.

This will allow us to have a UI which will pause and resume topics.

This is especially useful when I subscribe to topics using a wildcard, I nowhere have the entire list.

It would be great if this could be included in a release soon.

Meanwhile I'll use a fork 🍴 😄

Issue #1263, opened 2022-01-05
Using third party libraries


Lehks:
I am connecting to a Kafka service that requires additional libraries (these are kakfa-oauth-common, kafka-oauth-client, keycloak-common and keycloak-core). In the tutorial that I followed, these libraries would be downloaded and put into a certain directory next to the kafka executable.

Is it possible to use these libraries with kafkajs?

The guide I followed was this one: [https://developer.mercedes-benz.com/products/connect_your_business/docs](https://developer.mercedes-benz.com/products/connect_your_business/docs). Look for the caption "Connect to the Push API & Consume Vehicle Signals".

Nevon:
Those libraries are for the Java Kafka client, so no, you can't use them in NodeJS. Looking at the Mercedes documentation, it looks like they're using the oauthbearer mechanism, which we do support, but you have to write the integration towards their oauth server yourself (or use some generic oauth integration library). The documentation you linked to links to [an example repo integration with a Python client](https://github.com/Daimler/kafka-integration-samples/blob/master/kafka-sample/python/kafka-oauth-demo.py#L51-L64), which does something similar.

[See our documentation](https://kafka.js.org/docs/configuration#oauthbearer-example) for a generic example of using the oauthbearer mechanism in KafkaJS. The bit that's missing is the part where you integrate with the oauth server to get the token.

PR #1264, opened 2022-01-05
Add example for oauthbearer implementation

Nevon:
The actual integration with the oauth server is beyond our responsibility, but the docs are currently _very_ sparse on detail. With Typescript you can kinda figure out how to do it, but if you're using Javascript I'd be very impressed if you figured out what to do. 

Related to #1263

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -132,7 +132,77 @@ The OAuth bearer token must be an object with properties value and
 request.
 
 The implementation of the oauthBearerProvider must take care that tokens are
-reused and refreshed when appropriate.
+reused and refreshed when appropriate. An example implementation using
+[`simple-oauth2`](https://github.com/lelylan/simple-oauth2) would look something
+like the following:
+
+```ts
+import { AccessToken, ClientCredentials } from 'simple-oauth2'
+interface OauthBearerProviderOptions {
+  clientId: string;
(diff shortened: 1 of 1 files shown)


Nevon:
@uwburn: I based this example on your example from #674. If you're still around, I'd love to have you take a quick look at the example, since I just wrote it from the type definitions and haven't actually tested it against an oauth server.

uwburn:
Yes, i'm still around and still using kafkajs, i will take a look at it asap.

uwburn (changes requested):
LGTM, the only change i strongly recommend it's wrapping all the code inside `refreshToken` function inside a try-catch, and reschedule the refresh in the catch block, otherwise any error would permanently stop the token refresh.

```javascript
 async function refreshToken() {
    try {
        if (accessToken == null) {
          accessToken = await client.getToken({})
        }

        if (accessToken.expired()) {
          accessToken = await accessToken.refresh()
        } else {
          const refreshIn = accessToken.token.expires_in * 1000 - options.refreshThreshold;

          setTimeout(() => {
            tokenPromise = refreshToken()
          }, refreshIn);
        }

        return accessToken.token.access_token;
    } catch(err) {
        setTimeout(() => {
          tokenPromise = refreshToken();
        }, options.retryDelay);
        return null;
    }
}
```

PR #1266, opened 2022-01-10
Support to a different test environment

helio-frota:
This commit changes the code to make easier to run the integration
tests against Red Hat managed Kafka service.


Nevon:
Hey Helio, and thanks for taking a look at the RedHat Managed Kafka compatibility!

This code is getting rather messy now from all these conditionals everywhere, leading to some confusing stuff like the `createConnectionBuilder` function now using SASL connection options or not depending on whether or not you've defined a host override. I would suggest we re-think this a bit.

Here's my understanding of the requirements:
* The tests need to be able to get connection details for using different authentication mechanisms and options
* The hosts and ports for those different options are not necessarily different (the RedHat Kafka case)
* There can be 1 or more hosts/ports for any given configuration
* The code should not have to know whether we're connecting to RedHat Kafka or some other Kafka cluster

My suggestion therefore is that we define a few different environment variables:
* `BOOTSTRAP_BROKERS_PLAIN`
* `BOOTSTRAP_BROKERS_SASL`
* `BOOTSTRAP_BROKER_SSL`

All of these three would be strings containing both host and port, separated by a `:`. In case of multiple brokers, they would be separated by a `,`. For example `BROKERS_PLAIN="localhost:9092,localhost:9095,localhost:9098". In the test script, we just parse that string and use those brokers where appropriate. When running with RedHat Kafka, you'd just define a single broker, and in our usual tests we'd define 3. Since 99% of the time the tests are run against the local cluster defined in the docker-compose, we should default these values to the current default:

```ts
const bootstrapBrokersPlain = process.env.BOOTSTRAP_BROKERS_PLAIN || 'localhost:9092,localhost:9095,localhost:9098'
```

Similarly with the username and passwords, we can simply have an override per mechanism and default to the current values:

* `SASL_PLAIN_USERNAME` & `SASL_PLAIN_PASSWORD`
* `SASL_SCRAM_256_USERNAME` & `SASL_SCRAM_256_USERNAME`
* `SASL_SCRAM_512_USERNAME` & `SASL_SCRAM_512_USERNAME`

There shouldn't be a need to change functions like `createCluster` or `createTopic` as long as the functions for generating appropriate connection options for each authentication mechanism can handle a dynamic number of brokers.

helio-frota:
@Nevon thank you for the feedback and I agree/understand about the messy part.

> When running with RedHat Kafka, you'd just define a single broker, and in our usual tests we'd define 3.

Yeah, we need to point/use only the bootstrap address.

The issue is that we have some restrictions with authentication and connection mechanisms.
Another restriction is related to the replication factor that should be `3` to avoid topic creation errors.
As feedback, we have 85%+ tests passing. And of course, we skipped some tests that are specific for the KafkaJS docker-compose environment and/or related to other restrictions. Also some code changes (more conditionals and etc).

So because of that maybe parts of the code will make no sense for KafkaJS and we understand that.
For example, I need to add another conditional check to change the replication factor to 3.

So I'm trying to find a good balance to make it easy to run the tests with minimal changes on both sides

Nevon:
Sure, I'm just trying to find a way that we can avoid having loads of conditionals everywhere, and can generalize those needed options so that they just become different configurations for running the tests. So for example the default replication factor we can just have an environment variable like `DEFAULT_REPLICATION_FACTOR` that we default to 1, but when you run the tests against RedHat Kafka, you run it with 3. Same with the brokers, where we would normally run it with 3 brokers and you run it with one. The code would be the same - just the configuration changes.

helio-frota:
> I'm just trying to find a way that we can avoid having loads of conditionals everywhere.

> So for example the default replication factor we can just have an environment variable like DEFAULT_REPLICATION_FACTOR that we default to 1, but when you run the tests against RedHat Kafka, you run it with 3.

:+1:  totally agree.

helio-frota:
@Nevon hi, since we have other small parts that need to be changed, I updated the PR to make an initial focus on the env vars only on this PR. :+1:

helio-frota:
@Nevon when you have time, any feedback would be good :+1:

PR #1274, opened 2022-01-25
KafkaJSNumberOfRetriesExceeded should reflect the original retry-ability

### LINKED ISSUE #1271: NonRetryable errors cause the consumer to restart, causing a loop.
**Describe the bug**
When an error occurs in consuming a message from a topic and the error is not handled it bubbles up to the
retry/index.js and will return a:
```  reject(new KafkaJSNumberOfRetriesExceeded(e, { retryCount, retryTime }))```

However in the consumer/index.js onCrash there is a check if the server should restart by checking if the error is retryable.

That check is:
```const isErrorRetriable = e.name === 'KafkaJSNumberOfRetriesExceeded' || e.retriable === true```

Hoever that name is of a NonRetryableError! Causing the consumer to duly restart, rinse repeat..

Possible hacky fix: additionally check for # retries > 0

JanGroot:
The KafkaJSNumberOfRetriesExceed now reflects if the original error was retryable.
If so: The onCrash will restart the consumer (regarding shouldRestartOnFailure)
If not: The onCrash will not.

closes #1271

### DIFF
--- a/src/admin/__tests__/deleteTopicRecords.spec.js
+++ b/src/admin/__tests__/deleteTopicRecords.spec.js
@@ -338,7 +338,7 @@ describe('Admin > deleteTopicRecords', () => {
     }
 
     expect(error).toBeDefined()
-    expect(error.name).toBe('KafkaJSNumberOfRetriesExceeded')
+    expect(error.name).toBe('KafkaJSNonRetriableError')
     expect(error.originalError.name).toBe('KafkaJSDeleteTopicRecordsError')
     expect(error.originalError.retriable).toBe(false)
     expect(error.originalError.partitions).toEqual([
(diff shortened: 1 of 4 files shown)


Nevon:
The intention is that the consumer should always retry if the reason for the crash was due to some temporary error that we could recover from by tossing away the current state and restarting. That's why we restart if the error is either retriable or KafkaJSNumberOfRetriesExceeded.

So the intention of this change, to make it so that we restart only if the error that caused us to run out of retries was a retriable one, makes sense. It's just unfortunate that the way to do it is to make `KafkaJSNumberOfRetriesExceeded` itself potentially retriable. At least to me, it seems unintuitive that an error saying that we're out of retries is retriable. It's kind of repurposing the property to mean something different than what it means for all other errors.

What do you think about keeping `KafkaJSNumberOfRetriesExceeded` as non-retriable, but change the restart check to `e.retriable || (e.name === 'KafkaJSNumberOfRetriesExceeded' && e.originalError.retriable)`?

JanGroot:
Good point, above anything, it should be clear what the intention is. I have another idea, I'll update the PR.

JanGroot:
New and more clear approach:
If the error was not retriable to begin with, reject with KafkaJSNonRetriableError.

closes #1271

Nevon (approved):

Issue #1273, opened 2022-01-25
Document the clientId a little better


justinmchase:
**Is your feature request related to a problem? Please describe.**
The docs just show `'my-app'` for the clientId everywhere. It would be immensely helpful to have a little blurb describing what a good clientId value should be. For example, if I have a horizontally scaling api, each with a kafka client, should they all have distinct clientId values or should they all have the same clientId? I'm having a hard time determining what is more correct.

**Describe the solution you'd like**
Just a paragraph in the docs somewhere with the keyword `clientId` linked to it, since that was what I was searching for.

**Additional context**
Perhaps a link to any relevant kafka documentation as well, I am having a hard time finding it though I'm assuming its there somewhere. This is just where I started the search and it would probably help others as well.

Nevon:
Sounds good! My understanding is that the client id is only used for client quotas and for request logging inside Kafka.

[The Kafka documentation says](https://kafka.apache.org/documentation/#design_quotasgroups):

> **Client-id is a logical grouping of clients with a meaningful name chosen by the client application.** The tuple (user, client-id) defines a secure logical group of clients that share both user principal and client-id. Quotas can be applied to (user, client-id), user or client-id groups.

Based on this, I would use the clientId to represent "my application". Each instance of that application within the same environment would use the same client id, and I would not use the same client id in multiple different applications. Hence why the example is "my-app" - something that uniquely identifies my application.

I think it can deserve its own heading next to "Broker Discovery" [here](https://kafka.js.org/docs/configuration#broker-discovery). Want to take a crack at it?

justinmchase:
Perfect, thank you! Your interpretation of that documentation is the same as how I am understanding it.

I'm happy to make a PR, you did all the heavy lifting though 😄

PR #1276, opened 2022-01-25
Add clientId documentation

justinmchase:
Adds documentation for the `clientId` field.

* Fixes https://github.com/tulios/kafkajs/issues/1273

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -16,6 +16,8 @@ const kafka = new Kafka({
 ```
 
 ## Client Id
+A logical identifier of an application. Can be used by brokers to apply quotas or trace requests to a specific application. Example: `booking-events-processor`.
+
 The [kafka documentation](https://kafka.apache.org/documentation/#design_quotasgroups) describes the `clientId` as:
 
 > Client-id is a logical grouping of clients with a meaningful name chosen by the client application. The tuple (user, client-id) defines a secure logical group of clients that share both user principal and client-id. Quotas can be applied to (user, client-id), user or client-id groups.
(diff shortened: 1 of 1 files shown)


docs/Configuration.md
```diff
@@ -15,6 +15,19 @@ const kafka = new Kafka({
...
+
+> Client-id is a logical grouping of clients with a meaningful name chosen by the client application. The tuple (user, client-id) defines a secure logical group of clients that share both user principal and client-id. Quotas can be applied to (user, client-id), user or client-id groups.
+
+It [also says](https://kafka.apache.org/documentation/#producerconfigs_client.id):
+
+> `client.id`
+>
+> An id string to pass to the server when making requests. The purpose of this is to be able to track the source of requests beyond just ip/port by allowing a logical application name to be included in server-side request logging.
+
+Therefore the `clientId` should be shared across multiple instances in a cluster or horizontally scaled application, but distinct for each application. Consumer group instances will also have [`clientHost`](../types/index.d.ts#L715) and [`memberId`](../types/index.d.ts#L717) fields which can be used to distinguish between instances with the same [`clientId`](../types/index.d.ts#L715).
```
Nevon, line 29: Those links won't resolve, since they are relative links to files in the repo, not to pages on the website.

Either way, I would remove this part, since it's only relevant for people operating Kafka clusters, not users of KafkaJS. If you're operating a Kafka cluster, you already know that you identify logical applications by clientId+principal, and individual instances by `memberId` or ip.

Our users just need to know what a good value to set is, so maybe try to condense this a bit. Something like "A logical identifier of an application. Can be used by brokers to apply quotas or trace requests to a specific application. Example: `booking-events-processor`". Short and sweet 😃

justinmchase: Ok, sounds good. I added your phrase to the top verbatim and I left in the first sentence of this paragraph as that seems to be the sentence I was wishing was there. I think the reader can figure it out by reading and comprehending the quoted passages but it should help to spell it out plainly.

Issue #691, opened 2020-04-06
ERROR: The group is rebalancing, so a rejoin is needed


oemergenc:
Hi,

i need some help with the error message from the title. I read this issue, which seems to be related:

https://github.com/tulios/kafkajs/issues/226

I also read the documentation here, which also seems to be related:
https://kafka.js.org/docs/faq#what-does-it-mean-to-get-rebalance_in_progress-errors

However, it doesnt matter what i try the errors are remaining. It only occures when i redeploy my application and unfortunately it always triggers our alerting system, which is annoying.

Here is the log from my application:

```
{"loglevel":"error","msg":"ERROR [Connection] Response Heartbeat(key: 12, version: 1) {\"timestamp\":\"2020-04-06T10:32:41.900Z\",\"logger\":\"kafkajs\",\"broker\":\"kafka-2:9093\",\"clientId\":\"my-custom-client-id-server\",\"error\":\"The group is rebalancing, so a rejoin is needed\",\"correlationId\":16,\"size\":10}"}
{"loglevel":"error","msg":"ERROR [Runner] The group is rebalancing, re-joining {\"timestamp\":\"2020-04-06T10:32:41.901Z\",\"logger\":\"kafkajs\",\"groupId\":\"my-custom-client-id-v001-server\",\"memberId\":\"my-custom-client-id-server-bea7349c-32f2-4855-92a2-c2549f172082\",\"error\":\"The group is rebalancing, so a rejoin is needed\",\"retryCount\":0,\"retryTime\":3560}"}
{"loglevel":"error","msg":"ERROR [Connection] Response SyncGroup(key: 14, version: 1) {\"timestamp\":\"2020-04-06T10:33:41.906Z\",\"logger\":\"kafkajs\",\"broker\":\"kafka-2:9093\",\"clientId\":\"my-custom-client-id-server\",\"error\":\"The group is rebalancing, so a rejoin is needed\",\"correlationId\":18,\"size\":14}"}
[...]

PR #1279, opened 2022-02-03
Log warn instead of error when rebalancing

MelvinVermeer:
This PR addresses an old issue (https://github.com/tulios/kafkajs/issues/691)

I would suggest that it is easier for devops teams to add extra monitoring for this `warn` then it would be to ignore `error` entries.

If you are open to extending the KafkaConfig, we could also enable this to be set as config setting. Then we could let error be the default, and let users Opt-In to a `warn`.

### DIFF
--- a/src/consumer/runner.js
+++ b/src/consumer/runner.js
@@ -404,7 +404,7 @@ module.exports = class Runner extends EventEmitter {
         }
 
         if (isRebalancing(e)) {
-          this.logger.error('The group is rebalancing, re-joining', {
+          this.logger.warn('The group is rebalancing, re-joining', {
             groupId: this.consumerGroup.groupId,
             memberId: this.consumerGroup.memberId,
             error: e.message,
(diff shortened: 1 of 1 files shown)


Nevon:
> I would suggest that it is easier for devops teams to add extra monitoring for this warn then it would be to ignore error entries.

I still strongly believe that this should be an error, and that no one should base their observability strategy around alerting on logs, but rather metrics around their SLOs (i.e. is the consumer doing its job). However, it's clear from the the fact that this comes up every now and again, with quite a bit of support in the original issue, that some people are using logs to drive alerting and that's not something I'm going to be able to change, so I guess I'll disagree and commit.

> If you are open to extending the KafkaConfig, we could also enable this to be set as config setting. Then we could let error be the default, and let users Opt-In to a warn.

I basically see this as a capability we already provide via the custom logger, so I wouldn't want to bloat the options for such a small and specific change. The only reason I would see for introducing this as an opt-in would be to not break existing monitors people have set up where for whatever reason _want_ to alert on rebalancing (and are using the logs for this, instead of emitting metrics from the instrumentation events), but then I've consistently said that you shouldn't base alerting on logs, so this will just be one more example of why.

MelvinVermeer:
@Nevon Thanks for the clear explanation

PR #1283, opened 2022-02-07
Lint tests

Nevon:
#1256 had a focused test that would have gotten merged into master if I didn't happen to spot it, which would have disabled all other tests in that file.

This PR adds linting rules for the tests. I based the rules around the `jest/recommended` configuration, but with a few changes to fit how we use tests. Mostly I didn't include a few rules that prevent conditional tests, which we use a fair bit to conditionally enable tests depending on Kafka versions or authentication methods.


julienvincent:
My bad 🤦

PR #1313, opened 2022-03-12
Make consumer.subscribe accept an array of topics

benvan:
**Feature:** `consumer.subscribe` now accepts String, String[] or RegExp

**Motivation**: At work, many of our consumers subscribe to a handful of related topics. Reducing the number of `consumer.subscribe` calls in the proposed manner significantly reduces our consumer startup time (50%)

**Justification**: Using RegExp is already essentially supplying a string array, just indirectly. If you know the topics you want to subscribe to, this avoids you needing to craft a regex which is built from those topic names (and the cost of the subsequent call to the broker to download the topic list)

Note: This PR does not have an associated issue - if this is a necessary requisite then apologies in advance and I'm happy to oblige


Nevon:
Makes sense. I would just like to deprecate the `topic` argument and instead have `topics: string[] | RegExp`. We can still accept the `topic` argument for backwards compatibility, but throw an error if both `topic` and `topics` are provided. Feels weird to have an argument called `topic` that takes an array.

benvan:
Definitely agree re. `topics` vs `topic`. I'll submit an update, hopefully this weekend. 

Regarding the plural regex/string question - I'll make it take multiple regexes unless you have any objections

Nevon:
Originally I was thinking `string[] | RegExp[]` because mixed-type arrays always make me feel like I must be doing something wrong, but in this case I agree it makes sense to do `(string | RegExp)[]`

Nevon:
I went ahead and updated this with the interface we discussed.

benvan:
Thank you @Nevon!

PR #1320, opened 2022-03-22
Adds StickyPartitioner

jmbeach:
This change adds a new partitioner that works more like [the default partitioner in the java client](https://github.com/a0x8o/kafka/blob/master/clients/src/main/java/org/apache/kafka/clients/producer/internals/StickyPartitionCache.java)

The idea is that it tries to send batches of messages that don't have keys to the same partition as a performance optimization.

### DIFF
--- a/src/producer/partitioners/default/partitioner.js
+++ b/src/producer/partitioners/default/partitioner.js
@@ -1,15 +1,8 @@
 const randomBytes = require('./randomBytes')
+const toPositive = require('./toPositive')
 
 // Based on the java client 0.10.2
 // https://github.com/apache/kafka/blob/0.10.2/clients/src/main/java/org/apache/kafka/clients/producer/internals/DefaultPartitioner.java
-
-/**
- * A cheap way to deterministically convert a number to a positive value. When the input is
- * positive, the original value is returned. When the input number is negative, the returned
- * positive value is the original value bit AND against 0x7fffffff which is not its absolutely
- * value.
- */
-const toPositive = x => x & 0x7fffffff
 
 /**
  * The default partitioning strategy:
--- a/src/producer/partitioners/default/toPositive.js
+++ b/src/producer/partitioners/default/toPositive.js
@@ -1,0 +1,7 @@
+/**
+ * A cheap way to deterministically convert a number to a positive value. When the input is
+ * positive, the original value is returned. When the input number is negative, the returned
+ * positive value is the original value bit AND against 0x7fffffff which is not its absolutely
+ * value.
+ */
+module.exports = x => x & 0x7fffffff
--- a/src/producer/partitioners/index.js
+++ b/src/producer/partitioners/index.js
@@ -1,7 +1,9 @@
 const DefaultPartitioner = require('./default')
 const JavaCompatiblePartitioner = require('./defaultJava')
+const StickyPartitioner = require('./sticky')
 
 module.exports = {
   DefaultPartitioner,
   JavaCompatiblePartitioner,
+  StickyPartitioner,
 }
--- a/src/producer/partitioners/sticky/index.js
+++ b/src/producer/partitioners/sticky/index.js
@@ -1,0 +1,4 @@
+const murmur2 = require('../default/murmur2')
+const createStickyPartitioner = require('./stickyPartitioner')
+
+module.exports = createStickyPartitioner(murmur2)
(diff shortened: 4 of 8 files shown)


Nevon:
Since we don't have something equivalent to `linger.ms`, I guess the idea here is that if you were to send X messages in a single `send` call, you'd want all X to go to the same partition (assuming they don't have a key or partition set) so that you only have to make a request to 1 broker rather than potentially all brokers. In clients that do message batching over time, such as the Java client, this is also supposed to "fill up" requests more efficiently - we don't get that property because we don't have a linger mechanism.

**Context:**
* [KIP-480](https://cwiki.apache.org/confluence/display/KAFKA/KIP-480%3A+Sticky+Partitioner) - initial introduction of the UniformStickyPartitioner, which became the default partitioner
* [KIP-794](https://cwiki.apache.org/confluence/display/KAFKA/KIP-794%3A+Strictly+Uniform+Sticky+Partitioner) - Replaces the UniformStickyPartitioner. I highly recommend reading this as well as the linked discussions.

# KIP-794

KIP-794 brings up a lot of issues with the uniform sticky partitioner that makes it neither sticky nor uniform. From my reading, I think those issues are due to the interaction with the linger mechanism. Basically, the switching of the "sticky" partition is triggered from batch creation. If a broker is slow to process a particular partition, that will mean that the "sticky" partition will take longer to be switched. This means more messages will pile up to that partition. So essentially slow partitions get allocated more messages.

The TL;DR of KIP-794 is that they are going to switch to instead having the "sticky" partition be switched every `batch.size` kb. This should lead to a more even distribution across partitions, no matter if a particular partition is slower to be drained, and should make it so that requests end up being "filled" efficiently.

One interesting thing about this is that in order to do this, they are going to need to lift some of the partitioning logic up into the producer instead of keeping it in the partitioner itself. The reason being that the partitioner alone doesn't have enough information to know how many bytes will be sent. I believe we have a similar issue.

# Issues with our approach

In this PR, essentially every message without a key will be produced to the same partition until there is a pause in producing for `EXPIRE_AFTER_MS`. The intention of this is to detect "batches" of messages - something that the producer _knows_ but the partitioner has to _guess_.

Using time for this has some issues. Most importantly, if you have a busy producer, you'll never have `EXPIRE_AFTER_MS` of inactivity and now **every single message without a key will go to the same partition**.

## Alternative approach 👎 

We could imagine using a simple counter instead, and say that we'll switch sticky partition every 100 messages or something. This would ensure a more even distribution across partitions, similar to the round robin partitioner.

However, it then becomes really difficult for the user to predict the resulting batch size (in kb). They unfortunately need to be able to do this currently, since we don't split the messages into multiple requests according to the max batch size. At least with the current round robin they can know that the request size will be `total_size / partitions`.

There's of course also no relationship between the number we pick to roll over the sticky partition and an actual "batch", so it's unclear if it would even have any real benefits. Any approach that relies on "guessing" what constitutes a batch will have the same issue.

---

The more I think about this, the more I think that if we want to be able to have more advanced partitioners, we first need to:

1. Divorce `producer.send`/`producer.sendBatch` from the actual network requests by introducing an accumulator.
2. Have the accumulator create requests based on the max batch size to work around issues where a sticky partitioner would create batches that are too large to be accepted by the broker.
3. As KIP-794 concludes, move the responsibility of the default partitioning up from the partitioner to somewhere in the producer that has access to more information to be able to make better decisions.

That's a significant amount of work, but it would unlock some great potential improvements in throughput through more efficient batching.

Nevon:
There would be _some_ benefit, since it ensures that if you are producing a batch of several messages within a single `producer.send` call, all the messages would end up in the "sticky" partition, so there would only be a single request.

```ts
const messages = [ ... ] // 100 messages without a key

// Here, there would be 1 request with all messages going to the sticky partition
await producer.send({
  topic: 'test-topic',
  messages,
})
```

But if you're doing multiple `producer.send` calls, there's no accumulation of messages across calls, so there wouldn't be any benefit. In fact, there would be a big drawback as now there would be loads of requests all going towards the same broker, instead of being spread out across the cluster.

```ts
// In this scenario, there would be 100 requests made to the same broker. All messages would end up in the "sticky" partition.
for (const message of messages) {
  await producer.send({
    topic: 'test-topic',
    messages: [message]
  })
}

```

Nevon:
Closing this for now. If we end up introducing an accumulation mechanism we can return to this and implement a sticky partitioner on top of that, along the lines of KIP-794

PR #1322, opened 2022-04-05
Document example of using a proxy

Nevon:
In my organization, services are deployed to private subnets with no internet access, requiring egress traffic towards the internet to go through a proxy where traffic filtering rules can be applied. One of the teams needed to connect to a third-party Kafka cluster over the internet, and had quite a challenging time getting it working, so I thought it'd be good to document an example of how to make it work.


justinmchase:
How do you apply the socket factory?

Issue #1013, opened 2021-01-29
List of upcoming breaking changes


Nevon:
For many years now, we have never made a breaking change to KafkaJS (at least not intentionally), which is great for our users in the short term, but in the long term it also means we accumulate cruft and some slightly awkward APIs.

The intention of this issue is to collect all of the things that we have said we want to change in a single major version, so that when we feel that we have accumulated enough things we can actually put together a project to do them.

# Planned breaking changes

* [ ] Never commit from `consumer.seek` (#1012 #395)
* [ ] Remove `ResourceTypes` (#649)
* [ ] Remove `topic` argument from `admin.fetchOffsets` (#992)
* [ ] Remove `admin.getTopicMetadata` (#331)
* [ ] Remove `enforceRequestTimeout` and enable by default (#955)

Please leave a comment if you find more changes that should be on this list.

ankon:
Remove typo-type: https://github.com/tulios/kafkajs/blob/addcd6f8bd1169b053fb633ee11c8cc3b9ae608e/types/index.d.ts#L729-L730

t-d-d:
Switch default producer partitioner to ```JavaCompatiblePartitioner```.
I realise that changing this has issues as it would cause problems for users that upgrade kafkajs version for their current apps..  But as-is there are also issues when interacting with Java (or other) clients.  We actually recently got bitten by this in production after moving to a new Kafka cluster where topics had been mirrored across via MirrorMaker (i.e. the Java producer.)

One option could be to not have a default partitioner but require one to be specified.  Not a nice API but maybe it could be like that through the betas or for one dot version.

Nevon:
Pushed a branch where I removed the `topic: string` argument from `consumer.subscribe`, which is replaced by `topics: string[]` but I'm waffling back and forth on whether or not it's a good change. The amount of complexity to support both is not that much, since we're pretty much just checking that we have one of the two arguments and then converting from `string` to `string[]` if needed.

PR #1339, opened 2022-05-02
Make JavaCompatiblePartitioner new default

Nevon:
The JavaCompatiblePartitioner is renamed DefaultPartitioner. The old DefaultPartitioner becomes LegacyPartitioner.

This is a bit of a spicy one, but the alternatives weren't any better. Originally I removed the default partitioner. This would force people to make an active decision about what partitioner to use, which would be good for people upgrading from an earlier version. Once I started updating tests and docs, I realized how the developer experience becomes really bad for new users. You just want to try out producing some messages, and now you have to learn what the heck a partitioner is? I also see it as a bit futile, since as soon as we release 2.1.0 we'd make the new DefaultPartitioner actually be the default, so anyone going from <2.0.0 to >=2.1.0 would anyway be screwed.

I also considered keeping the `JavaCompatiblePartitioner` export as an alias of the new `DefaultPartitioner`, so that people using the JavaCompatiblePartitioner today wouldn't need to do anything, but I figured we're doing a breaking change anyway so why start adding new deprecated stuff already in 2.0.0.

Related to https://github.com/tulios/kafkajs/issues/1013#issuecomment-778749221


Nevon:
Had a discussion with @tulios and settled on a compromise where we print a warning when configuring a producer without specifying a partitioner. This warning only gets logged once for the lifetime of the application, and can be silenced with an environment variable or by specifying which partitioner to use. In a later release we can remove the warning.

PR #1364, opened 2022-05-17
Provide a `pause()` helper to eachMessage/eachBatch

brianphillips:
This function will take care of pausing (and optionally, resuming) message consumption on the current topic/partition when processing messages either within the `eachMessage` or `eachBatch` handler functions.

### DIFF
--- a/docs/Consuming.md
+++ b/docs/Consuming.md
@@ -102,7 +102,7 @@ await consumer.run({
 * `uncommittedOffsets()` returns all offsets by topic-partition which have not yet been committed.
 * `isRunning()` returns true if consumer is in running state, else it returns false.
 * `isStale()` returns whether the messages in the batch have been rendered stale through some other operation and should be discarded. For example, when calling [`consumer.seek`](#seek) the messages in the batch should be discarded, as they are not at the offset we seeked to.
-* `pause()` can be used to pause the consumer for the current topic/partition. All offsets resolved up to that point will be committed (subject to `eachBatchAutoResolve` and [autoCommit](#auto-commit) settings). If you want to pause in the middle of the batch, you should throw an exception if you do not wish to resolve the current batch's offset or, alternatively, make sure that `eachBatchAutoResolve` is turned off. You can used the returned callback function to resume processing of the relevant partition and topic (i.e. using `setTimeout` or some other asynchronous trigger). See [Pause & Resume](#pause-resume) for more information about this feature.
+* `pause()` can be used to pause the consumer for the current topic/partition. All offsets resolved up to that point will be committed (subject to `eachBatchAutoResolve` and [autoCommit](#auto-commit) settings). If you want to pause in the middle of the batch, you should throw an exception after pausing if you do not wish to resolve the current batch's offset or, alternatively, make sure that `eachBatchAutoResolve` is turned off. You can used the returned callback function to resume processing of the relevant partition and topic (i.e. using `setTimeout` or some other asynchronous trigger). See [Pause & Resume](#pause-resume) for more information about this feature.
 
 ### Example
 
@@ -253,7 +253,7 @@ kafka.consumer({
 
 ## <a name="pause-resume"></a> Pause & Resume
 
-In order to pause and resume consuming from one or more topics, the `Consumer` provides the methods `pause` and `resume`. It also provides the `paused` method to get the list of all paused topics. Note that pausing a topic means that it won't be fetched in the next cycle. You may still receive messages for the topic within the current batch.
+In order to pause and resume consuming from one or more topics, the `Consumer` provides the methods `pause` and `resume`. It also provides the `paused` method to get the list of all paused topics. Note that pausing a topic means that it won't be fetched in the next cycle and subsequent messages within the current batch won't be passed to an `eachMessage` handler.
 
 Calling `pause` with a topic that the consumer is not subscribed to is a no-op, calling `resume` with a topic that is not paused is also a no-op.
 
@@ -330,7 +330,7 @@ await consumer.run({ eachMessage: async ({ topic, message, pause }) => {
 
 Depending on your use case, be sure to throw an exception after `pause()` is called (as shown above) or the current message's offset will be resolved and committed.
 Alternatively, if you want to pause processing of a topic's partition without retrying the current message, no exception needs to be thrown.
-In either case, the `eachMessage` callback will not be called for the current topic/partition until the consumer is resumed. 
+In either case, the `eachMessage` callback will not be called again for the current topic/partition until the consumer is resumed for this topic/partition.
 
 It's possible to access the list of paused topic partitions using the `paused` method.
 
--- a/docs/Consuming.md
+++ b/docs/Consuming.md
@@ -42,7 +42,7 @@ The `eachMessage` handler provides a convenient and easy to use API, feeding you
 
 ```javascript
 await consumer.run({
-    eachMessage: async ({ topic, partition, message, heartbeat }) => {
+    eachMessage: async ({ topic, partition, message, heartbeat, pause }) => {
         console.log({
             key: message.key.toString(),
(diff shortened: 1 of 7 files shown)


docs/Consuming.md
```diff
@@ -100,6 +101,7 @@ await consumer.run({
 * `uncommittedOffsets()` returns all offsets by topic-partition which have not yet been committed.
 * `isRunning()` returns true if consumer is in running state, else it returns false.
 * `isStale()` returns whether the messages in the batch have been rendered stale through some other operation and should be discarded. For example, when calling [`consumer.seek`](#seek) the messages in the batch should be discarded, as they are not at the offset we seeked to.
+* `pause(optionalTimeout)` can be used to pause the consumer for the current topic/partition for an optional period of time. All offsets resolved up to that point will be committed (subject to [autoCommit](#auto-commit) settings) and processing of this batch will be aborted. If no timeout is specified, the consumer will be paused indefinitely until `consumer.resume()` is called with the appropriate topic/partition. See [Pause & Resyume](#pause-resume) for more information about this feature.
```
Nevon, line 104:
```suggestion
* `pause(optionalTimeout)` can be used to pause the consumer for the current topic/partition for an optional period of time. All offsets resolved up to that point will be committed (subject to [autoCommit](#auto-commit) settings) and processing of this batch will be aborted. If no timeout is specified, the consumer will be paused indefinitely until `consumer.resume()` is called with the appropriate topic/partition. See [Pause & Resume](#pause-resume) for more information about this feature.
```

Typo

Nevon:
What's the intended use-case that makes increasing the API surface here worth it? The example [from the documentation](https://kafka.js.org/docs/consuming#a-name-pause-resume-a-pause-resume), just for comparison:

```ts
await consumer.run({ eachMessage: async ({ topic, message }) => {
    try {
        await sendToDependency(message)
    } catch (e) {
        if (e instanceof TooManyRequestsError) {
            consumer.pause([{ topic }])
            setTimeout(() => consumer.resume([{ topic }]), e.retryAfter * 1000)
        }

        throw e
    }
}})
```

And this proposal:

```ts
await consumer.run({ eachMessage: async ({ topic, message, pause }) => {
    try {
        await sendToDependency(message)
    } catch (e) {
        if (e instanceof TooManyRequestsError) {
            pause(e.retryAfter * 1000) // returns control to KafkaJS until timeout has expired
        }
        throw e
    }
}})
```

To me they look very similar.

The one benefit that I find interesting is that it allows a way to communicate back to the calling code that we actually want to exit out of the message loop without having to throw an error and everything this entails. Instead, we have a way for the consumer to know that we should continue operating as usual, just without processing any further fetched messages on that topic for as long as it's paused. The way it's communicated, by throwing a special error, I'm not so keen on, but the same thing can be achieved in other ways.

Then there's the thing about keeping the timer for resuming. Any time I see stateful code that includes timers, I get a bit nervous, because it's so hard to predict whether things are in a valid state to proceed with what you're doing when the timer fires. For example, does the consumer still exist? Is it still running? Is it still subscribed to the same topics? Is it still assigned the relevant partition? What if the user explicitly pauses the topic afterwards? And so on. Some of these might not be relevant in this case - I'm just pointing out that whenever you're dealing with doing something in the future based on nothing but a timer, things tend to get complicated and lead to complex bugs down the line.

Nevon:
If the provided `pause` callback is just a simple convenience over `consumer.pause` so that you don't have to provide the topic and partition, then I think it's a reasonable convenience with a  small downside. I like the design of `pause` returning a corresponding `resume` function.

However, we need to ensure that this functionality works the same whether you are pausing using `pause` or `consumer.pause`. Essentially, the implementation of `pause` can only be a simple delegation to `consumer.pause`. With the current design, the consumer will work differently depending on whether you've used the `pause` convenience function of `consumer.pause`, because only `pause` is affecting whether to break out of the consumer loop or not, which seems like a big footgun.

Funnily enough, someone else was doing similar work in #1382. What they found was that if you pause a topic-partition within the eachMessage or eachBatch functions, and then threw an error, the error would bubble up to the retrier and the retrier would retry processing the same batch from the now paused partition. There, the solution is to not invoke `onBatch` for paused topic-partitions (the list of paused topic-partitions can be gotten via the `ConsumerGroup`). I think very similarly, the same thing can be done here by simply bailing out of `eachMessage`/`eachBatch` if the current topic-partition is paused, but rather than keeping a local `paused` variable, we can expose `SubscriptionState.isPaused` via `ConsumerGroup` and check whether the current topic-partition is paused through there. That way it will work the exact same way whether you're pausing through `consumer.pause` or the convenience `pause` function. Make sense?

Nevon:
Just to set expectations, I will be going on vacation tomorrow and will be back on July 25th. So if you haven't heard from me before then, I'm not ghosting you, I'm just enjoying a coconut drink on a beach. 😄

src/consumer/runner.js
```diff
@@ -416,6 +434,16 @@ module.exports = class Runner extends EventEmitter {
           })
           return
         }
+        if (this.consumerGroup.isPaused(batch.topic, batch.partition)) {
```
Nevon, line 437: I wonder if this needs to come after the `if (isRebalancing(e) ...` condition, since I would assume that even if this topic-partition is paused, we would still want to rejoin the group and potentially receive new assignments.

docs/Consuming.md
```diff
@@ -100,6 +102,7 @@ await consumer.run({
 * `uncommittedOffsets()` returns all offsets by topic-partition which have not yet been committed.
 * `isRunning()` returns true if consumer is in running state, else it returns false.
 * `isStale()` returns whether the messages in the batch have been rendered stale through some other operation and should be discarded. For example, when calling [`consumer.seek`](#seek) the messages in the batch should be discarded, as they are not at the offset we seeked to.
+* `pause()` can be used to pause the consumer for the current topic/partition. All offsets resolved up to that point will be committed (subject to `eachBatchAutoResolve` and [autoCommit](#auto-commit) settings). If you want to pause in the middle of the batch, you should throw an exception after pausing if you do not wish to resolve the current batch's offset or, alternatively, make sure that `eachBatchAutoResolve` is turned off. You can used the returned callback function to resume processing of the relevant partition and topic (i.e. using `setTimeout` or some other asynchronous trigger). See [Pause & Resume](#pause-resume) for more information about this feature.
```
Nevon, line 105:
```suggestion
* `pause()` can be used to pause the consumer for the current topic/partition. All offsets resolved up to that point will be committed (subject to `eachBatchAutoResolve` and [autoCommit](#auto-commit) settings). If you want to pause in the middle of the batch without resolving the current offset, you should throw an error. Alternatively, make sure that `eachBatchAutoResolve` is turned off. You can used the returned callback function to resume processing of the relevant partition and topic (i.e. using `setTimeout` or some other asynchronous trigger). See [Pause & Resume](#pause-resume) for more information about this feature.
```

docs/Consuming.md
```diff
@@ -304,6 +307,31 @@ consumer.run({
 })
 ```
 
+As a convenience, the `eachMessage` callback provides a simple callback function to pause the specific topic/partition for the message currently being processed. You can also specify a timeout for when that topic/partition will automatically be resumed:
```
Nevon, line 310:
```suggestion
As a convenience, the `eachMessage` callback provides a `pause` function to pause the specific topic-partition of the message currently being processed.
```

Nevon (changes requested):
Alright, I am back from vacation and have gone through this another round. Looks good to me overall, just have some notes on the comms as well as one possibly important note on handling errors that require rejoining.

src/consumer/runner.js
brianphillips: Yes, that sounds right to me.

Nevon:
Merged master and resolved the conflicts. With #1382 the handling of paused partitions during error handling is already taken care of.

PR #1372, opened 2022-05-25
Ability to use custom authenticators

markgaylard:
Starting from https://github.com/tulios/kafkajs/pull/1101, this is a rework to allow custom authentication mechanisms to be written based on the discussions in that PR around the desired interface for custom authentication mechanisms.

To use a custom authenticator, the `sasl` object becomes
```javascript
{ 
      mechanism: string; 
      authenticationProvider: ( host: string, port: number, logger, saslAuthenticate) => { authenticate: () => Promise<void> }
}
```

Fixes https://github.com/tulios/kafkajs/issues/840


types/index.d.ts
```diff
@@ -16,10 +16,34 @@ export class Kafka {
...
+}
+
+export type Authenticator = {
+  authenticate: () => Promise<void>
+}
+
+export type Mechanism = {
+  mechanism: string
+  authenticationProvider: (
+    connection: { host: string; port: number },
```
markgaylard, line 34: I'm wondering if this field should just be flattened and host and port passed in directly.

Nevon:
> I haven't updated the existing authenticators at this stage as was done in the previous PR, but I can if that is desired.

When I was working on this, I found that to be the best way to test out the interface and how awkward or not awkward it was to use it, so I would highly recommend doing that.

types/index.d.ts
Nevon: Probably makes sense. I think the built-in authenticators receive the entire `connection` object because they also want access to the `SASLOptions` for credentials and such, but that's not necessary for the custom authenticators.

Nevon: Thinking more about it, do you even need the host and the port? The authentication is done through `saslAuthenticate`, not by some bespoke connection to the broker anyway. I think the host and port are literally only used for logging in the existing authenticators. That does not seem necessary. The logger will log which broker it's communicating with when the SASL requests are made anyway.

markgaylard: The AWS MSK IAM auth uses the host as part of generating the request object (as part of the generated signature) so because I don't know what other mechanisms will need I'll leave in the host/port fields but I'll pull them to the top level.

mehul-sg:
Hi, @Nevon When can we expect this to be released? Thanks.

Nevon:
> When can we expect this to be released?

Once it's ready. I will be going on vacations tomorrow, however, and won't be back until July 25th, so very unlikely to be before then.

markgaylard:
@Nevon I think I'm done with this now. I've ported all the changes from your PR to this one. I'm glad you had done most of the hard work otherwise this would have taken me a lot longer :)

Nevon:
@christianpurple: If you are in a rush to use code that has not been merged, you can always use a branch reference by changing your dependency to `"kafkajs": "markgaylard/kafkajs#pluggable-auth"`. Like I said before, it will be merged when it is ready, and when I've had the time to properly review it and both me and @markgaylard are happy with it.

Nevon (changes requested):
Looks good, and tests are passing, so it seems to be working well. The only thing I think is important to do before merging is to turn the arguments of the authenticationProvider into an object instead of positional arguments, to make it easier to evolve the interface if needed.

src/broker/saslAuthenticator/index.js
```diff
@@ -33,22 +32,16 @@ module.exports = class SASLAuthenticator {
...
-    }
-
     const handshake = await this.connection.send(this.saslHandshake({ mechanism }))
     if (!handshake.enabledMechanisms.includes(mechanism)) {
       throw new KafkaJSSASLAuthenticationError(
         `SASL ${mechanism} mechanism is not supported by the server`
       )
     }
 
-    const saslAuthenticate = async ({ request, response, authExpectResponse }) => {
```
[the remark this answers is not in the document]
tulios: @Nevon It's used as part of the SCRAM handshake - https://github.com/tulios/kafkajs/blob/4246f921f4d66a95f32f159c890d0a3a83803713/src/broker/saslAuthenticator/scram.js#L171-L174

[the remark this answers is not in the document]
Nevon: Yeah, no I get what it was being used for, just not if there was a good reason to not just use the presence of `response` to determine whether or not to wait for a response (like if there would be a response that we didn't want to decode, for some weird reason), but that doesn't seem to be the case.

docs/Configuration.md
```diff
@@ -265,6 +265,23 @@ A complete breakdown can be found in the IAM User Guide's
...
+### Custom Authentication Mechanisms
+
+If an authentication mechanism is not supported out of the box in KafkaJS, a custom authentication
+mechanism can be introduced as a plugin:
+
+```js
+{ 
+  sasl: { 
+      mechanism: <mechanism name>,
+      authenticationProvider: (host, port, logger, saslAuthenticate) => { authenticate: () => Promise<void> }
```
Nevon, line 277:
```suggestion
      authenticationProvider: ({ host, port, logger, saslAuthenticate }) => { authenticate: () => Promise<void> }
```

docs/CustomAuthenticationMechanism.md
```diff
@@ -0,0 +1,166 @@
...
+---
+
+To use an authentication mechanism that is not supported out of the box by KafkaJS,
+custom authentication mechanisms can be introduced:
+
+```js
+{ 
+  sasl: { 
+      mechanism: <mechanism name>,
+      authenticationProvider: (host, port, logger, saslAuthenticate) => { authenticate: () => Promise<void> }
```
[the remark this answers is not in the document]
markgaylard: I made myself a note to come back and fix the docs when I changed the parameters and still missed it :(

Nevon:
Thanks for the contribution. Let's get some people to try out this new interface before we make a stable release out of it, to see if we missed anything in how this works.

MichaelAllenHardeman:
@Nevon This is great. I'm trying to use this new feature to implement a GSSAPI mechanism. I've looked at https://github.com/adaltas/node-krb5 and https://github.com/mongodb-js/kerberos for native kerberos bindings but at initial glance neither of them seem suited for your custom auth mechanisms? in node-krb5 I can get a ticket granting ticket and spnego token and it says I should add it as an `Authorization: Negociate <token>` header on all http requests to services similar to a bearer token but I don't really see how that would be done.

m-ketan:
@michael-hardeman  Did you have any luck implementing the GSSAPI mechanism? I myself have spent a great deal of time in the implementation but couldn't get it to work. I went as far as issuing a ticket, encoding it but always get an "Invalid credentials" error.

@Nevon Can you please point us in the right direction on how to implement GSSAPI mechanism in KafkaJS?

michael-hardeman:
> @michael-hardeman Did you have any luck implementing the GSSAPI mechanism? I myself have spent a great deal of time in the implementation but couldn't get it to work. I went as far as issuing a ticket, encoding it but always get an "Invalid credentials" error.

No. I had issues as well and there was a severe lack of documentation on kerberos auth and, at the time, the authentication portion here. I ended up taking a different route in design instead.

PR #1382, opened 2022-06-01
Fix consumer pause resume functionality

### LINKED ISSUE #1376: Pausing the consumer doesn't work in 2.x versions as before
**Describe the bug**
Calling `consumer.pause` inside `eachMessage` or `eachBatch` and breaking the consumption flow, does not pause the consumption from a given topic-partition. The same message batch is received again, however, the consumption from that topic-partition is paused.

**To Reproduce**
1. Create a new consumer instance.
2. Connect and subscribe to a given topic.
3. Create a handler function that pauses consumption from the received topic-partition, sets a timer for the consumption, and throws an error.
4. Register a runner with the created handler function.
5. The paused topic-partitions are ignored and messages are still received

```ts
  await consumer.connect();
  await consumer.subscribe({ topic });

  await consumer.run({ eachMessage: async ({ topic, partition, message }) => {
      try {
        doWork(message);
      } catch (e) {
        consumer.pause([{ topic, partitions: [partition] }])
        setTimeout(() => consumer.resume([{ topic, partitions: [partition] }]), 1000)

        throw e
      }
  }})
```

**Expected behavior**
When pausing consumption from a particular topic-partition and breaking the consumer (throwing an error inside `eachMessage`, `eachBatch`), the consumer should not receive messages from that topic-partition, until it's resumed.

**Observed behavior**
When pausing consumption from a particular topic-partition and breaking the consumer (throwing an error inside `eachMessage`, `eachBatch`), the consumer receives the same message batch even if the consumption is not resumed for that topic-partition.

**Environment:**
 - OS: [e.g. Windows 10]
 - KafkaJS version [e.g. 2.0.0] (It works with version 1.4.0)
 - NodeJS version [e.g. 14.16.0]

**Additional context:**
I did some research. As I can see in the fetch phase the topic-partitions are filtered from where the messages should be fetched, and then the processing phase begins.
Before 2.0.0 both phases were wrapped inside a retry strategy. After 2.0.0 only the message processing phase is wrapped inside a retry strategy, and because of that on a retry, the active partitions are not checked.

arszen123:
This fixes #1376 

The problem was that when in the `eachMessage` or `eachBatch` the consumption from a topic-partition is paused and an error is thrown (breaks further consumption of messages). The same batch (that was already fetched) is processed again, instead of fetching a new batch of messages.

### DIFF
--- a/src/consumer/runner.js
+++ b/src/consumer/runner.js
@@ -90,10 +90,25 @@ module.exports = class Runner extends EventEmitter {
     this.scheduleFetchManager()
   }
 
-  async scheduleFetchManager() {
+  scheduleFetchManager() {
+    if (!this.running) {
+      this.consuming = false
+
+      this.logger.info('consumer not running, exiting', {
+        groupId: this.consumerGroup.groupId,
+        memberId: this.consumerGroup.memberId,
+      })
+
+      return
+    }
+
     this.consuming = true
 
-    while (this.running) {
+    this.retrier(async (bail, retryCount, retryTime) => {
+      if (!this.running) {
+        return
+      }
+
       try {
         await this.fetchManager.start()
       } catch (e) {
@@ -110,7 +125,7 @@ module.exports = class Runner extends EventEmitter {
           })
 
           await this.consumerGroup.joinAndSync()
-          continue
+          return
         }
 
         if (e.type === 'UNKNOWN_MEMBER_ID') {
@@ -122,16 +137,33 @@ module.exports = class Runner extends EventEmitter {
 
           this.consumerGroup.memberId = null
           await this.consumerGroup.joinAndSync()
-          continue
+          return
         }
 
-        this.onCrash(e)
-        break
-      }
-    }
+        if (e.name === 'KafkaJSNotImplemented') {
+          return bail(e)
+        }
 
-    this.consuming = false
-    this.running = false
+        this.logger.debug('Error while scheduling fetch manager, trying again...', {
+          groupId: this.consumerGroup.groupId,
+          memberId: this.consumerGroup.memberId,
+          error: e.message,
+          stack: e.stack,
+          retryCount,
+          retryTime,
+        })
+
+        throw e
+      }
+    })
+      .then(() => {
+        this.scheduleFetchManager()
+      })
+      .catch(e => {
+        this.onCrash(e)
+        this.consuming = false
+        this.running = false
+      })
   }
 
   async stop() {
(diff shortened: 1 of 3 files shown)


src/consumer/runner.js
```diff
@@ -90,10 +90,25 @@ module.exports = class Runner extends EventEmitter {
...
+        memberId: this.consumerGroup.memberId,
+      })
+
+      return
+    }
+
     this.consuming = true
 
-    while (this.running) {
+    this.retrier(async (bail, retryCount, retryTime) => {
```
Nevon, line 107: This will now crash once the retrier runs out of attempts in case of repeated rebalances. Before, it would just continue to attempt to fetch. I can't really tell if that's a feature or a bug, but it is what I assume is an unintended change.

src/consumer/runner.js
```diff
@@ -404,39 +440,7 @@ module.exports = class Runner extends EventEmitter {
       await this.heartbeat()
     }
 
-    return this.retrier(async (bail, retryCount, retryTime) => {
-      try {
-        await onBatch(batch)
```
Nevon, line 409: This now means that on every error, we are going to re-fetch the data for no real reason. The data that we received is still valid, it's just that we don't want to process some of it. I would instead suggest that we conditionally invoke `onBatch` only if the topic-partition is not paused. That way if the topic-partition was paused in the execution of the handler, we just continue processing the other data included in the fetch response without having to do any additional data fetching, and we don't have to do any particularly intrusive changes to the runner. It already has access to the `ConsumerGroup` which holds the state of which topic-partitions are paused.

Nevon (changes requested):

src/consumer/runner.js
arszen123: Correct me if I am wrong, but based on the fetch manager implementation, the rest of the batches (from other partitions) will be processed in that cycle. Therefore there will be no unnecessary data fetching.
If the processing of a batch fails, it is most likely that part of that batch will be re-fetched in the next cycle, which I think is okay, because we don't know what the last committed offset was to filter out messages. It is possible to fetch the offsets but I think it's more convenient to drop and re-fetch.
Thus the messages from the paused partition will be fetched in the next fetch cycle, right after the consumption from that partition is resumed.

Nevon: What you're saying is correct. What I'm considering is the regular case where the topic-partition is not paused. If there is a retriable error that happens during the processing of a particular batch, previously the retrier here would end up retrying the operation before eventually giving up and yielding back to the runner. Now, the rejection will always bubble up to the runner, causing re-fetching of the same batch again. You are correct that it will only re-fetch the particular partition that caused the error, however (of course it'll fetch others too, potentially, but with the next offsets).

We could filter out already processed messages from the batch by looking at the locally resolved offsets, but I do see how that might get tricky. Looking at the code in master, it looks like currently the whole batch would end up being re-processed, even if some of the offsets had already been resolved, which seems worse than re-fetching from the last resolved offset.

Nevon (approved):

PR #1406, opened 2022-06-30
Update ConsumerExample.md

stilyng94:
**ConsumerSubscribeTopic** is deprecated and has been replaced with **ConsumerSubscribeTopics**.
**eachBatchPayload** was spelled wrong as **eatchBatchPayload**.
**eachBatchPayload** does not directly have `{topic, partition}` as properties. These properties can be directly be found on the `{batch}` property rather.


Nevon (approved):
Excellent! Thank you for your contribution.

Issue #1384, opened 2022-06-02
Consumer runner tries to start the fetch manager without delay


adripc64:
**Describe the bug**
When the FetchManager detects that a kafka node is lost, it tries to rebalance the fetchers. If there are no kafka nodes available, no fetchers are created. That causes the consumer runner to call FetchManager.start() in an infinite loop without any delay.
**This blocks any other processing like responding to http requests**. Also, the consumer is not disconnected properly nor can recover when kafka is available again.

**To Reproduce**
1. Having a kafka with a single node
2. Run a consumer that subscribes to a topic and consumes messages
3. Stop kafka instance
4. The consumer runner tries to rebalance fetchers. As there are no nodes, no fetchers are created. The consumer runner is calling FetcherManager.stat() continuosly without any delay, so blocking any other kind of processing in node.js

**Expected behavior**
The consumer should be disconnected properly and start retrying to connect it again.
As a hint, we have managed to solve this issue adding a simple delay of 1ms after the call to FetcherManager.stat(). Example:

```ts
async scheduleFetchManager() {
    this.consuming = true

    while (this.running) {
      try {
        await this.fetchManager.start()
        await new Promise(resolve => setTimeout(resolve, 1));
      } catch (e) {
      ...
      }
    }
    ...
}
```

**Observed behavior**
```
{"level":"DEBUG","timestamp":"2022-06-02T09:57:05.213Z","logger":"kafkajs","message":"[FetchManager] Starting..."}
{"level":"DEBUG","timestamp":"2022-06-02T09:57:05.213Z","logger":"kafkajs","message":"[FetchManager] Created 0 fetchers","nodeIds":[],"concurrency":1}
{"level":"DEBUG","timestamp":"2022-06-02T09:57:05.213Z","logger":"kafkajs","message":"[FetchManager] Starting..."}
{"level":"DEBUG","timestamp":"2022-06-02T09:57:05.213Z","logger":"kafkajs","message":"[FetchManager] Created 0 fetchers","nodeIds":[],"concurrency":1}
{"level":"DEBUG","timestamp":"2022-06-02T09:57:05.213Z","logger":"kafkajs","message":"[FetchManager] Starting..."}
{"level":"DEBUG","timestamp":"2022-06-02T09:57:05.213Z","logger":"kafkajs","message":"[FetchManager] Created 0 fetchers","nodeIds":[],"concurrency":1}
[...]

PR #1408, opened 2022-07-05
Fix consumer connection issue

arszen123:
In the #1382 PR, I made a mistake that has performance implications.

When a `KafkaJSConnectionError` (ex. Connection timeout) is thrown during the fetching + processing, the consumer restarts. I think it is not supposed to restart but retry the fetch.
Based on the tests it's only supposed to restart when there are no available brokers (ex. the cluster is disconnected) https://github.com/tulios/kafkajs/blob/master/src/consumer/__tests__/connection.spec.js#L62-L103

Based on the above-mentioned I made a modification so the consumer will retry on the `KafkaJSConnectionError` error, and restarts when there are no available brokers.


src/consumer/fetchManager.spec.js
```diff
@@ -94,4 +94,10 @@ describe('FetchManager', () => {
       await expect(fetchManagerPromise).rejects.toThrow('Node not found')
     })
   })
+
+  it('should throw an error when there are no avaiblae brokers', async () => {
```
michie1, line 98: `avaiblae` -> `available`?

ekanth:
@arszen123 - Thanks for this fix. And I think this fix is also needed to handle the scenario where `KafkaJSConnectionClosedError` causes an infinite loop when no brokers are reachable (reproducible easily with a single broker docker setup). Same as this [issue](https://github.com/tulios/kafkajs/issues/1384#issuecomment-1160598427) but caused by `KafkaJSConnectionClosedError`. @Nevon @tulios - appreciate if you could review/merge this fix when you get a chance.

src/consumer/fetchManager.spec.js
Saeger, line 98:
```suggestion

  it('should throw an error when there are no brokers available', async () => {
```

kosarajupavankumar:
ERROR [Connection] Connection error: getaddrinfo EAI_AGAIN edEVENTHUBS01.servicebus.windows.net {"timestamp":"2023-08-14T16:31:54.290Z","logger":"kafkajs","broker":"EVENTHUBS01.servicebus.windows.net:9093","clientId":"microservices/loan/consumer","stack":"Error: getaddrinfo EAI_AGAIN edEVENTHUBS01.servicebus.windows.net\n at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:107:26)\n at GetAddrInfoReqWrap.callbackTrampoline (node:internal/async_hooks:130:17)"}

ERROR [Producer] Failed to send messages: Connection error: Client network socket disconnected before secure TLS connection was established {"timestamp":"2023-08-14T16:31:10.114Z","logger":"kafkajs","retryCount":0,"retryTime":249}

@Nevon and @arszen123 after using the latest version of the 2.2.4 still we are getting the connection errors from the kafka side

Issue #1181, opened 2021-09-02
KIP-455: Implement alterPartitionReassignments/listPartitionReassignments admin methods


adalbert-homa:
These methods allow for Kafka partition managements.

I can take a look at implementing this.

Nevon:
If anyone wants to pick this up, I think it should be a nice and simple first issue:

* [KIP-455](https://cwiki.apache.org/confluence/display/KAFKA/KIP-455%3A+Create+an+Administrative+API+for+Replica+Reassignment)
* [ListPartitionReassignments protocol definition](https://kafka.apache.org/protocol.html#The_Messages_ListPartitionReassignments)
* [AlterPartitionReassignments protocol definition](https://kafka.apache.org/protocol.html#The_Messages_AlterPartitionReassignments)

Nevon:
This is now available in [kafkajs@2.2.0-beta.2 ](https://www.npmjs.com/package/kafkajs/v/2.2.0-beta.2) thanks to @jonathanhaviv

PR #1419, opened 2022-07-20
KIP-455: Implement alterPartitionReassignments/listPartitionReassignments admin methods

jonathanhaviv:
This PR adds the protocols and methods for alterPartitionReassignments and listPartitionReassignments referenced in #1181. There are also a few updates to the encoder and decoder for working with the types added by [KIP-482](https://cwiki.apache.org/confluence/display/KAFKA/KIP-482%3A+The+Kafka+Protocol+should+Support+Optional+Tagged+Fields):
- writeUVarIntArray was updated to be able to encode a null array as per the documentation.
- readUVarIntString was updated to slice from the buffer the length of the bytes - 1 (the length of the string) and move the offset based on the length of the string instead of the byte length since the byte length was the length of the string + 1.

One note on tagged fields. Requests and responses with tagged fields have them in two places:
- At the top level of the request/response 
- Wherever there is a TAG_BUFFER

The implication of this is that if there are no tagged fields a zero byte needs to be written/read in multiple places depending on the rest of the request/response. 

Hope that helps with how to handle any protocols that use the tagged fields or compact types.


src/protocol/requests/alterPartitionReassignments/v0/request.js
```diff
@@ -0,0 +1,43 @@
...
+ *    replicas => INT32
+ */
+
+module.exports = ({ topics, timeout = 5000 }) => ({
+  apiKey,
+  apiVersion: 0,
+  apiName: 'AlterPartitionReassignments',
+  encode: async () => {
+    return new Encoder()
+      .writeUVarIntBytes()
```
Nevon, line 20: According to the KIP for tagged fields, the tagged fields in the request should be at the end of the request header (before the request payload itself). The protocol also dictates that for requests supporting tagged fields (called "flexible versions"), the clientId should be a compact string rather than a string. https://cwiki.apache.org/confluence/display/KAFKA/KIP-482%3A+The+Kafka+Protocol+should+Support+Optional+Tagged+Fields

We don't have an explicit request header construct at the moment, but the equivalent is here: https://github.com/tulios/kafkajs/blob/7e9970ba580b4b0d48d10504ea6b248351a61a99/src/protocol/request.js#L5-L10

So we need to introduce the request header concept, in order to support both V0 and V1 (flexible versions and non-flexible versions) and then have a way to determine which header version a particular protocol request version should use, even if at the moment we never actually send any tagged fields.

On the response side of things, it's saying that at the end of the response header, there will be a tagged fields buffer, but then also at the end of each response. 
We can see this in the [headers](https://kafka.apache.org/protocol.html#protocol_messages) section of the protocol definition. The tagged fields at the end of the response should probably be read by each protocol response itself, but the ones that are part of the response headers should probably not be read by each individual protocol implementation, since they apply to all flexible requests, similar to how we construct the request headers in [protocol/request.js](https://github.com/tulios/kafkajs/blob/7e9970ba580b4b0d48d10504ea6b248351a61a99/src/protocol/request.js#L5-L10).

I don't think doing a simple "skip 1 varint" implementation will work, since I assume the broker can choose to return tagged fields at any time if the protocol version says it supports them, so at the very least we need to read the length and then skip over that number of bytes. For the tagged fields at the end of the request, it kinda works to just ignore them, as we don't need to read past them anyway, but for the ones in the request header we need to know how many bytes to skip so that we know where the message payload begins. The structure of the tagged fields buffer seems to be `number_of_fields (uvarint)` + `tag (uvarint) size (uvarint) value (size bytes)` * number_of_fields. So we'd need to read the number of fields, then for each field read a uvarint followed by another uvarint, then the number of bytes indicated by the size field uvarint.

src/protocol/requests/alterPartitionReassignments/v0/request.spec.js
```diff
@@ -0,0 +1,25 @@
+const RequestV0Protocol = require('./request')
+
+/* TO-DO: Update the fixtures, data, and errors */
```
Nevon, line 3: Still a todo?

Nevon (changes requested):
I'm afraid I haven't had time to go through this really thoroughly yet, as I'm busy with my real job, but there are a couple of things that will need to be looked at. Most importantly the handling of the tagged fields in the response header, since the current implementation won't work if we receive any tagged field, even though the API version indicates that we support them.

src/protocol/requests/alterPartitionReassignments/v0/request.js
jonathanhaviv: Gotcha, the tagged fields being the request header and in the locations of the tagged buffer makes sense since I was finding the tagged fields (in this case empty since we didn't have any) right before the first field in the request/response. I'm not sure if this is a case where the documentation on tagged fields is unclear but from the responses I was looking at I saw tagged fields appearing in both locations. This might have also been since I was using the current implementation.

Instead of the skip 1 var int solution right now, I'm thinking that for now reading the tagged fields in the format you mentioned should work inside the response and requests. And can implement this into the current request and response. When I looked at the Java code for this request I saw them reading tagged fields in multiple locations so I implemented it within the request/response as well. The structure for tag buffers you mentioned looks similar to the compact array but will take another look at the code to confirm this.

For the request headers, I think the client id is a nullable string so the current implementation would work. We would just need to add a field before the encoding of the payload to account for them. Similarly reading them before passing the payload on.

I can try to take a look at how to determine if a request/response is using flexible versions and then assign them the right header. This seems like it might be bigger than this issue so let me know if it would be better to focus on just updating these requests/responses to read the tag buffers correctly or try to handle both pieces at once?

jonathanhaviv: Been thinking about this and have a few proposed solutions I wanted to get thoughts on before implementing.

1. Having the protocol request/response handle this. We can add an argument to each protocol request version that can use tagged fields and it will correctly encode the tagged fields based on that. For right now it can default to null and be added as an argument to the broker. As a contributor adds a version of a protocol that is the first to accept tagged fields they would update the request and broker method to accept them. For responses reading the tagged fields at the response level would also make sense. Besides the tagged fields the previous request/response headers are the same as the v2/v1 request and response headers.

2. Building out an object/set that contains each protocol and the request/response version that introduces tagged fields. Based on this logic it will look at the API version in the request/response and then for a version that is greater than or equal to the version in our lookup object it would handle it. For a request, it would use a v2 header which would look for a tagged_field argument and encode it in the headers. For a response, it would decode this at the process data level.

For both options, it would mean adding a tagged field argument to the requests and broker methods. If we choose to add the ability to send tagged fields it would only need an argument added to the admin method for these requests. We can also return this on the broker method and not add it as returned on the admin method for now.

I prefer option 1 because it requires less overall maintenance of the mapping and which versions can support the tagged fields. Additionally tagged fields appear in both the headers and body which means that protocols will need to decode them at both levels either way. Contributors who come and create the first protocol version for a request that supports tagged fields they can add that to the new protocol that is being written. Also, the versions that accept tagged fields will need updates regardless to handle compact strings and compact arrays.

Very possible I'm missing something but let me know what you think and I can start working on the implementation.

Nevon: Having slept on it, and then briefly reading through the tagged field implementations of the Java client and franz-go to see how they handle it, I think I've come to the conclusion that properly supporting tagged fields is _way_ outside the scope of this PR.

Those other clients went with your second option. But they are able to do so because they made the very wise decision to generate their protocol encoding/decoding code from an API schema. So in the schema for each API they have a field that says that from version X this api is "flexible", and then in the code generator they check if the version they're generating code for is flexible, and if so, adds the code to decode/encode fields where needed. We have only really talked about the "static" tagged fields, which are in the request/response header and after each message, but in actuality tagged fields are present at the end of any struct, so each individual protocol request/response has to handle this anyway.

In a proper implementation I suspect we'll need to add some layer in between the protocol request/response and the Decoder/Encoder, so that the request/response doesn't work on such a low level as it does today, and instead instructs this middle layer "I want to read a struct of this shape", so that the middle layer can decide whether or not it needs to also read tagged fields.

But, as I said, I think this is way outside of the scope of this PR. To land this PR, we just need to make sure that if there are any tagged fields in the response, we skip over them. The API version you are implementing doesn't depend on any tagged field, so we can safely ignore them for now.

So my suggestion is that we simply have a "skipTaggedFields" function that's colocated with your request/response functions that just knows how to decode a tagged field to the point where it's able to skip past the correct number of bytes. That way it's explicit that this is not a full implementation and the day that we implement proper support it'll be obvious that this needs to be removed. When we start implementing support for other flexible API versions we'll first need to add real support for tagged fields.

docs/Admin.md
```diff
@@ -680,3 +680,64 @@ Be aware that the security features might be disabled in your cluster. In that c
...
+
+## <a name="alter-partition-reassignments"></a> Alter Partition Reassignments
+This is used to reassign the replicas that partitions are on. This method will throw exceptions in the case of errors.
+
+```javascript
+await admin.alterPartitionReassignments({
+  topics: <IPartitionReassignment[]>
+  timeout: <Number> // optional - 5000 default
+})
+```
```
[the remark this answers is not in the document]
jonathanhaviv: Think this should work now. Still tried to maintain the types definition that is other pieces of the [documentation](https://kafka.js.org/docs/admin#create-partitions). If an example that explained it further would be better though let me know.

Nevon (approved):
Looks good. I just have a question about what looks like a bug in reading uvarintbytes, and then we're good to go.

src/protocol/decoder.js
```diff
@@ -156,7 +157,7 @@ module.exports = class Decoder {
     }
 
     const stringBuffer = this.buffer.slice(this.offset, this.offset + byteLength)
```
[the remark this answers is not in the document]
Nevon: Having done a bit of reading, I think we have a mixup of different types.

The protocol makes a distinction between nullable and not-nullable fields. You can have COMPACT_STRING or COMPACT_NULLABLE_STRING, where the difference is that in the nullable version a length of 0 means null whereas in the not-nullable version a length of 0 means an empty string (same for the other COMPACT types).

At the moment, `readUVarIntBytes` is being used for both cases (see the `errorMessage` and `responses.name` in the AlterPartitionReassignmentsResponse), which won't work properly. We'll need to ensure that we have `readUVarIntString` and `readUVarIntNullableString`, etc, so that we can differentiate between the two cases.

Nevon (changes requested):
Just noticed the mixup of nullable and non-nullable fields.

src/protocol/decoder.js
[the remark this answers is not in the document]
Nevon: Nevermind. I wrote some tests to see if I could help you out with this, and I now see that in practice this doesn't matter. On the wire, there's no difference between a compact string and a compact nullable string. The only difference is whether we should allow a 0 value or if we should throw an error when reading/writing such a value - but if we treat all these types as nullable, then the only issue is if we receive an invalid message and interpret that as null instead of throwing an error - not a big deal.

Nevon (approved):

PR #1429, opened 2022-07-27
[draft] Support manual assignment of partitions to consumer

alee101:


src/manualConsumer/manualConsumer.js
```diff
@@ -161,221 +95,40 @@ module.exports = class ConsumerGroup {
...
-
-    if (this.isLeader()) {
-      this.logger.debug('Chosen as group leader', { groupId, generationId, memberId, topics })
-      const assigner = this.assigners.find(({ name }) => name === groupProtocol)
-
-      if (!assigner) {
-        throw new KafkaJSNonRetriableError(
-          `Unsupported partition assigner "${groupProtocol}", the assigner wasn't found in the assigners list`
-        )
+  async assignPartitions() {
```
alee101, line 98: This is the main difference from the regular consumer group - instead of issuing a JoinGroup and then SyncGroup to set up a consumer group and assign partitions, we instead construct a mapping of all relevant topic partitions in memory, resolve their oldest/newest offsets, then immediately start issuing Fetch requests to the brokers. The brokers maintain no state for manual consumers.  A manual consumer instead tracks and updates an in-memory mapping (in offsetManager) of the last offset it's processed for each partition, and uses those offsets to construct its Fetch requests. As a result, none of the usual consumer group functionality around committing offsets back to the broker or sending heartbeats are needed for the manual consumer.

Nevon: Based on your description, would it be reasonable to think we could base the regular consumer on top of the manual consumer, either by wrapping it or extending it (or a shared base class)? Conceptually it sounds like there's actually not much difference between the two. It sounds to me like the regular consumer has some extra initialization to do, and needs to be able to hook in to do heartbeating and group offset management in a few places, but most of the mechanics around the fetch loop should be the same. Both of them receive assignments from "outside", it's just that the manual consumer gets it from the user and the regular consumer gets it from the consumer group leader.

But I haven't dissected in detail what the differences are, so I'm sure there will be some tricky spots to handle. What do you think is the way forward?

JaapRood: Hey team, just chiming in from the original issue around adding this. At the time I discussed a possible implementation, informed by the "original" Java client, with Tulio. Not sure if you've seen it, but there's some detail there and might be of help: https://github.com/tulios/kafkajs/issues/373#issuecomment-555927883

@Nevon as you mention, there really isn't much difference between the consumers, other than the source of their toppar assignments. Instead of calling `subscribe` and allowing the group coordinator to provide a toppar assignment, the user assigns what toppars to consume. Signals like receiving assignments and group rebalances can just go ignored. Note: committing offsets should still be possible; even if it's not particularly useful in most cases, there are some possible use cases empowered by it. In the end, committing an offset is just producing a message to a specific topic. 

Let me know if I there's anything in particular you'd like to chat about.

alee101: @JaapRood thanks for the context! Reading over that thread, there seem to be two slightly different use cases around manual topic partition assignment:
1. Consumer manually assigns topic partitions and specifies a groupId, which means that the consumer can still commit offsets if it wants to. This matches the example in the comment you linked.
2. Consumer manually assigns topic partitions and does not specify a groupId, so it cannot commit offsets.

I was originally only thinking of handling the second case with this change. I haven't done enough investigation yet to understand the scope of also including the first case. I'm not sure if it's almost exactly the same as the second case, just with the ability to commit offsets, or whether it's closer to the regular consumer group case, where apart from the initial topic partition assignments, everything else (making the broker aware of the existence of the group, sending heartbeats, committing offsets, etc.) all work the same.

If we're ok with just moving forward with the second case, I can work on extracting out the common functionality so that the regular consumer wraps or extends the manual consumer or a common base class. I agree that there's definitely a lot more in common between the two than different.

Nevon: As long as we don't paint ourselves into a corner where when the time comes we want to handle case 1, I'm in favor of reducing the scope to just manual assignment and no offset committing.

Great to hear from you again @JaapRood btw. Been a minute! 😄

src/consumer/consumer.js
```diff
@@ -371,6 +372,41 @@ module.exports = class ConsumerGroup {
...
+        partitions: partitionMetadata.map(metadata => metadata.partitionId),
+      }
+    })
+    this.subscriptionState.assign(memberAssignment)
+    this.partitionsPerSubscribedTopic = this.generatePartitionsPerSubscribedTopic()
+
+    const offsetManagerMemberAssignment = memberAssignment.reduce(
+      (partitionsByTopic, { topic, partitions }) => ({
+        ...partitionsByTopic,
+        [topic]: partitions,
```
snirad, line 389: isnt this anti parttern /.
https://www.richsnapp.com/article/2019/06-09-reduce-spread-anti-pattern

src/consumer/offsetManager/manualOffsetManager.js
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    const pendingPartitions = this.topics
+      .map(topic => ({
+        topic,
+        partitions: this.memberAssignment[topic]
+          .filter(invalidOffset(topic))
+          .map(partition => ({ partition })),
+      }))
+      .filter(t => t.partitions.length > 0)
```
snirad, line 105: reduce instead?

src/consumer/offsetManager/manualOffsetManager.js
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    const hasUnresolvedPartitions = () => unresolvedPartitions.some(t => t.partitions.length > 0)
+
+    if (hasUnresolvedPartitions()) {
+      const topicOffsets = await this.cluster.fetchTopicsOffset(unresolvedPartitions)
+      const offsets = initializeConsumerOffsets(topicOffsets, topicOffsets)
+      offsets.forEach(({ topic, partitions }) => {
+        this.resolvedOffsets[topic] = partitions.reduce(indexPartitions, {
+          ...this.resolvedOffsets[topic],
```
snirad, line 132: can you simplify this ? a bit  hard to understand what going on here

alee101: This merges the unresolved partitions into `this.resolvedOffsets`. I kept it like this for consistency with how we do the merging of partitions into `this.committedOffsets` in the existing group offset manager.

PR #1450, opened 2022-09-27
doc: fix invalid function toWinstonLogLevel

javigaralva:
Fix `toWinstonLogLevel` function syntax error.


Nevon:
Thanks for the fix!

PR #1459, opened 2022-10-13
Fix saslAuthenticate parameter type definitions

### LINKED ISSUE #1458: Incorrect Typescript interface for SASLAuthenticate handler (custom mechanisms)
**Describe the bug**
The Typescript interface is defined as:

```ts
  saslAuthenticate: <ParseResult>(
    request: SaslAuthenticationRequest,
    response?: SaslAuthenticationResponse<ParseResult>
  ) => Promise<ParseResult | void>
```

While the implementation expects a single arg that is a wrapper object with `request` and `response` properties: https://github.com/tulios/kafkajs/blob/master/src/broker/saslAuthenticator/index.js#L42

**Expected behavior**
The TS interface should match the implementation

**Observed behavior**
Implementing a custom SASL mechanism to the TS interface will result in an error at runtime due to the unexpected shape of the arg.

markgaylard:
These got missed in the original PR when the interface changed at the end. 

Fixes #1458

### DIFF
--- a/docs/CustomAuthenticationMechanism.md
+++ b/docs/CustomAuthenticationMechanism.md
@@ -24,13 +24,17 @@ configure your brokers.
 A custom authentication mechanism needs to fulfill the following interface:
 
 ```ts
+type SaslAuthenticateArgs<ParseResult> = {
+  request: SaslAuthenticationRequest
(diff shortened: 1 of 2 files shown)


Nevon (approved):

Nevon:
Thanks for the fix!

PR #1462, opened 2022-10-14
Allow disconnect on rebalance, currently if trying to rebalance kafka…

samueltuckey:
Trying to fix the bug where if we call disconnect on kafka consumer and a rebalance is  in process then the disconnect does not work.

### DIFF
--- a/src/consumer/index.js
+++ b/src/consumer/index.js
@@ -86,15 +86,18 @@ module.exports = ({
     )
   }
 
+  let allowCrashReconnect = true
   /** @type {import("../../types").Consumer["connect"]} */
   const connect = async () => {
+    allowCrashReconnect = true
     await cluster.connect()
     instrumentationEmitter.emit(CONNECT)
   }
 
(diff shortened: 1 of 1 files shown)


trentstehbenshl (approved):

samueltuckey:
@tulios @Nevon 
This pull request will allow the consumer to call disconnect with an optional parameter allowReconnectAfterRebalance.
if allowReconnectAfterRebalance is set to false, then if the consumer calls disconnect while a rebalance is occurring then the rebalance wont automatically reconnect the consumer.

It does not change the default or affect any of the default behaviour that kafkajs provides it should not cause any problems to existing code.

Given the way the unit tests are structured I am unsure about the way to engineer a unit test. Any reccomendations would be appreciated.

typescript definitions are updated.

Please give me advice on how to get this pull request accepted.

Thanks in advance for your time.

trentstehbenshl (approved):

PR #1474, opened 2022-11-02
Consider ILLEGAL_GENERATION error as rebalancing error

### LINKED ISSUE #1009: [Question] Specified group generation id is not valid
**Describe the bug**
Consumers (sometimes) encounters the following exception: `Specified group generation id is not valid`. The consumers that encounters that error become a kind of Zombie. They are still connected as consumers to a partition but not consuming messages.

**To Reproduce**
Can't reproduce

**Expected behavior**
The consumer should reconnect to the consumer grouop.

**Observed behavior**
Logs:
1. [ConsumerGroup] Consumer has joined the group
2. The group is rebalancing, so a rejoin is needed
3. Specified group generation id is not valid

**Environment:**
 - OS: [alpine]
 - KafkaJS version [1.15.0]
 - Kafka version [2.4.1]
 - NodeJS version [14]

**Additional context**
It's probably not related to KafkaJs, but there is a mention of that error [error.js](https://github.com/tulios/kafkajs/blob/f7a166488321216a0feec4428f8e589b116eb31f/src/protocol/error.js#L145) and maybe you have any idea why it's happening?

### LINKED ISSUE #1466: "Specified group generation id is not valid" after broker maintenance, consumer stops receiving events
Hi, we are having an issue similar to https://github.com/tulios/kafkajs/issues/1009 but it happens after a broker maintenance.
We have consumers running parallelly on different machines, with a heartbeat check triggered on eachBatch.
We consume multiple topics, with a specific instance of our service per topic.
All of this works fine but we had issues (twice already) when brokers go on maintenance.
Some of the instance (thus some of the topics) stop consuming events, but don't throw errors nor crash (if it crashed we would respawn and everything would be ok).
We do see the error message:
`[Consumer] Crash: KafkaJSNonRetriableError: Specified group generation id is not valid`
But it doesn't actually crash, and the instance is stale, it won't consume any new message or trigger the heartbeat. If we restart the instance it will consume all pending traffic (given the offset is still current).
Odd thing is some of the topics keep working fine after the maintenance, so the overall system seems to be "up" unless we check each specific topic.

jakewins:
This error is indicating that the consumer is trying to commit offsets, but the consumer group has changed to a new generation.

Retrying within the existing session will indeed not work, but rejoining the group and re-trying should be successful.

Fixes #1466

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -259,7 +259,9 @@ class KafkaJSAggregateError extends Error {
 class KafkaJSFetcherRebalanceError extends Error {}
 
 const isRebalancing = e =>
-  e.type === 'REBALANCE_IN_PROGRESS' || e.type === 'NOT_COORDINATOR_FOR_GROUP'
+  e.type === 'REBALANCE_IN_PROGRESS' ||
+  e.type === 'NOT_COORDINATOR_FOR_GROUP' ||
+  e.type === 'ILLEGAL_GENERATION'
 
 const isKafkaJSError = e => e instanceof KafkaJSError
 


h0od:
I just encountered the error which causes my consumer to end up in crashed state (which sends shutdown signal to the process).
So as the PR suggests, a simple rejoin should be sufficient in this case.

rpastore-wolt:
Also we are facing this error. @jakewins do you know of you PR is going to be merged or should be used a different way to handle the error ?

kevinhering:
We're also watching this one closely. We have quite a number of kafka consumers & we recently had to create custom crash listeners to deal with just this case.

rpastore-wolt:
> We're also watching this one closely. We have quite a number of kafka consumers & we recently had to create custom crash listeners to deal with just this case.

hey @kevinhering :) thanks !
Could you share with me your solution ?

ErlendFax:
Any reason this is not merged? 

@Nevon, sorry to tag you. If you have time please consider taking a look at this and https://github.com/tulios/kafkajs/issues/1466

patrykwegrzyn:
`ILLEGAL_GENERATION` has become a pain for use recently , any idea when this will be merged anytime soon or if is going to be merged at all?

robinbijlani:
Also been watching this PR for months, very eager to see it hopefully merged in. 🤞 🤞

jgoldsmith613:
We are waiting on this too.  Any timeline on a a fix?
This is straight from confluent support
"To reduce impact for both cases above, the client will need to be designed to retry its connection to Kafka if the ILLEGAL_GENERATION exception is encountered. "

josh-cain:
We're hitting this too, and it's causing a bit of toil.  Would really appreciate a merge 🙏 !

ecalvert:
We're hitting this as well. Causing major issues when confluent cloud performs their cluster rolls. We see consumer crashes every 5-10 minutes for a 2 hour span

srMarquinho:
Just checking this issue is active and being looked after. This is exactly our problem and it is absolutely major to us. Thank you all. 🙌

Nevon:
This bug has been resolved in v2.2.4. Thanks @jakewins for the fix!

PR #1480, opened 2022-11-17
fix: dot buffer twice make sasl-auth error

ricardochen1996:
The method _request.encode()_ has returned a **Buffer** object , [[(PR)[Fix return type of request.encode to actually be a Buffer]](https://github.com/tulios/kafkajs/commit/533db2f155ceb6fcdee39e87acc67e51782b1e68)](https://github.com/tulios/kafkajs/commit/533db2f155ceb6fcdee39e87acc67e51782b1e68) ,but after ```const requestPayload = await request.encode()```，this line ```this.socket.write(requestPayload.buffer, 'binary')``` get buffer again, it will throws an unhandle exception  which makes sasl-auth error. This bug is in v2.2.0 +, maybe need **emergency** fix.

### DIFF
--- a/src/network/connection.js
+++ b/src/network/connection.js
@@ -355,7 +355,7 @@ module.exports = class Connection {
         const requestPayload = await request.encode()
 
         this.failIfNotConnected()
-        this.socket.write(requestPayload.buffer, 'binary')
+        this.socket.write(requestPayload, 'binary')
       } catch (e) {
         reject(e)
       }


Nevon:
Do you have a way to reproduce this issue? The tests are using sasl auth, so I would expect them all to fail if this was a problem, not to mention that I have services running using the latest version of KafkaJS using SASL that are behaving normally. I'm wondering if this is maybe an issue in a specific API version implementation or something, which would explain why it's apparently not breaking the tests.

ricardochen1996:
@Nevon #1483

PR #1501, opened 2022-12-08
docs: clarify wording around compression codecs

xvello:
Small wording changes in the documentation to avoid confusing new users: other codecs are readily available (and documented!), so the "might be considered in the future" wording is outdated.


xvello:
Hello @Nevon, could you please take a look at this doc PR? :bow:

PR #1516, opened 2023-01-06
Fix(producer): sending messages as string occurs error

barisx:
If messages send as string it occurs error line:72 messages.find.


Nevon:
`messages` [is defined as `Message[]`](https://github.com/tulios/kafkajs/blob/e52044de6c8b84b06804c70350316079a886825f/types/index.d.ts#L731) so passing in a string is wrong and should throw an error.

Issue #660, opened 2020-03-11
Kafkajs disconnect itself and do not attempt to reconnect


dusan-dragon:
Hello,
first of all maybe I just did not understand something properly and has wrong expectations, but I was expecting that kafkajs will do autoreconnect / autodiscovery of new brokers but instead it will just stop consuming without emmiting crash event or something.

Here are logs from our server, I removed repeating cycle of heartbeat and fetching:
```
2020-03-11T09:18:05.612Z,Kafka server has closed connection
2020-03-11T09:18:05.612Z,disconnected
2020-03-11T09:18:05.612Z,disconnecting...
2020-03-11T09:17:00.552Z,Kafka server has closed connection
2020-03-11T09:17:00.552Z,disconnected
2020-03-11T09:17:00.552Z,disconnecting...
2020-03-11T09:08:05.612Z,"Request Fetch(key: 1, version: 7)"
2020-03-11T09:08:05.612Z,Fetching from 1 partitions for 1 out of 1 topics
2020-03-11T09:08:05.611Z,"Response Heartbeat(key: 12, version: 1)"
2020-03-11T09:08:05.611Z,"Request Heartbeat(key: 12, version: 1)"
2020-03-11T09:08:05.610Z,"Response Fetch(key: 1, version: 7)"
2020-03-11T09:08:00.608Z,"Request Fetch(key: 1, version: 7)"
2020-03-11T09:08:00.606Z,Fetching from 1 partitions for 1 out of 1 topics
2020-03-11T09:08:00.605Z,"Response Heartbeat(key: 12, version: 1)"
2020-03-11T09:08:00.604Z,"Request Heartbeat(key: 12, version: 1)"
2020-03-11T09:08:00.604Z,"Response Fetch(key: 1, version: 7)"
2020-03-11T09:07:55.602Z,"Request Fetch(key: 1, version: 7)"
2020-03-11T09:07:55.602Z,Fetching from 1 partitions for 1 out of 1 topics
2020-03-11T09:07:55.602Z,"Response Heartbeat(key: 12, version: 1)"
2020-03-11T09:07:55.601Z,"Request Heartbeat(key: 12, version: 1)"
2020-03-11T09:07:55.601Z,"Response Fetch(key: 1, version: 7)"
2020-03-11T09:07:50.599Z,"Request Fetch(key: 1, version: 7)"
2020-03-11T09:07:50.598Z,Fetching from 1 partitions for 1 out of 1 topics
[...]

PR #1532, opened 2023-02-17
Ensure checkPendingRequests is run when the throttle delay is marginal

ekanth:
We have observed 10 minute pauses on consumers on fetch and it recovers after the Kafka broker resets the connection. Here are the logs from the client:
```
1:12:48.629 PM	
     correlationId: 67644
     expectResponse: true
     logger: kafkajs
     message: Request Fetch(key: 1, version: 11)
     namespace: Connection
     size: 143
```
```
1:12:48.630 PM	
     correlationId: 67644
     logger: kafkajs
     message: Request enqueued
     namespace: RequestQueue
```
```
1:22:48.632 PM	
     logger: kafkajs
     message: Kafka server has closed connection
     namespace: Connection
```
On debugging this further and reading the code, we observed that if the throttle delay is marginal (say in this case 1ms), in `push`, the request will not be sent immediately. When scheduleCheckPendingRequests is called, it computes `this.throttleUntil - Date.now()` again. When Data.now() progresses 1ms (race condition), `this.throttleUntil - Date.now()` is zero and checkPendingRequests() is not scheduled at all. Now this connection goes into a limbo until the brokers kills it based on `connections.max.idle.ms` configuration on broker (which defaults to 10 minutes).

This fixes the root cause of the same issue called out in #660.

The code changes:
- Ensure checkPendingRequests is run even when the throttle delay is marginal. When there are pending requests, we keep scheduling checkPendingRequests() periodically. I am open to changing the default interval if there are concerns.
- Ensure maybeThrottle accounts for negative values. This is extra protective and Java [client](https://github.com/apache/kafka/blob/1ed61e4090353bc9e4802ee09366f512dc60884f/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java#L854) does the same.
- (not specifically related to this bug but noticed it) Move the lock acquisition code out of try block to ensure we only release (in the finally) the lock after it is taken successfully
- Add a test that exposes the problem which passes with the fix.

### DIFF
--- a/src/network/requestQueue/index.js
+++ b/src/network/requestQueue/index.js
@@ -9,6 +9,7 @@ const PRIVATE = {
 }
 
 const REQUEST_QUEUE_EMPTY = 'requestQueueEmpty'
+const CHECK_PENDING_REQUESTS_INTERVAL = 10
 
 module.exports = class RequestQueue extends EventEmitter {
   /**
@@ -102,23 +103,14 @@ module.exports = class RequestQueue extends EventEmitter {
   }
 
   maybeThrottle(clientSideThrottleTime) {
-    if (clientSideThrottleTime) {
+    if (clientSideThrottleTime !== null && clientSideThrottleTime > 0) {
+      this.logger.debug(`Client side throttling in effect for ${clientSideThrottleTime}ms`)
       const minimumThrottledUntil = Date.now() + clientSideThrottleTime
       this.throttledUntil = Math.max(minimumThrottledUntil, this.throttledUntil)
     }
   }
 
-  /**
-   * @typedef {Object} PushedRequest
-   * @property {import("./socketRequest").RequestEntry} entry
-   * @property {boolean} expectResponse
-   * @property {Function} sendRequest
-   * @property {number} [requestTimeout]
-   *
-   * @public
-   * @param {PushedRequest} pushedRequest
-   */
-  push(pushedRequest) {
+  createSocketRequest(pushedRequest) {
     const { correlationId } = pushedRequest.entry
     const defaultRequestTimeout = this.requestTimeout
     const customRequestTimeout = pushedRequest.requestTimeout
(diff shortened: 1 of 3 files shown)


src/network/requestQueue/index.js
```diff
@@ -300,12 +309,15 @@ module.exports = class RequestQueue extends EventEmitter {
     // will be fine, and potentially fix up a new timeout if needed at that time.
     // Note that if we're merely "overloaded" by having too many inflight requests
     // we will anyways check the queue when one of them gets fulfilled.
-    const timeUntilUnthrottled = this.throttledUntil - Date.now()
-    if (timeUntilUnthrottled > 0 && !this.throttleCheckTimeoutId) {
+    let scheduleAt = this.throttledUntil - Date.now()
+    if (!this.throttleCheckTimeoutId) {
+      if (this.pending.length > 0) {
+        scheduleAt = scheduleAt > 0 ? scheduleAt : CHECK_PENDING_REQUESTS_INTERVAL
```
Nevon, line 315: Why is it that we need a minimum timeout here? I'm thinking that if the throttle time was marginal, say 1ms, and 1ms has elapsed by the time we check it, then `this.throttledUntil - Date.now()` will be either zero or negative. At that point, couldn't we just set the timeout value to `Math.max(scheduleAt, 0)`? That would just result in the pending check request happening on the next turn around the event loop, which is what we want in this case since the throttle period has already ended and we don't want to delay the requests further.

The only problem I see with that is that we need to make sure `this.throttledUntil` is cleared after we've scheduled the check, so that we don't end up always doing `setTimeout(..., 0)` as soon as we've been throttled once.

Does that make sense?

ekanth: Thanks @Nevon for the quick review!

Doing `Math.max(scheduleAt, 0)` has the exact problem you mentioned - checkPendingRequests() and scheduleCheckPendingRequests() keep calling each other in a loop without an exit criteria. Given pending is tracking requests both because of throttling and max inflight, I felt the simplest way to implement this would be to use a minimum timeout that kicks in when pending is not empty and stop otherwise.

ekanth: @Nevon - would love to hear if you have any follow-up thoughts on the approach for this fix.

ekanth:
@Nevon - I'd appreciate another look and help in merging this PR. This issue has been happening to us and would love to merge this and have an official release that fixes this.

src/network/requestQueue/index.js
Nevon: I'm not sure I'm sold on the idea of introducing up to a 10ms delay whenever there's been marginal throttling, but it's still an improvement over the current situation. This can always be improved if it turns out that it actually happens with some amount of frequency.

Issue #1556, opened 2023-03-20
CPU hike on v2.2.4


Hyjaz:
**Describe the bug**
When upgrading to v2.2.4 we saw a 190% increase in our CPU usage. 

**To Reproduce**
Not sure how you can reproduce it. We do however send thousand of messages per second. 

If none of the above are possible to provide, please write down the exact steps to reproduce the behavior:
1. Run a producer that continuously produces messages to a topic

**Expected behavior**
No cpu usage hike between v2.2.3 and v.2.2.4

**Observed behavior**
A clear and concise description of what *did* happen. Please include any relevant logs [with the log level set to debug](https://kafka.js.org/docs/configuration#logging).

**Environment:**
 - OS: N/A
 - KafkaJS version: 2.2.4
 - Kafka version: 
 - NodeJS version:16.18.0

**Additional context**

Nevon:
Do you have a CPU profile that could show where CPU time is being spent? You can use something like [0x](https://github.com/davidmarkclements/0x#readme) to generate a flame graph, if you don't instrument your application with some APM solution.

Hyjaz:
Hello @Nevon, it seems to be coming from the scheduleCheckPendingRequest. I noticed that there was a change related to this in the requestQueue/index.js in the latest release. Let me know if you need more details.

<img width="1518" alt="Screenshot 2023-03-21 at 08 41 31" src="https://user-images.githubusercontent.com/16614359/226609189-3c7adf15-81c9-40a7-80fd-19ecd2ca0c5e.png">

Hyjaz:
This is v2.2.3. You can see there is a huge hump in the cpu usage between both versions.

<img width="1516" alt="Screenshot 2023-03-21 at 09 06 00" src="https://user-images.githubusercontent.com/16614359/226614836-fb97f69f-849c-43c7-8b3b-dfb6bff93a14.png">

JoseGoncalves:
Also noticed this CPU increase in my app.
In idle v2.2.3 uses almost no CPU, while v2.2.4 uses around 8%.

MDSLKTR:
I kinda want to have the best of 2.2.3 and 2.2.4 but the CPU spike is way too much to upgrade currently, which is why we pinned it to 2.2.3

siimsams:
[...]

PR #1572, opened 2023-05-08
fix regression when throttle timeout is marginal

MDSLKTR:
Fixes
https://github.com/tulios/kafkajs/issues/1556

Tries to target the regression that was caused by the mentioned issue. The change to check whether the throttleTimeout is actually greater than 0 is readded. To cover for cases the result of the `timeUntilUnthrottled` calculation was actually off by a small margin might be fixed by storing the timestamp when the throttleing was applied. And after the check runs, updating it. Please let me now if i missed something as this is my first draft for this. The test that was added works with this but it might still be lacking something.

### DIFF
--- a/src/network/requestQueue/index.js
+++ b/src/network/requestQueue/index.js
@@ -323,7 +323,7 @@ module.exports = class RequestQueue extends EventEmitter {
         this.throttleCheckTimeoutId = null
         this.throttleCurrentTimestamp = Date.now()
         this.checkPendingRequests()
-      }, Math.max(timeUntilUnthrottled, 0))
+      }, timeUntilUnthrottled)
     }
   }
 }


src/network/requestQueue/index.js
```diff
@@ -309,15 +316,14 @@ module.exports = class RequestQueue extends EventEmitter {
     // will be fine, and potentially fix up a new timeout if needed at that time.
     // Note that if we're merely "overloaded" by having too many inflight requests
     // we will anyways check the queue when one of them gets fulfilled.
-    let scheduleAt = this.throttledUntil - Date.now()
-    if (!this.throttleCheckTimeoutId) {
-      if (this.pending.length > 0) {
-        scheduleAt = scheduleAt > 0 ? scheduleAt : CHECK_PENDING_REQUESTS_INTERVAL
-      }
+    const timeUntilUnthrottled = this.throttledUntil - this.throttleCurrentTimestamp
```
Nevon, line 319: I feel like I'm missing something here. From reading this code it looks to me like we're going to be checking every single time around the event loop, since the `setTimeout` is always called with 0.

So the very first time we get throttled, we'll set `throttleCurrentTimestamp` to 0 and `throttledUntil` to 10. `timeUntilThrottled` is `10 - 0`, which is greater than 0 and we don't already have a scheduled timer, so we schedule it for next tick. Next tick happens so we null out the timeout id and we update `throttleCurrentTimestamp` with the current time, let's say it's 1. Now `timeUntilThrottled` is 10 - 1 which is still greater than 0, so we do the same thing over again.

MDSLKTR: you are right, this would probably create many more calls then necessary, i could imagine that instead using Math.max(timeUntilUnthrottled, 0) would then be more applicable?

Nevon: It might be helpful to read the previous implementation (and the original one). If you check the PR that introduced this (https://github.com/tulios/kafkajs/pull/1532#discussion_r1110809796), I made a similar suggestion. We just need to make sure we don't end up in an infinite loop

MDSLKTR: Is there a way to move this forward? so far it looks good and not causing any infinite loops

siimsams:
@Nevon can you take another look? This solved our event loop issues in prod.

Hyjaz:
@Nevon I would greatly appreciate it if you could review this pull request once again and ultimately approve and merge it. We are currently facing errors that have already been fixed in version 2.2.4, but unfortunately, that version is not currently usable. Thank you!

src/network/requestQueue/index.js
allanpaiste: Ought there to be a test that would guard against this kind of performance degradation to happen in the future?

How we ended up bumping into this was that there was an integration test that did time jumps with Sinon FakeTimer where it was clear that the internals of FakeTimer ended up throttling the actual Event Loop due to this setTimeout (and another setInterval) causing there to be a huge backlog of timer callbacks. Not saying that this is the strategy that one ought to use to guard against this, but that would indirectly create somewhat easy way of verifying that there are no overly-aggressive setTimeout-zero loops in the system.

src/network/requestQueue/index.js
```diff
@@ -309,15 +316,14 @@ module.exports = class RequestQueue extends EventEmitter {
...
+
+    if (timeUntilUnthrottled > 0 && !this.throttleCheckTimeoutId) {
       this.throttleCheckTimeoutId = setTimeout(() => {
         this.throttleCheckTimeoutId = null
+        this.throttleCurrentTimestamp = Date.now()
         this.checkPendingRequests()
-      }, scheduleAt)
+      }, Math.max(timeUntilUnthrottled, 0))
```
allanpaiste, line 326: Considering the conditional above ( > 0), this Math.max is probably not needed.

allanpaiste:
Ought there to be a test that would guard against this kind of performance degradation to happen in the future?

How we ended up bumping into this was that there was an integration test that did time jumps with Sinon FakeTimer where it was clear that the internals of FakeTimer ended up throttling the actual Event Loop due to this setTimeout (and another setInterval) causing there to be a huge backlog of timer callbacks. Not saying that this is the strategy that one ought to use to guard against this, but that would indirectly create somewhat easy way of verifying that there are no overly-aggressive setTimeout-zero loops in the system.

src/network/requestQueue/index.js
siimsams, line 326:
```suggestion
      }, timeUntilUnthrottled)
```

siimsams:
As this has been open since May we have decided to abandon hope and downgrade to v1.6 to not have these issues. :(

siimsams (approved):

MDSLKTR:
i feel like this PR was used to vent a lot, however in order to contribute, i think it makes sense to move this forward and release it as a patch version

ekanth (approved):
Thanks @MDSLKTR! I caught up on this change and the fix sounds reasonable to me. @Nevon - @MDSLKTR has addressed your earlier comments. While we understand that you might be busy, really appreciate if you could merge this fix and release 2.2.5. It looks like a lot of people are waiting for this one! 🙏🏼

MDSLKTR:
Any update here?

MDSLKTR:
Feel free to try out installing: https://github.com/MDSLKTR/kafkajs.git#2.2.5, in case that helps your case

sefiros1:
@MDSLKTR thank a lot for your work! How soon can we expect this change in the official release? It seems to me that the previous reviewer (@Nevon) has gone inactive it this thread. @ekanth may you help somehow to push this through?

MDSLKTR:
Our team will be switching back to node-rdkafka while keep in mind that the official client will be ready at some point https://github.com/confluentinc/confluent-kafka-javascript
So i wont pursue this any furthee

Issue #818, opened 2020-07-20
Support KIP-320: Handle log truncation


Nevon:
**Is your feature request related to a problem? Please describe.**

#778 introduces support for Fetch V9, which adds support for new errors to communicate to a consumer that they are trying to fetch from a broker which is not in sync with the leader. #778 only added support for the new API version, but not the functionality that it enables.

> When unclean leader election is enabled, we may lose committed data. A consumer which is reading from the end of the log will typically see an out of range error, which will cause it to use its auto.offset.reset policy. To avoid losing data, users should use the "earliest" option, but that means consuming the log from the beginning.
> 
> It is also possible that prior to sending the next fetch, new data is written to the log so that the consumer's fetch offset becomes valid again. In this case, the consumer will just miss whatever data had been written between the truncation point and its fetch offset.
> 
> Neither behavior is ideal, but we tend to overlook it because the user has opted into weaker semantics by enabling unclean leader election. Unfortunately in some situations we have to enable unclean leader election in order to recover from serious faults on the brokers. Some users have also opted to keep unclean leader election enabled because they cannot sacrifice availability ever. We would like to offer better client semantics for these situations.

**Describe the solution you'd like**

> The proposal in this KIP is to have the consumer behave more like a follower. The consumer will obtain the current leader epoch using the Metadata API. When fetching from a new leader, the consumer will first check for truncation using the OffsetForLeaderEpoch API. In order to enable this, we need to keep track of the last epoch that was consumed. If we do not have one (e.g. because the user has seeked to a particular offset or because the message format is older), then the consumer will skip this step. To support this tracking, we will extend the OffsetCommit API to include the leader epoch if one is available.
> 
> Leader changes are detected either through a metadata refresh or in response to a FENCED_LEADER_EPOCH error. It is also possible that the consumer sees an UNKNOWN_LEADER_EPOCH in a fetch response if its metadata has gotten ahead of the leader. 
> 
[...]

PR #1594, opened 2023-06-24
Implement static membership feature on client side (KIP-345)

sandeepkumarcel:
implement kip-345

add test fixtures

ts-defs


index.js
```diff
@@ -12,6 +12,8 @@ const ResourcePatternTypes = require('./src/protocol/resourcePatternTypes')
 const { isRebalancing, isKafkaJSError, ...errors } = require('./src/errors')
 const { LEVELS } = require('./src/loggers')
 
+// local two
```
distracteddev, line 15: errant comment maybe?

distracteddev:
This is exciting work. It seems like this has also implemented most of (all of?) KIP-320.

Can you help me understand what parts of Nevon's earlier [checklist](https://github.com/tulios/kafkajs/issues/884#issuecomment-768870422) might still be left to complete?

> 1. Implement [KIP-320](https://cwiki.apache.org/confluence/display/KAFKA/KIP-320%3A+Allow+fetchers+to+detect+and+handle+log+truncation). This is probably gonna be an epic in and of itself. The reason we need this is because we need to support Fetch v12 in order to get the currentLeaderEpoch field, which we need in order to support OffsetFetch v6-7, which in turn we need because it includes the groupInstanceId field that's needed for KIP-345.
> 2. Add support for OffsetFetch v7
> 3. Change the behavior of a consumer leaving the group to not issue a LeaveGroup request if the instance id is set.
> 4. Expose a way of configuring the instance id on the consumer.
> 5. Update the partition assigner to take into account the instance id (the KIP has a good example of how)
I have not really mapped out the work for KIP-320, so I don't know how big or small it is, but it's definitely the first step needed. We have another issue open for that https://github.com/tulios/kafkajs/issues/818

Rez0k:
can we advance this pr?
we can really use this ability

Jasper-DoubleMe-Me:
merge please

swan2488:
This feature will be very helpful in our case.
Is there any way to speed up the pr

abbccdda:
Same here, this is really useful.

olfedorenk:
This pr will help us a lot

usbtrs:
Agree, that this feature will help many developers

Gekeris:
Please don't delay this pr. I need it

larysakutyrkina:
+1, would help a lot with a current project

src/consumer/consumerGroup.js
```diff
@@ -182,13 +187,14 @@ module.exports = class ConsumerGroup {
...
+    this.groupInstanceId = groupData.groupInstanceId
     this.groupProtocol = groupData.groupProtocol
   }
 
   async leave() {
-    const { groupId, memberId } = this
+    const { groupId, memberId, groupInstanceId } = this
     if (memberId) {
-      await this.coordinator.leaveGroup({ groupId, memberId })
+      await this.coordinator.leaveGroup({ groupId, memberId, groupInstanceId })
```
abbccdda, line 197: We should not send leave group if static membership is enabled, i.e groupInstancId is defined.

abbccdda (changes requested):

Exc1usive:
@sandeepkumarcel check please comments, realy waiting for this PR

src/consumer/consumerGroup.js
diyorbekrustamjonov: Can't you change it yourself? We're just waiting for this feature and no one is reacting!

priyesh2609: @abbccdda It is not necessary that anybody who uses static membership would never want to stop the graceful termination of consumption. This is a valid change. If one does not want to call leave group then just have to not call the consumer.stop() & consumer.disconnect() & leave group won't be called.

priyesh2609:
@sandeepkumarcel have you tested this PR somewhere? if yes, then let us know, at least I am ok to use your forked version in my use case.
Thanks 🙇

src/consumer/consumerGroup.js
benthayer: Hey it seems like a lot of people are waiting for this (including me). I'd be willing to take this on since it seems like it's so close to being done. What's there left to do? Just this?

vputsenko: is that the only one blocker for this PR? @abbccdda  @tulios  ? If yes then looks like two years we are waiting for someone to remove that line or put it under if statement ?)

PR #1627, opened 2023-09-29
[KIP-601] Increase connection timeout

emasab:
being closer to the defaults in KIP-601.

In KafkaJS the connection establishment limit corresponds to the `connectionTimeout` + `authenticationTimeout`.

`authenticationTimeout` has been increased to 10s, but `connectionTimeout` is 1s.

KIP-601 `socket.connection.setup.timeout.ms` is about connection and in librdkafka corresponds to 30s, while in kafka-clients is exponential from a minimum of `socket.connection.setup.timeout.ms` (10s) to a maximum of `socket.connection.setup.timeout.max.ms` (30s).

This reduces the gap by setting the minimum to 10s and the sum of them to 20s.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -40,7 +40,7 @@ module.exports = class Client {
    * @param {Object} options.ssl
    * @param {Object} options.sasl
    * @param {string} options.clientId
-   * @param {number} [options.connectionTimeout=1000] - in milliseconds
+   * @param {number} [options.connectionTimeout=10000] - in milliseconds
    * @param {number} options.authenticationTimeout - in milliseconds
    * @param {number} options.reauthenticationThreshold - in milliseconds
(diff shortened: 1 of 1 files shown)


stanislavkozlovski:
Thanks for this PR @emasab 

I think it's important to bump this limit too @Nevon - otherwise it makes the client too fragile to connection storm issues - we've seen cases where clients connect to brokers at once and due to either throttling or simple overload of the acceptor thread in Kafka, the broker can take more than a second to establish the connection.

If clients are too sensitive to this timeout, they risk perpetuating the issue

stanislavkozlovski (commented):
We should edit this excerpt too: https://github.com/tulios/kafkajs/blob/master/docs/Configuration.md#connection-timeout

stanislavkozlovski:
Is this project still actively maintained?

PR #1631, opened 2023-10-18
Range partition assigner

### LINKED ISSUE #1564: Add range strategy assigner
By default there is only RoundRobin strategy. Is it possible to implement others strategy like range ?

It would be great !

pimpelsang:
Hi,

Here is range partition assigner using the same algorithm as in java client. Please review :)

closes #1564


src/consumer/assigners/rangeAssigner/index.js
```diff
@@ -0,0 +1,77 @@
...
+      const consumersWithExtraPartition = numPartitionsForTopic % membersCount
+
+      for (var i = 0; i < membersCount; i++) {
+        const start = numPartitionsPerConsumer * i + Math.min(i, consumersWithExtraPartition)
+        const length = numPartitionsPerConsumer + (i + 1 > consumersWithExtraPartition ? 0 : 1)
+
+        const assignee = sortedMembers[i]
+
+        for (let partition = start; partition < start + length; partition++) {
+          if (!assignment[assignee]) {
```
Saeger, line 46: I think we have 3 levels of for loop which doesn't seem ideal, on the other hand, this is only used once assigning members to the consumers right? So it should be fine I guess. I wanted to ask if you have reasoned about this before as well?

pimpelsang: Yes, same flow in java client. There isn't that much data to loop through and like you said it happens once. I see no reason to optimise.

Saeger: LGTM

Saeger (approved):

yduman:
What is preventing this PR from being merged? Our team is currently migrating from [node-rdkafka](https://github.com/Blizzard/node-rdkafka) to KafkaJS and since the `partition.assignment.strategy` is set by default to `[range, roundrobin]` in [librdkafka](https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md), we are getting a `KafkaJSProtocolError` pointing to `InconsistentGroupProtocolException`, because KafkaJS uses by default only `roundrobin`. This PR would allow to fix this problem, where users can import both strategies for `partitionAssigners`. The alternative, without this PR, would be to copy the implementation of this solution and introduce it as a class manually by yourself.

jaredbarranco:
Similar to @yduman my team is migrating from a Java monolith to NodeJS microservices for async etl work via Kafka. The Java library by default uses range, and kafkajs uses roundRobin. Would be nice to have a seamless cutover.

Issue #957, opened 2020-11-12
Provide additional metadata in errors


eliw00d:
**Is your feature request related to a problem? Please describe.**
Currently, there are many errors that do not provide certain metadata. For example, TOPIC_AUTHORIZATION_FAILED does not provide the topic(s) that failed to authorize. We provide our producer with multiple brokers and use sendBatch to send to multiple topics. It appears that some messages go through to one broker but not another but a lot of research needs to be done to figure out the problematic topic(s). It would be nice if the error message provided more information for diagnosing issues.

**Describe the solution you'd like**
For error messages related to producing or consuming from topics it would be useful to have the topics that caused the error in the error.

Nevon:
This is a good idea and something that can be done incrementally on a case-by-case basis.

The way that protocol errors work is that whenever we receive a response from the API, there can be one or more error codes returned. For example, the response for CreateTopics v0 looks like this:

```
CreateTopics Response (Version: 0) => [topics] 
  topics => name error_code 
    name => STRING
    error_code => INT16
```

This can be read as an array of `topics`, where each `topics` has two properties: `name` (`string`) and `error_code` (`int16`).

The error code [can then be looked up in this list](https://kafka.apache.org/protocol#protocol_error_codes). So if we for example receive the number 29, that corresponds to TOPIC_AUTHORIZATION_FAILED.

In most cases, this is handled generically using the `createErrorFromCode` function:

https://github.com/tulios/kafkajs/blob/28e5be5e292dca853ce075feb242a76df5478a04/src/protocol/error.js#L580-L582

`CreatTopics` is a typical example of how this is done:

https://github.com/tulios/kafkajs/blob/28e5be5e292dca853ce075feb242a76df5478a04/src/protocol/requests/createTopics/v0/response.js#L25-L32

In this case, we are simply checking all the `topics` in the returned list, and if there's any error, we throw the first one we find. Obviously we could do a much better job by creating a specific error class that extends the protocol error with more information, and also aggregates all of the errors rather than just picking the first one.

prithvijit-dasgupta:
@Nevon From what I gather from the issue it seems like that each of the protocol calls would need a separate wrapped error class (depending on the use case).
I am very much interested in helping with this project as I will be using this library in our company's Kafka consumers.

Nevon:
[...]

Issue #129, opened 2018-09-17
Include topic & partition in UNKNOWN_TOPIC_OR_PARTITION error


Nevon:

theliuk:
I might take this.

shintani0607:
Has this issue been resolved yet? Everyone give me a solution for this issue...

sujayneglur123:
I am picking this issue. Fork created at https://github.com/sujayneglur123/kafkajs

mgirard772:
👍 to see this resolved, this error message is not very helpful without this information.

dhananjaysa92:
looking into it.

atiquefiroz:
has this been planned to resolve?

PR #1646, opened 2023-12-03
Create delete topic error structure

ivanvs:
Connected to https://github.com/tulios/kafkajs/issues/957 and https://github.com/tulios/kafkajs/issues/129

Add information which topics cannot be deleted when user tries to delete topics.

1. Add KafkaJSDeleteTopicError that is extending KafkaJSProtocolError
2. Modify v0 parser for deleteTopics to handle errors better
3. Modify admin API to deal with new deleteTopics error structure

### DIFF
--- a/src/admin/index.js
+++ b/src/admin/index.js
@@ -272,7 +272,7 @@ module.exports = ({
       } catch (e) {
         if (
           e.name === 'KafkaJSAggregateError' &&
-          e.errors.some(error => error.type.includes('UNKNOWN_TOPIC_OR_PARTITION'))
+          e.errors.some(error => error.type === 'UNKNOWN_TOPIC_OR_PARTITION')
         ) {
           logger.warn('Could not delete topics', {
             error: e.message,
@@ -283,7 +283,7 @@ module.exports = ({
           throw e
         }
 
-        if ('NOT_CONTROLLER'.includes(e.type)) {
+        if (e.type === 'NOT_CONTROLLER') {
           logger.warn('Could not delete topics', {
             error: e.message,
             retryCount,


src/admin/index.js
```diff
@@ -270,8 +270,25 @@ module.exports = ({
...
+          logger.warn('Could not delete topics', {
+            error: e.message,
+            retryCount,
+            retryTime,
+            topics: e.errors.map(error => error.topic),
+          })
+          throw e
+        }
+
+        if ('NOT_CONTROLLER'.includes(e.type)) {
```
gfoltz, line 286: Same here, this should be `e.type === 'NOT_CONTROLLER'`.

src/admin/index.js
```diff
@@ -270,8 +270,25 @@ module.exports = ({
 
         await cluster.refreshMetadata()
       } catch (e) {
-        if (['NOT_CONTROLLER', 'UNKNOWN_TOPIC_OR_PARTITION'].includes(e.type)) {
-          logger.warn('Could not delete topics', { error: e.message, retryCount, retryTime })
+        if (
+          e.name === 'KafkaJSAggregateError' &&
+          e.errors.some(error => error.type.includes('UNKNOWN_TOPIC_OR_PARTITION'))
```
gfoltz, line 275: This should be `error => error.type === 'UNKNOWN_TOPIC_OR_PARTITION'`.  The use of includes in the original code was on an Array. But here you're using it on a string, which works but I think only accidentally, and I'm pretty sure is not what you intend.

src/admin/index.js
ivanvs: @gfoltz  you are totally right. I have changed how error type is checked.

Thank you.

src/admin/index.js
ivanvs: This one is also changed.

Thank you.

PR #1657, opened 2024-01-06
Include the offending topic when metadata request returns unknown or unauthorized

gfoltz:
This fixes  https://github.com/tulios/kafkajs/issues/957 and https://github.com/tulios/kafkajs/issues/129

A generic `KafkaJSProtocolError` is thrown when a kafka client tries to use a topic that is unknown to the broker or unauthorized for the client. Including the topic name in these cases is very useful for debugging because it allows the engineer to more rapidly know what needs to change with their cluster.

This PR includes the following:
1. A new `KafkaJSUnknownTopic` error which includes a `topic` property
2. A new `KafkaJSTopicAuthorizationFailed` error which includes a `topic property
3. When logging is enabled, the `error.topic` value is included in the log context with the log line.

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -38,6 +38,21 @@ class KafkaJSOffsetOutOfRange extends KafkaJSProtocolError {
   }
 }
 
+class KafkaJSUnknownTopic extends KafkaJSProtocolError {
+  constructor(e, { topic }) {
+    super(e, { retriable: false })
+    this.topic = topic
+    this.name = 'KafkaJSUnknownTopic'
+  }
+}
+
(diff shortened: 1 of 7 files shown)


TiMESPLiNTER:
Thanks a lot for your work. We're also missing something like this. So we can see what permission is missing for which topic. I really hope this one will make it into a a release asap.

gfoltz:
@Nevon @tulios Can you approve?

DjellalAbdou:
I was doing literally this PR before finding it here, I think it is a really good thing to add ASAP

PR #1663, opened 2024-02-01
Update Consuming.md seek documentation

CalderWhite:
The type offset is expecting is a string, but in the documentation it is a number. Funny enough, it is provided as a string afterwards. I changed it be a string both times for consistency and to make the sample runnable in typescript.

Screenshot 2024-02-01 at 2 28 36 PM


CalderWhite:
Tagging @Nevon to TAL :)

Issue #1737, opened 2025-02-18
Improve performance of arrayDiff


mhahnenberg:
**Is your feature request related to a problem? Please describe.**
We've observed arrayDiff in some profiling we've done in production when subscribing to a topic with a nontrivial number of partitions (e.g. > 1000) with a relatively high fetch rate. After checking out the code it appears that arrayDiff is written in a way that is O(n*m) since it's iterating over `a` and calling `b.indexOf()` for each element. 

**Describe the solution you'd like**
Using a Set could be an easy way to make each of the checks inside the loop O(1) instead of O(n) with indexOf, thus making the overall function O(n+m) rather than O(n*m).

**Additional context**
I've written a simple PR that does this, will attach after filing this issue.

mhahnenberg:
PR for a simple fix is #1738

PR #1738, opened 2025-02-18
perf(utils): Use Set to make arrayDiff O(n+m) vs O(n*m)

mhahnenberg:
This is used in a couple places where the values for n and m could be medium size (e.g. when checking for new partitions within a topic), and we've seen it appear in some of our CPU profiling. So making this linear instead of quadratic should help slightly when checking topics with higher partition counts.


mhahnenberg:
Proposed fix for #1737

omid-mamandi (approved):

ousiaresearch:

StantonMatt (commented):
I took a focused pass over this since `arrayDiff` is used in partition/topic bookkeeping paths.

The change looks correct for the current call sites. I checked the helper users in `consumerGroup` and `brokerPool`; they pass topic names, broker ids, or partition ids. For representative primitive inputs, the `Set` version matches the old `indexOf` behavior while avoiding the repeated scan through `b`.

Local checks I ran:

- `yarn install --frozen-lockfile`
- `yarn jest src/utils/arrayDiff.spec.js --runInBand --no-watchman`
- `./node_modules/.bin/eslint src/utils/arrayDiff.js src/utils/arrayDiff.spec.js`
- `./node_modules/.bin/prettier --list-different src/utils/arrayDiff.js src/utils/arrayDiff.spec.js`
- `git diff --check origin/master...HEAD`

I also ran a small direct Node comparison against the old implementation for numbers, strings with duplicates, `null`, and `undefined`; those matched. The only semantic difference I found is the expected `NaN` edge (`indexOf` does not match `NaN`, `Set#has` does), but I do not see a KafkaJS call site where `NaN` is a meaningful array member here.

No upstream checks are reported on this PR branch, so I am only claiming the local utility-level checks above.

PR #1768, opened 2025-12-26
Fix scheduleAt calculation to prevent negative values

### LINKED ISSUE #1751: TimeoutNegativeWarning emitted when running on Node 24.0.1
**Describe the bug**
When running on Node 24.0.1, a `TimeoutNegativeWarning` is emitted when the client makes a connection. The timeout value appears to be a negative timestamp, presumably resulting from [this subtraction](https://github.com/tulios/kafkajs/blob/v2.2.4/src/network/requestQueue/index.js#L312).

**To Reproduce**

With a trivial app:
```javascript

const kafka = new Kafka({
  brokers: ['localhost:8092']
});

producer.connect();
```

Run with warnings enabled:
```shell
KAFKAJS_NO_PARTITIONER_WARNING=1 node --trace-warnings index.js
```

Creating a producer, consumer, or admin client all cause the same emitted warning.

**Expected behavior**
No warnings or errors emitted.

**Observed behavior**
```
(node:25173) TimeoutNegativeWarning: -1747359280440 is a negative number.
Timeout duration was set to 1.
    at new Timeout (node:internal/timers:195:17)
    at setTimeout (node:timers:120:19)
    at RequestQueue.scheduleCheckPendingRequests (/tmp/kafkajs-warning/node_modules/kafkajs/src/network/requestQueue/index.js:317:37)
    at RequestQueue.checkPendingRequests (/tmp/kafkajs-warning/node_modules/kafkajs/src/network/requestQueue/index.js:297:10)
    at RequestQueue.fulfillRequest (/tmp/kafkajs-warning/node_modules/kafkajs/src/network/requestQueue/index.js:206:10)
    at Connection.processData (/tmp/kafkajs-warning/node_modules/kafkajs/src/network/connection.js:529:25)
    at Socket.onData (/tmp/kafkajs-warning/node_modules/kafkajs/src/network/connection.js:184:14)
    at Socket.emit (node:events:507:28)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
```

**Environment:**
 - OS: MacOS 13.7.2
 - KafkaJS version: 2.2.4
 - Kafka version: 3.7.2
 - NodeJS version: 24.0.1

robertrossmann:
This avoids annoying "negative setTimeout value" warnings in Node.js v24 and above. If `throttledUntil` is `-1` (default) then subtracting current unix timestamp would result in negative number.

Fixes #1751


danieleloscozzese (approved):

rmartins0:
Will this ever get merged? Is this library dead?

tejas-contentstack:
@danieleloscozzese, is anything preventing the merge of this PR?

danieleloscozzese:
I don't have any authority here: I approved it because I could and hoping it would send a signal, but I don't have any rights on this repo.

BigButovskyi:
I think the proposed approach suppresses the warning but leaves the underlying infinite 1ms polling loop intact — the node will still burn CPU cycles spinning in circles doing nothing every 1ms.

I think this alternative is better:

```js
if (!this.throttleCheckTimeoutId) {
  if (this.pending.length > 0) {
    scheduleAt = scheduleAt > 0 ? scheduleAt : CHECK_PENDING_REQUESTS_INTERVAL
  } else if (scheduleAt <= 0) {
    return
  }
}
```
Since it kills the useless loop entirely.

ousiaresearch:

rmartins0:
@Nevon  ?

StantonMatt (commented):
I checked this locally because it is tied to the Node 24 `TimeoutNegativeWarning` reports in #1751. The one-line clamp removes the negative timeout value, but I think it still leaves the underlying no-work timer loop in place.

On `origin/master`, calling `scheduleCheckPendingRequests()` with no pending requests and no active throttle schedules a timeout with a large negative delay. On this PR branch, the same probe schedules a timeout with delay `0`:

```text
origin/master: pending=0, throttledUntil=-1, timeoutCalls=[-1780380366736]
pr-1768:      pending=0, throttledUntil=-1, timeoutCalls=[0]
```

That means `setTimeout(..., 0)` still runs when there is no pending request to flush. When it fires, `checkPendingRequests()` calls `scheduleCheckPendingRequests()` again, so the warning is suppressed but the queue can still keep an immediate timer alive. The focused request-queue suite also showed the symptom: `yarn jest src/network/requestQueue/index.spec.js --runInBand --no-watchman` passed 18/18 tests, then Jest did not exit because an async handle was still open.

The smallest shape I would expect is to return before scheduling when `pending.length === 0` and the computed throttle delay is not in the future. The `CHECK_PENDING_REQUESTS_INTERVAL` fallback still makes sense when there are pending requests but the throttle has already expired.

Other local checks on the touched file passed:

```text
./node_modules/.bin/eslint src/network/requestQueue/index.js
./node_modules/.bin/prettier --list-different src/network/requestQueue/index.js
git diff --check origin/master...HEAD
```

The visible Azure checks on this PR are also still red (`Admin`, `Broker`, `Consumer`, `OauthBearer`, `Others`, `Producer`, and the aggregate build), so I would not treat the branch as merge-ready until this timer case is adjusted and those jobs are rerun or explained.

PR #1779, opened 2026-05-19
Handle GROUP_LOAD_IN_PROGRESS and re-discover coordinator after NOT_COORDINATOR_FOR_GROUP

shailesh:
## Summary

Two related consumer-group bugs surfaced under broker churn (rolling restarts, MSK/HF failover):

1. **`GROUP_LOAD_IN_PROGRESS` (error code 14) is not treated as a rebalance.** When a coordinator is loading the `__consumer_offsets` partition for a group, it returns this error to `JoinGroup`/`SyncGroup`/`OffsetCommit`. Today it bubbles up as a hard failure and the consumer crashes the join cycle instead of retrying.
2. **`OffsetManager` does not re-discover the coordinator on coordinator change.** Two paths assume the cached `coordinator` is still authoritative:
   - `getCoordinator()` calls `cluster.findBroker(this.coordinator)` when the broker is disconnected — that just looks up the *same* (now-stale) broker by node id.
   - The `commitOffsets()` `NOT_COORDINATOR_FOR_GROUP` catch calls `refreshMetadata()` but never asks the cluster who the new coordinator actually is.

   In both cases the next `OffsetCommit` is sent to the wrong broker, which replies with `NOT_COORDINATOR_FOR_GROUP` again, and the loop keeps stalling until something else triggers a rejoin.

## Changes

- `src/errors.js` — add `GROUP_LOAD_IN_PROGRESS` to `isRebalancing()`. The existing rebalance retry path in `joinAndSync` already handles `REBALANCE_IN_PROGRESS` / `NOT_COORDINATOR_FOR_GROUP` / `ILLEGAL_GENERATION`; this just teaches it that "coordinator is still loading group state" is the same kind of transient.
- `src/consumer/offsetManager/index.js`:
  - `getCoordinator()` now calls `cluster.findGroupCoordinator({ groupId })` instead of `cluster.findBroker(this.coordinator)` so a disconnected/stale coordinator gets *replaced*, not re-fetched by id.
  - `commitOffsets()` `NOT_COORDINATOR_FOR_GROUP` handler now re-discovers the coordinator after `refreshMetadata()` so the next attempt targets the right broker.

## Tests

- `src/errors.spec.js` — covers all four `isRebalancing` cases (existing three + new `GROUP_LOAD_IN_PROGRESS`) and negative cases.
- `src/consumer/offsetManager/__tests__/getCoordinator.spec.js` — covers connected coordinator (no lookup), disconnected coordinator (uses `findGroupCoordinator`), and the stale-coordinator-after-failover case where `findBroker` would have returned the wrong broker.
- `src/consumer/offsetManager/__tests__/commitOffsets.spec.js` — adds one test asserting `findGroupCoordinator` is called on `NOT_COORDINATOR_FOR_GROUP`.

Locally: `eslint` clean and `jest src/errors.spec.js src/consumer/offsetManager/__tests__/` is green (27/27).

## Background

Hit in production on a Kafka cluster behind a managed-broker rolling restart. Symptom was consumers stuck in a rebalance loop logging `NOT_COORDINATOR_FOR_GROUP` repeatedly, plus join failures during periods when the coordinator was loading group state (`GROUP_LOAD_IN_PROGRESS`). Both went away with this patch.

## Test plan

- [x] `eslint` clean for changed files
- [x] Unit tests pass for `errors.spec.js`, `offsetManager/__tests__/*`
- [ ] Reviewer to confirm `cluster.findGroupCoordinator({ groupId })` is the correct primitive (matches what `consumerGroup.js` already uses on initial join)


ousiaresearch:

StantonMatt (commented):
I checked the coordinator path locally because the main question here is whether `findGroupCoordinator({ groupId })` is the right primitive after `NOT_COORDINATOR_FOR_GROUP`.

That looks right to me. The existing consumer-group join path already discovers the coordinator with `cluster.findGroupCoordinator({ groupId })`; using `findBroker(this.coordinator)` in `OffsetManager` can just reattach to the previous broker id, which is exactly the stale-coordinator case this retry path needs to escape.

Local checks I ran on `233f225f`:

```text
YARN_CACHE_FOLDER=.codex-tmp/yarn-cache yarn test:types
YARN_CACHE_FOLDER=.codex-tmp/yarn-cache yarn jest src/errors.spec.js src/consumer/offsetManager/__tests__/commitOffsets.spec.js src/consumer/offsetManager/__tests__/getCoordinator.spec.js --runInBand --no-watchman
YARN_CACHE_FOLDER=.codex-tmp/yarn-cache yarn jest src/consumer/offsetManager/__tests__ src/errors.spec.js --runInBand --no-watchman
YARN_CACHE_FOLDER=.codex-tmp/yarn-cache yarn eslint src/errors.js src/errors.spec.js src/consumer/offsetManager/index.js src/consumer/offsetManager/__tests__/commitOffsets.spec.js src/consumer/offsetManager/__tests__/getCoordinator.spec.js
YARN_CACHE_FOLDER=.codex-tmp/yarn-cache yarn prettier --check src/errors.js src/errors.spec.js src/consumer/offsetManager/index.js src/consumer/offsetManager/__tests__/commitOffsets.spec.js src/consumer/offsetManager/__tests__/getCoordinator.spec.js
git diff --check origin/master...HEAD
```

All passed.

The visible Azure status is still red for the Kafka-backed jobs (`Admin`, `Broker`, `Consumer`, `OauthBearer`, `Others`, `Producer`) while `Lint` and `Typescript types` pass. I could not read the Azure logs anonymously, but those failing jobs are the ones that run the current pipeline's direct `docker-compose` calls. The same CI matrix is green on #1781 after switching the scripts/pipeline to Compose v2-compatible invocation, so I would not treat the current red Kafka-backed jobs as evidence of a regression in this patch without a rerun after that CI fix.

